PR #10, opened 2016-12-06
Import whitelisting

bglebrun:
Does a string compare for host names. I implemented an interface for all of our document types that includes a function GetHostName which will return the host name field in the document. For DNS and UCONN structs, this returns an empty string which my function interprets as no input, then allows. Otherwise we use golang's string comparison to see if our host names match, then returns true or false based on that.

If, passing that, our string isn't white listed, we add the document to our database. Otherwise if we detect a white listed host, our docwriter will then check if the user is allowing import of white listed domains. If true, writes the document to the database, otherwise skips the document.


Zalgo2462:
Good work so far, just a few changes need to be made.

commands/analyze.go 98: Remove comment
etc/rita.go 40: Remove WIP comment

parser/conn.go 55-59: Let's change func (in Conn) GetHostName() string to func (in Conn) isWhiteListed(string[]) bool. The isWhiteListed function on each document will loop through the white list and return whether or not it matches. This will result in some code duplication for now (looping through the array and checking), but it will allow us to build in more features down the road. For example, we can migrate away from simple string checking and pass in whitelist objects which could specify something like "Whitelist if doc type is conn and dest is 192.168.1.1 on dest port 80" 

parser/conn.go and parser/dns.go: I would check the whitelist against the Destination fields, but requirements planning is needed before we change that.

parser/docwriter/docwriter.go 107-109: Remove this function. It only gets called in one place, and it is pass by value (no pointer/reference) so it will copy the document each time it is used.

parser/docwriter/docwriter.go 117-113: If we change to calling isWhiteListed on each document by passing the white list to each document, we can get rid of this function. Also, this function should have taken in a document pointer, as it stands now, it would copy the document everytime it is called.

parser/docwriter/docwriter.go 153-160: It looks like we are still writing the document to the database if the document has been whitelisted. Under the proposed changes, we can simplify this to if (!(d.ImportWl && doc.isWhitelisted(d.Whitelist)) then write document. Notice d.ImportWl comes first in the statement. This way the condition will short circuit if we don't want to use the white list, and we don't have to check the whole white list against the document. 

Thanks for your time and effort!
Logan L

Zalgo2462:
I think we're almost good to go, just remove the uncalled isWhitelisted function in docwriter 107-109 and change doc.Doc.isWhitelisted to towrite.isWhitelisted on line 130 in docwriter.
Cheers and Happy Holidays,
Logan L

bglebrun:
Both done, also cleaned up a bit.

PR #13, opened 2017-01-06
Install scripts

bglebrun:
Created bash scripts that will automatically install all necessary dependencies, creates a golang dev environment and sets up the needed version of MongoDB.


Zalgo2462:
The netcat found in the package manager is not compatible with team cymru (one intelligence source we use) unfortunately. The netcat found in the repo is openbsd's netcat. GNU netcat can be found at http://netcat.sourceforge.net/ This will have to be manually installed in /usr/local/rita. (or /usr/local/bin and change the default option in the config to point towards it)

Zalgo2462:
Also, can we integrate config script 2 into install.sh? (It looks like they do a lot of the same stuff).
Side note I like the colors :)

Zalgo2462:
Also, after some research it appears ubuntu's default golang package is out of date :( 

https://github.com/golang/go/wiki/Ubuntu

Zalgo2462:
Ugh. The version located in the repo mentioned there is also out of date. Looks like installing from https://golang.org/dl/ into /usr/local/ is the way to go.

Zalgo2462:
https://golang.org/doc/install has the correct instructions.

bglebrun:
Aware of a bug, the install script doesn't recognize 'go get' as a command, investigating right now.

bglebrun:
I know why, path, gopath, and gobin aren't set because ./install.sh requires us to run it as sudo which I remember now why I wanted to split these as their own config files. Do we want to install golang as root and install rita to the gopath in root's directory?

JonZeolla:
Does using `sudo -E` resolve your issue?  Sounds better than doing everything as root to me.

bglebrun:
Doesn’t, the instructions I used call for go to be installed at /usr/local/go/bin , and since we’re running as root and that location is not a part of the normal path we can’t directly invoke it.

I’m about to test a potential patch, will report how it works and push.

bglebrun:
Done, and yes for now the installer only works if we invoke it with 'sudo -E ./install.sh', I'll look into more elegant ways to maybe invoke the installers.

Zalgo2462:
Just making sure we use the most up to date apt repo for bro.
The repo instructions are here: https://www.bro.org/download/packages.html

JonZeolla:
I would like to help with this script.  Is there a plan to merge it into master soon?  If so, then I can provide a follow-on PR with some further enhancements.

install.sh
```diff
@@ -87,7 +87,7 @@ __install() {
   rm netcat-0.7.1.tar.gz
   cd netcat-0.7.1
   mkdir -p ~/go/src/github.com/ocmdev/rita/nc
-  ./configure --prefix=~/go/src/github.com/ocmdev/rita/nc
+  ./configure --prefix=$HOME/go/src/github.com/ocmdev/rita/nc
```
JonZeolla: Why would you want to put this under the go source folder?  It seems like it should be under something like `/usr/local/bin`.

Zalgo2462:
GNU netcat is a runtime dependency which RITA relies on in order to talk to Team Cymru. Normally, we would install this under /usr/local/bin, but since GNU netcat is very different from the OpenBSD netcat most people use, we do not want to overwrite the existing binary of the same name. It was decided we should store GNU netcat under /usr/local/rita. 
(We don't want to install it into the source tree)

JonZeolla:
That sounds reasonable to me.

bglebrun:
Ok, yeah that was short sighted of me, fixing it right now. Did we need to put /usr/local/rita as a path variable or just make the config default to that location?

Zalgo2462:
We need to change bglebrun/rita back to ocmdev/rita before we can merge this.

PR #15, opened 2017-01-08
Incorporate log mover scripts, fix tbd segfault, improve user experience, code consistency updates

Zalgo2462:
Log Mover Scripts: A push based system for moving logs from bro nodes to a Rita server has been written in logmover/push.

TBD suffered a segfault in the new code when two hosts communicated more than the threshold within the same second. When we remove sub-second connections, we may not have enough data to analyze.

User experience improvements include the ability to import a directory directly into a database (rita import -i [log-dir] -d [database-name]). This was made possible by implementing a DefaultDatabase option in the bro config, such that if a file doesn't match the DirectoryMap, it will be placed in the DefaultDatabase if it has been specified. Analyze will no longer re-analyze analyzed datasets.The option to reset analysis has been greatly simplified. Additionally, users can now delete databases managed by RITA. 

Consistency has been enforced for a couple vocabulary choices. All references to 'dataset' have been replaced by 'database.' All references to 'analyse' have been replaced by 'analyze'


joelillo:
Thanks Logan!!

PR #19, opened 2017-01-23
Make sure beaconing runs on a single core

Zalgo2462:
p.s. go doesn't have a min/max for ints in the standard library


joelillo:
Thanks Logan!

joelillo:
Oops wrong button

PR #23, opened 2017-02-03
Added cross reference analysis for scanning and beaconing. Additionally fixed aggregations.

Zalgo2462:
Aggregations were using .all to pull the results back into the client every time we aggregated. I changed the method to return an iterator. Additionally, I added cross reference analysis to the analysis modules. It is designed with an expansible selector framework. Scanning and beaconing selectors have been created.

WARNING!!!!! Unfortunately, this code is poorly tested. It compiles and looks correct, but I am not in a position to test the code.

This pull request serves to inform developers of code to come. The crossref analysis will be represented by two collections: internXREF and externXREF. The schema for records in these collections is string: name, string: host where name is the name of the analysis module in which this host was found.
Mongo does not enforce unique restraints on it's indexes so for hosts found in multiple modules, multiple documents are created.

### DIFF
--- a/analysis/TBD/TBD.go
+++ b/analysis/TBD/TBD.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	mgo "gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
 
 	"github.com/ocmdev/rita/database"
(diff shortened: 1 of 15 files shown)


joelillo:
@Zalgo2462 I'll be working on doing some testing of the staging branch this week. I'm going to start with the existing changes. When those have been tested I would like to pull this in and do some additional testing. Do you feel like it is ready to pull this week?

PR #30, opened 2017-02-22
Fix imporing bug https://github.com/ocmdev/rita/issues/27

Zalgo2462:
- GOMAXPROCS on all operations
- Fix human readable flag in commands (was using global var)
- Removed unused padding funcs in commands
- Added control of thread creation for parser
- Majorly refactored the Watcher -> Parser -> Writer system
- Removed Thread config in bro config section
- Added support for deleting date based databases
  - MetaDB was not being updated correctly
- Better support for detecting re-parses
- Better output on parse

### DIFF
--- a/analysis/blacklisted/blacklisted.go
+++ b/analysis/blacklisted/blacklisted.go
@@ -3,7 +3,6 @@ package blacklisted
 import (
 	"crypto/md5"
 	"fmt"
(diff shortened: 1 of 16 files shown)


joelillo:
Thank you Logan! Looking over it now. Will be merging soon.

PR #31, opened 2017-02-22
Add Version Information

joelillo:
- Added version field in system config
- Added VERSION global variable in config that will be loaded at compile time
- Created Makefile to load version at build time
- Added make commands to install script.
- Added version to meta database
- Added version to app information

Uses _git describe_ to get version string. This will use the most recent release version/commit from github. 

Example Usage:
./rita -v
rita version v0.9.1-177-gcc1d528-dirty


Zalgo2462:
Very cool!

PR #74, opened 2017-05-26
New Parser Fixes #51

### LINKED ISSUE #51: Implement Mongo bulk insert for data import
The current implementation of the import feature uses individual Mongo insert commands for each data point that goes into the database.  This area is highly threaded, but it is worth researching using bulk inserts instead.

The current proposal that would require the least messing with the current threading code (a large undertaking) is to add in a buffer of some sort that sits between the import code and the Mongo database.  Instead of inserting directly into Mongo, the threads would place the data into an in-memory buffer.  When the buffer is full, additional code will take that data and perform a bulk insert into Mongo.

Zalgo2462:
There is a lot of new code here. The old parser has been ripped out and a new one has been put in. The new parser uses the core reflection concepts of the old one but has a different threading model.
The new parser uses N threads (usually CPUs/ 2) to open all of the files in the import directory, parse out their headers, figure out the target database and collection. Then, it checks if we have parsed any of the files previously and removes them from the list of files to parse. Then, it uses N threads to parse lines out of the files. Each file is processed by a singular thread rather than being shared across multiple threads. For each destination collection, a new thread and buffer is spun up, when a buffer is filled the corresponding thread issues an upload to MongoDB.  After uploading all of the data, the parser issues the commands to create the indexes on the newly created collections one by one.

Fixes #51 

Note: there are some changes across the DB control files that resulted from fixing linting errors in the files the parser links against.

Note: this pull changes the config file to include a buffer size variable under the bro config.


joelillo:
@Zalgo2462 Looking good!

Issue #85, opened 2017-06-07
Logging only enabled on import


ethack:
Logs are only written to the RitaLogDirectory when the `import` command is run.  They should be written during other commands as well except for `show-*` command output.

PR #87, opened 2017-06-08
Logging Fix Fixes Issues #85, #90

Zalgo2462:
Added options to enable/disable file logging and logging to the database. Made sure log hooks were enabled at all times. Fixed some linting issues.

### DIFF
--- a/analysis/beacon/beacon.go
+++ b/analysis/beacon/beacon.go
@@ -94,7 +94,6 @@ func newBeacon(res *database.Resources) *Beacon {
 
 // Run Starts the beacon hunt process
 func (t *Beacon) run() {
-	t.log.Info("Running beacon hunt")
 	session := t.resources.DB.Session.Copy()
 	defer session.Close()
 
--- a/analysis/blacklisted/blacklisted.go
+++ b/analysis/blacklisted/blacklisted.go
@@ -28,27 +28,29 @@ import (
 type (
 	// Blacklisted provides a handle for the blacklist module
 	Blacklisted struct {
-		db              string                    // database name (customer)
-		batch_size      int                       // BatchSize
-		prefetch        float64                   // Prefetch
-		resources       *database.Resources       // resources
-		log             *log.Logger               // logger
-		channel_size    int                       // channel size
-		thread_count    int                       // Thread count
-		blacklist_table string                    // Name of blacklist table
-		safeBrowser     *safebrowsing.SafeBrowser // Google safebrowsing api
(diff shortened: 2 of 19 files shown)


ethack:
I get the same error for most commands.
```
➜  rita git:(73d0721) ./rita show-databases
ERRO[0000] newMetaDBHandle failed to build database (aborting)  error="must pass name of collection to create"
➜  rita git:(73d0721) ./rita test-config
ERRO[0000] newMetaDBHandle failed to build database (aborting)  error="must pass name of collection to create"
```
I even bumped the LogLevel up to 3 and it didn't change the output.

Zalgo2462:
Sorry, I should have been more clear. This changes the config file. You will have to update it from etc/rita.yaml

ethack:
Got it.  My previous goto was to run `test-config` and look for empty config values. I only learned that trick recently :)

Do I have to make any modifications to the config file to enable logging?  LogToFile and LogToDB are both set to their default true values.  I didn't change anything in the config from the default actually.  For my tests I'm specifying the bro files and database manually from command line.

Several points:
* I'm not getting any output at all on the command line when running `rita analyze`.
* ~/.rita/logs is not being created.
* I do see the logs collection in the metadatabase and it has entries from running `rita import` but not `rita analyze`.

Zalgo2462:
This pr now fixes issue #90 as well.

ethack:
I tested this out and couldn't find any issues.  All my previous points have been addressed.

Couple of minor points that aren't important right now but may be worth documenting later:
* Since different levels (debug, warn, info) are sent to different log files and timestamps are only down to the second, there isn't a way of reconstructing the exact order of messages between two or more files.  I don't see this as a limitation that needs addressing since logging to the database gives you everything you need.
* There's no analog to `delete-database` for logs.  You have to run a mongo command.  Again, not really an issue.  Maybe someone will someday want to reduce the space and not know how to run the right mongo command but that's probably better solved with documentation.
* rita-blacklists output is not in the same format.  (see sample below).  This is not a problem right now but if we want it to match in the future we'll have to decide how closely to integrate with RITA's logger and log config.

```
	[-] Running User Agent Analysis
	[-] Running Beaconing Analysis
	[-] Running Blacklisted Analysis
2017/06/09 18:42:55 Updating blacklist source:  myip-ms . This may take a few minutes...
2017/06/09 18:46:32 Blacklist: myip-ms parsed 15 of 318841 lines in file.
	[-] Running Scanning Analysis
	[-] Running Cross Reference Analysis
```

Zalgo2462:
Agreed on all points. Merging

PR #89, opened 2017-06-08
Add Confirmation for Reset-Analysis Fixes #84

### LINKED ISSUE #84: Confirm when reseting analysis
Add in a confirmation prompt (e.g. "Are you sure you want to reset analysis on 'X' database?") when using the `rita reset-analysis` command.  The caveat here is that if you don't specify the database it will reset all databases and the prompt needs to reflect that (e.g. "Are you sure you want to reset analysis on ALL databases?")

Similar to issue [#46](https://github.com/ocmdev/rita/issues/46).  And the implementation will likely be similar.

Zalgo2462:
Fixes #84.

### DIFF
--- a/commands/delete-database.go
+++ b/commands/delete-database.go
@@ -24,7 +24,7 @@ func init() {
 				return cli.NewExitError("Specify a database with -d", -1)
 			}
 
-			fmt.Println("Are you sure you want to delete database ", c.String("database"), " [Y/n]")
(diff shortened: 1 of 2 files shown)


ethack:
I tested this out and it worked great for me.  

To get around the prompt for every database when doing the global reset I ran`yes | rita reset-analysis`.

PR #93, opened 2017-06-09
Restore ability to split a file amongst multiple dates

Zalgo2462:
When we brought in the new parser, we set the target date for all entries in a bro file based on the first entry in the file, trusting that the file was split correctly before analysis. However, even when using the bro daemon, a log file may contain entries from multiple dates. I have reintegrated the ability to split a file amongst multiple databases when using dates.


commands/import.go
```diff
@@ -59,7 +59,7 @@ func doImport(c *cli.Context) error {
 
 	fmt.Printf("Importing %s\n", res.System.BroConfig.LogPath)
 	importer := parser.NewFSImporter(res, threads, threads)
-	datastore := parser.NewMongoDatastore(res.DB.Session,
+	datastore := parser.NewMongoDatastore(res.DB.Session, res.MetaDB,
```
Zalgo2462: The mongo datastore needs a handle to the meta db to add new databases when they are needed.

database/meta.go
```diff
@@ -86,9 +86,20 @@ func (m *MetaDBHandle) DeleteDB(name string) error {
 	ssn.DB(name).DropDatabase()
 
 	//delete any parsed file records associated
-	_, err = ssn.DB(m.DB).C("files").RemoveAll(bson.M{"database": name})
-	if err != nil {
-		return err
+	if db.UsingDates {
```
Zalgo2462: We need to remove the file entries which match the database to be removed. We split the database name and remove any files which were destined for the base database name and contained entries corresponding to the date.

parser/fileparsetypes/fileparsetypes.go
```diff
@@ -32,6 +32,7 @@ type IndexedFile struct {
 	Hash             string        `bson:"hash"`
 	TargetCollection string        `bson:"collection"`
 	TargetDatabase   string        `bson:"database"`
+	Dates            []string      `bson:"dates"`
```
Zalgo2462: A file can be split across multiple dates

parser/fsimporter.go
```diff
@@ -59,9 +62,8 @@ func (fs *FSImporter) Run(datastore *MongoDatastore) {
 
 	indexedFiles = removeOldFilesFromIndex(indexedFiles, fs.res.MetaDB, fs.res.Log)
 
-	createNewDatabases(indexedFiles, fs.res.MetaDB, fs.res.Log)
```
Zalgo2462: We can no longer create the databases ahead of time.

parser/mongodatastore.go
```diff
@@ -28,26 +30,30 @@ type collectionStore struct {
...
 	return &MongoDatastore{
-		dbMap:      make(map[string]map[string]collectionStore),
-		bufferSize: bufferSize,
-		session:    session,
-		logger:     logger,
-		waitgroup:  new(sync.WaitGroup),
-		mutex1:     new(sync.Mutex), //mutex1 syncs the first level of map access
-		mutex2:     new(sync.Mutex), //mutex2 syncs the second level of map access
+		dbMap:       make(map[string]map[string]collectionStore),
+		existingDBs: metaDB.GetDatabases(),
```
Zalgo2462: existingDBs holds the databases that existed before we called import. If an entry is destined for an already existing db, we exit with an error

Zalgo2462:
Running this code on a large dataset before merging. So far so good.

PR #99, opened 2017-06-23
Fix various cli bugs. Remove all flag.

Zalgo2462:
There isn't an open issue for this PR. However, while working on the new blacklist module, I became rather annoyed with rita's cli. I removed the all flag so it displays all results by default. I figure you can use `head` to limit the output. Then, I made sure the configuration file flag was actually being used on all of the commands. Finally, I fixed some errors that were being swallowed, and added some better error reporting. For instance if you issue a show command with an invalid database, it will respond with `No results were found for (database).`

### DIFF
--- a/commands/show-beacons.go
+++ b/commands/show-beacons.go
@@ -21,7 +21,7 @@ func init() {
 		Flags: []cli.Flag{
 			humanFlag,
 			databaseFlag,
-			allFlag,
+			configFlag,
 		},
 		Action: showBeacons,
 	}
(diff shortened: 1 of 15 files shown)


commands/show-beacons.go
```diff
@@ -33,28 +32,31 @@ func showBeacons(c *cli.Context) error {
...
 
 	var data []beaconData.BeaconAnalysisView
-	cutoffScore := .7
-	if c.Bool("all") {
-		cutoffScore = 0
-	}
 
 	ssn := res.DB.Session.Copy()
-	resultsView := beacon.GetBeaconResultsView(res, ssn, cutoffScore)
+	resultsView := beacon.GetBeaconResultsView(res, ssn, 0)
```
ethack: To me, replacing `cutoffScore` with `0` makes this harder to read.  I understand why a variable is no longer needed, but it's a shame to remove the symbolism.  I'd suggest either adding a comment to explain the 0 value or keeping the variable for readability's sake.

ethack: A similar line exists [here](https://github.com/ocmdev/rita/blob/5321fb65da13a98aef543944c2d0bbd06ecc1b6b/reporting/report-beacons.go#L29).  The `GetBeaconResultsView` function definition has the name cutoffScore for its parameter and it's not hard to look that up so I retract my suggestion about leaving the variable.  I guess what's not obvious is that the value 0 is the lowest possible score and thus equates to getting all results.  So would it make sense to make a comment with what the intention is?  e.g. `// Get all beacon results since 0 is lowest possible score`  Or it might make sense to document the possible values for the `cutoffScore` parameter in the `GetBeaconResultsView` [function](https://github.com/ocmdev/rita/blob/5321fb65da13a98aef543944c2d0bbd06ecc1b6b/analysis/beacon/beacon.go#L65)?

ocmdev: The latter suggestion seems to be more efficient in my opinion. The parameter exists so that cross reference analysis can pull only the "true" beacons out of the dataset. I will get  the alteration up shortly.

Zalgo2462:
I went the latter suggestion and updated the documentation on the function.

PR #100, opened 2017-06-26
New blacklist module. Fixes #48

### LINKED ISSUE #48: Migrate Safebrowsing code to rita-blacklist
Move the code that handles the Safebrowsing functionality in https://github.com/ocmdev/rita/blob/master/analysis/blacklisted/blacklisted.go and write a new hostlist module here https://github.com/ocmdev/rita-blacklist/tree/master/hostlist

Zalgo2462:
The previous blacklist module possessed some issues which involved mixing up ips, urls, and hostnames. Additionally, it had a hard time distinguishing between blacklisted hosts which initiated connections and received them. These kludges lead to difficulty in expanding the blacklisted module.

This new blacklist module accompanies [a new rita blacklist backend](http://github.com/ocmdev/rita-bl). The new backend is meant to handle various types of blacklists, enforce the database abstraction, and provide support for blacklists that are managed by outside software. This has allowed the removal of the google safebrowsing code from RITA and the addition of the code to rita-bl. (Fixes issue #48)

The manner in which user defined lists has changed, as has the RITA config file. 
User defined blacklists are expected to be simple line separated lists. These simple files are then registered with RITA in the config file.
Here is an example of a new blacklisted configuration section.
```
BlackListed:
    Database: rita-blacklist
    # These are blacklists built into rita-blacklist. Set these to false
    # to disable checks against them.
    myIP.ms: true
    MalwareDomains.com: true
    MalwareDomainList.com: true

    # Google SafeBrowsing requires an api key and a file to cache the results in.
    # If either APIKey or Database is an empty string, Google SafeBrowsing will 
    # not be queried.
    SafeBrowsing:
        APIKey: "YOUR API KEY"
        Database: $HOME/.rita/safebrowsing

    # These are custom blacklists that you may define. They are lists of either
    # file paths or urls. These custom blacklists are expected to be simple,
    # line separated text documents containing a list of blacklisted entries.

    # Example: CustomIPBlacklists: ["$HOME/.rita/myIPBlacklist.txt"]
    # myIPBlacklist.txt would look like this:
    # 192.168.0.1
    # 10.10.174.1

    #Lists containing both IPv4 and IPv6 addresses are acceptable
    CustomIPBlacklists: []
    #Lists containing hostnames, domain names, and FQDNs are acceptable
    CustomHostnameBlacklists: ["$HOME/.rita/hostnamesBL.txt"]
    #URLs must each contain a protocol, a host, and a resource
    #Ex: http://google.com/
    #Ex: ftp://myftpserver.com/a/file/over/here.txt
    CustomURLBlacklists: ["$HOME/.rita/urlsBL.txt"]

    # Table names
    SourceIPsTable: blSourceIPs
    DestIPsTable: blDestIPs
    HostnamesTable: blHostnames
    UrlsTable: blUrls
```

The new blacklist module brings with it four new commands. 
```
     show-bl-hostnames       Print blacklisted hostnames which recieved connections
     show-bl-source-ips        Print blacklisted IPs which initiated connections
     show-bl-dest-ips            Print blacklisted IPs which recieved connections
     show-bl-urls                   Print blacklisted URLs which were visited
```
The html output has been updated with a page for each of these commands.
Finally, crossref has been simplified to source and destination collections rather than internal/ external. The new blacklist information has been integrated into this new crossref configuration.


analysis/blacklist/blacklist.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/ocmdev/rita-bl/list"
+	"github.com/ocmdev/rita-bl/sources/lists"
+	"github.com/ocmdev/rita-bl/sources/rpc"
+	"github.com/ocmdev/rita/config"
+	"github.com/ocmdev/rita/database"
+	log "github.com/sirupsen/logrus"
+	mgo "gopkg.in/mgo.v2"
+)
+
+type resultsChan chan map[string][]blDB.BlacklistResult
```
Zalgo2462: We chunk up the checks against RITA-BL. RITA-BL returns maps of (hosts or ip addresses or urls) to their results. The channel is used for communication.

analysis/blacklist/blacklist.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+//buildCustomBlacklists gathers a custom blacklist from a url or file path
+func buildCustomBlacklists(entryType list.BlacklistedEntryType, paths []string) []list.List {
+	var blacklists []list.List
+	for _, path := range paths {
+		newList := lists.NewLineSeperatedList(
+			entryType,
+			path,
+			86400,
```
Zalgo2462: cache time, could use a comment

analysis/blacklist/blacklist.go
```diff
@@ -0,0 +1,223 @@
...
+			tryOpenFileThenURL(path),
+		)
+		blacklists = append(blacklists, newList)
+	}
+	return blacklists
+}
+
+//provide a closure over path to read the file into a line separated blacklist
+func tryOpenFileThenURL(path string) func() (io.ReadCloser, error) {
+	return func() (io.ReadCloser, error) {
```
Zalgo2462: Rita-bl expects a function which returns (io.ReadCloser, error) to get the data from.

analysis/blacklist/blacklist.go
```diff
@@ -0,0 +1,223 @@
...
+	)
+
+	//set up the lists to check against
+	ritaBL.SetLists(buildBlacklists(res.System)...)
+
+	//set up remote calls to check against
+	ritaBL.SetRPCs(buildBlacklistRPCS(res)...)
+
+	//update the lists
+	ritaBL.Update()
```
Zalgo2462: This line forces ritaBL to update its cache.

analysis/blacklist/blacklist.go
```diff
@@ -0,0 +1,223 @@
...
+	for _, collection := range collections {
+		ssn.DB(currentDB).C(collection).Create(&mgo.CollectionInfo{
+			DisableIdIndex: true,
+		})
+	}
+
+	//create the data
+	//TODO: refactor these into modules
+	bufferSize := 1000
+	buildBlacklistedIPs(uniqueSourceIter, res, ritaBL, bufferSize, true)
```
Zalgo2462: true is for source ip addresses, false if for destination ip addresses.

config/config.go
```diff
@@ -176,6 +183,12 @@ func expandConfig(reflected reflect.Value) {
 			expandConfig(f)
 		} else if f.Kind() == reflect.String {
 			f.SetString(os.ExpandEnv(f.String()))
+		} else if f.Kind() == reflect.Slice && f.Type().Elem().Kind() == reflect.String {
```
Zalgo2462: expand lists

analysis/blacklist/blacklist.go
```diff
@@ -0,0 +1,85 @@
+package blacklist
+
+import (
+	bl "github.com/ocmdev/rita-blacklist2"
```
lisaSW: At some point, we can probably just rename this rita-blacklist and get rid of the other one...

analysis/blacklist/blacklist.go
```diff
@@ -0,0 +1,85 @@
...
+		res.Log.Error("could not open up google safebrowsing for blacklist checks")
+	}
+
+	ritaBL.SetLists(lists.NewMyIPmsList(), lists.NewMdlList())
+	ritaBL.Update()
+
+	//get our data sources
+	ssn := res.DB.Session.Copy()
+	defer ssn.Close()
+	uniqueSourcesAggregation := getUniqueIPFromUconnPipeline("src")
```
lisaSW: Is there a chance that the "src" and "dst" strings could ever change? Should they be variables?

analysis/blacklist/blacklist.go
```diff
@@ -0,0 +1,85 @@
...
+		uniqueDestAggregation,
+	)
+	hostnamesIter := ssn.DB(res.DB.GetSelectedDB()).C(
+		res.System.DNSConfig.HostnamesTable,
+	).Find(nil).Iter()
+	urlIter := ssn.DB(res.DB.GetSelectedDB()).C(
+		res.System.UrlsConfig.UrlsTable,
+	).Find(nil).Iter()
+
+	bufferSize := 1000
```
lisaSW: Is this something that could or should be in the config file?

analysis/blacklist/blacklist.go
```diff
@@ -0,0 +1,223 @@
...
+	collections := []string{sourceIPs, destIPs, hostnames, urls}
+	for _, collection := range collections {
+		ssn.DB(currentDB).C(collection).Create(&mgo.CollectionInfo{
+			DisableIdIndex: true,
+		})
+	}
+
+	//create the data
+	//TODO: refactor these into modules
+	bufferSize := 1000
```
lisaSW: Could or should this be a config option?

analysis/blacklist/blacklist.go
```diff
@@ -0,0 +1,223 @@
...
+	buildBlacklistedIPs(uniqueSourceIter, res, ritaBL, bufferSize, true)
+
+	buildBlacklistedIPs(uniqueDestIter, res, ritaBL, bufferSize, false)
+
+	buildBlacklistedHostnames(hostnamesIter, res, ritaBL, bufferSize)
+
+	buildBlacklistedURLs(urlIter, res, ritaBL, bufferSize, "http://")
```
lisaSW: What happens in the case of a different url prefix?

analysis/blacklist/blacklist.go
lisaSW: comment would be very helpful :)

analysis/blacklist/ips.go
```diff
@@ -0,0 +1,162 @@
...
+		if i == bufferSize-1 {
+			//excuse the memory hacking to get better performance
+			//We need the buffer to be of type ipAggregateResult for
+			//proper marshalling, but we need strings for rita-blacklist.
+			//The underlying memory for ipAggregateResult is that of a string
+			//since it is the only field in the struct.
+			//So we can safely view buff as an array of strings using a
+			//reinterpret cast. Then, we can dereference the pointer to the array
+			//and use the variadic syntax to pass the array to CheckEntries.
+			indexesArray := (*[]string)(unsafe.Pointer(&buff))
```
lisaSW: This structure is being treated as an array of strings, but if that ever changes, this will break. Will we ever need to add fields to this structure?

analysis/blacklist/hostnames.go
```diff
@@ -0,0 +1,124 @@
...
+
+func checkRitaBlacklistHostnames(hostnames *mgo.Iter, blHandle *bl.Blacklist,
+	bufferSize int, resultsChannel resultsChan) {
+	i := 0
+	//read in bufferSize entries and check them. Then ship them off to the writer
+	var buff = make([]hostnameShort, bufferSize)
+	for hostnames.Next(&buff[i]) {
+		if i == bufferSize-1 {
+			//see comment in checkRitaBlacklistIPs
+			indexesArray := (*[]string)(unsafe.Pointer(&buff))
```
lisaSW: This structure is being treated as an array of strings, but if that ever changes, this will break. Will we ever need to add fields to this structure?

etc/rita.yaml
```diff
@@ -56,9 +56,44 @@ Bro:
...
+    # 192.168.0.1
+    # 10.10.174.1
+
+    # Lists containing both IPv4 and IPv6 addresses are acceptable
+    CustomIPBlacklists: []
+    # Lists containing hostnames, domain names, and FQDNs are acceptable
+    CustomHostnameBlacklists: []
+    # URLs must each contain a protocol, a host, and a resource
+    # Ex: http://google.com/
+    # Ex: ftp://myftpserver.com/a/file/over/here.txt
```
lisaSW: On [this line](https://github.com/ocmdev/rita/pull/100/files#diff-31394ce29daf4599be7f6fb16b6516a4R100), it looks like the url prefix is hardcoded to http...  Will other prefixes still work and if so, how?

analysis/blacklist/blacklist.go
Zalgo2462: This can be a config option, but I don't see any reason why it should be changed. It is the amount of entries to check against rita-bl at a time. The chunking is mainly for rita-bl RPC's such as google safebrowsing. Some RPC's can take advantage of this chunked behavior when querying the internet. The proper thing to do here is probably change it to a constant in the blacklist package.

lisaSW: Yep, please make this a constant

analysis/blacklist/blacklist.go
lisaSW: sounds good.

analysis/blacklist/hostnames.go
lisaSW: sounds good.

etc/rita.yaml
lisaSW: sounds good.

PR #116, opened 2017-07-07
Apply TLS to MongoDB Connection Fixes #105

### LINKED ISSUE #105: RITA does not support TLS encryption / MongoDB authentication
RITA should be able to communicate with MongoDB over encrypted channels, and properly authenticate to MongoDB. Note: this should be an optional configuration option.

Zalgo2462:
Fixes #105 

In order to configure RITA to  authenticate to a MongoDB server over tls, set up your config like so:
MongoDB:

```
  ConnectionString: mongodb://username:password@some.address:27017
  AuthenticationMechanism: SCRAM-SHA-1
  TLS:
    Enable: true
    CAFile: /path/to/my/private/certificate/authority
```
Of course the CAFile is not required if using a verified certificate.

### DIFF
--- a/analysis/blacklist/blacklist.go
+++ b/analysis/blacklist/blacklist.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"os"
 
-	"github.com/ocmdev/mgosec"
 	bl "github.com/ocmdev/rita-bl"
 	blDB "github.com/ocmdev/rita-bl/database"
 	"github.com/ocmdev/rita-bl/list"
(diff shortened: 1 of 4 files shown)


analysis/blacklist/blacklist.go
```diff
@@ -28,8 +27,22 @@ const ritaBLBufferSize = 1000
 func BuildBlacklistedCollections(res *database.Resources) {
 	//capture the current value for the error closure below
 	currentDB := res.DB.GetSelectedDB()
-
-	blDB, err := blDB.NewMongoDB(res.System.DatabaseHost, mgosec.None, "rita-bl")
+	var err error
+	var blDatabase blDB.Handle
+	if res.System.MongoDBConfig.TLS.Enabled {
+		blDatabase, err = blDB.NewSecureMongoDB(
```
Zalgo2462: pass the mgo sec details to ritabl

analysis/blacklist/blacklist.go
```diff
@@ -38,7 +51,7 @@ func BuildBlacklistedCollections(res *database.Resources) {
 
 	//set up rita-blacklist
 	ritaBL := bl.NewBlacklist(
-		blDB,
+		blDatabase,
```
Zalgo2462: blDB shadowed an import name

config/config.go
```diff
@@ -33,6 +35,20 @@ type (
 		Version           string
 	}
 
+	//MongoDBCfg contains the means for connecting to MongoDB
+	MongoDBCfg struct {
+		ConnectionString    string               `yaml:"ConnectionString"`
+		AuthMechanism       string               `yaml:"AuthenticationMechanism"`
+		AuthMechanismParsed mgosec.AuthMechanism `yaml:"AuthenticationMechanismParsed,omitempty"`
```
Zalgo2462: The system parses the AuthMechanism field and stores it in AuthMechanismParsed

database/resources.go
```diff
@@ -84,13 +88,39 @@ func InitResources(cfgPath string) *Resources {
 	if !metaDB.isBuilt() {
 		metaDB.createMetaDB()
 	}
+
+	//Begin logging to the metadatabase
 	if conf.LogConfig.LogToDB {
-		addMongoLogger(log, conf.DatabaseHost, conf.BroConfig.MetaDB,
-			conf.LogConfig.RitaLogTable)
+		log.Hooks.Add(
+			mgorus.NewHookerFromSession(
```
Zalgo2462: Take advantage of our already open session

analysis/blacklist/blacklist.go
ethack: So instead of using mgosec directly, it's used indirectly through rita-bl.  You just have to pass the config details over.  Is that right?

Zalgo2462: Rita uses mgosec to dial in on its own. Then RITA-BL uses the same parsed mgosec connection details to dial in separately.

PR #119, opened 2017-07-10
Enable a Custom Socket Timeout

### LINKED ISSUE #108: RITA may crash if MongoDB operations exceed 2 hours.
Currently, the socket time-out is hard coded at 2 hours. This should be moved a configuration value since large datasets may need to be analysed. A setting of 0 disable time-outs.

Zalgo2462:
Fixes #108


ethack:
Looks good to me.

Issue #133, opened 2017-08-14
RITA Drops Connections Which Persist Across Dates When UseDates Is Configured


Zalgo2462:
Bro IDS timestamps signify the start of the recorded event. However, Bro writes the event when the event has finished. A connection which opens at 23:58 on day 1 may persist until 00:04 on day 2. Bro will write the event to day 2's dataset with a timestamp recorded on day 1.

RITA uses timestamps to determine which dataset to import a record into. If the dataset for day 1 has already been parsed and analyzed, RITA will not insert the problem connection to the analyzed dataset., nor will it be inserted into day 2's dataset. RITA will log this with an error. 

Is there a better solution?

ethack:
We're deprecating the `UseDates` function.  We will remove it in favor of other solutions: possibly a separate script or even an import/analyze methodology that achieves the same result along with Bro's daily log rotation.

PR #134, opened 2017-08-14
Prevent Parser from Panicking on Import Into Analyzed DB

Zalgo2462:
Added more granular locking to parser, prevent panic when importing data into an already analyzed dataset.

Establishes base line behavior for issue #133

### DIFF
--- a/parser/datastore.go
+++ b/parser/datastore.go
@@ -1,0 +1,17 @@
+package parser
+
+import "github.com/ocmdev/rita/parser/parsetypes"
+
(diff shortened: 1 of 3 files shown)


Zalgo2462:
I have tested this, but I would like someone else to test it as well.

SamuelCarroll:
I'm on it!

SamuelCarroll:
Just tested (and after a couple hiccups) it looks good! I'll have a look at the code too

SamuelCarroll:
Looks good to me

Issue #103, opened 2017-06-28
Separate internal mongo settings from config.yaml


ethack:
Can we move the Mongo table settings to a separate configuration file (e.g. ~/.rita/database-structure.yaml)?  This is more of an open question/discussion that needs to be decided before implementing.

```
# NOTE: DO NOT CHANGE THE SETTINGS BELOW UNLESS YOU ARE FAMILIAR WITH THE CODE #
Structure:
    ConnectionTable: conn
    HttpTable: http
    DnsTable: dns
    UniqueConnectionTable: uconn
    HostTable: host

BlackListed:
    ThreadCount: 2
    ChannelSize: 1000
    BlackListTable: blacklisted
    Database: rita-blacklist

Dns:
    ExplodedDnsTable: explodedDns
    HostnamesTable: hostnames

Crossref:
    InternalTable: internXREF
    ExternalTable: externXREF
    BeaconThreshold: .7

Scanning:
    ScanThreshold: 50
    ScanTable: scan

Beacon:
    DefaultConnectionThresh: 24
    BeaconTable: beacon

Urls:
    UrlsTable: urls

UserAgent:
    UserAgentTable: useragent

MetaTables:
    FilesTable: files
    DatabasesTable: databases


# Adjusting batchsize and prefetch may help speed up certain database queries
BatchSize: 300
Prefetch: 0.33
```

This would allow the following improvements:

* When RITA code updates that adds or changes these fields, using an old config will cause errors.  However, this means the user needs to update their config file and possibly either overwrite their modifications or manually copy them over.  Having a separate file would allow changes to database structure without forcing users to update their config.
* This would declutter the user's config file and make it less confusing to edit.

Zalgo2462:
This will be easier to implement once the TLS and RITA-BL changes go through since they both alter the configuration file.

ethack:
Additionally, we should consider versioning config files and having rita automatically detect older configs and give the user a more useful error message.

Zalgo2462:
Thats not a bad thought. However, I'd like to get a proposal put together for that feature as that sounds like rat's nest of a feature to support (off the top of my head).

Should get in a PR for this issue by the end of day.

PR #135, opened 2017-08-17
Split config into two files, updated install script to be more interactive

Zalgo2462:
Split the table config out into its own file in hopes that that the table config will be more stable than the user config. This is a transition towards hard coding the table config into the RITA binary.

Also updated the installer. If you try to run the new installer you will get the following output:
image
The failure is due to table.yaml not being in master. The debconf errors are a result of running the script in docker without the apt-utils package installed.

However, if the file existed the following output would be displayed.
image

Fixes issue #103

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -19,12 +19,12 @@ set -o pipefail
 OLD_PS1=$PS1
 PS1=" "
 # Hack the interactive flag to get around other .bashrc's
-set -i
+set +i
 
 source $HOME/.bashrc
 
 # Clean up our hacks
-set +i
+set -i
 PS1=$OLD_PS1
 unset OLD_PS1
 
@@ -109,7 +109,7 @@ __install() {
 
 	# Check if RITA is already installed, if so ask if this is a re-install
 	if [ ! -z $(command -v rita) ] ||
-	[ -f $GOPATH/bin/rita ]
+	[ -d $HOME/.rita ]
 	then
 		printf "[+] RITA is already installed.\n"
 		read -p "[-] Would you like to erase it and re-install? [y/n] " -r
@@ -213,7 +213,7 @@ __install() {
 
 
 	# If the user is using sudo, give ownership to the sudo user
-	if [ -z ${SUDO_USER+x} ]
+	if [ ! -z ${SUDO_USER+x} ]
 	then
 		chown -R $SUDO_USER:$SUDO_USER $HOME/go
 		chown -R $SUDO_USER:$SUDO_USER $HOME/.rita
--- a/install.sh
+++ b/install.sh
@@ -3,24 +3,41 @@
 # RITA is brought to you by Offensive CounterMeasures.
 # offensivecountermeasures.com
 
+_NAME=$(basename "${0}")
+_FAILED="\e[91mFAILED\e[0m"
+_SUCCESS="\e[92mSUCCESS\e[0m"
+
+#Error handling
+#Kill 0 to kill subshells as well
+trap "printf '\n[!] Installation $_FAILED!\n'; kill 0" ERR INT
 set -o errexit
+set -o errtrace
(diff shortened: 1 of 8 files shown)


SamuelCarroll:
Quick glance through the code and it looks pretty good, I still need to run the changes

joelillo:
@Zalgo2462 I might need some help running this. My installation is failing without any info. 
image

install.sh
```diff
@@ -3,73 +3,116 @@
...
-_NAME=$(basename "${0}")
-_INSDIR="/usr/local"
+# Make sure to source the latest .bashrc
+# Hack the PS1 variable to get around ubuntu .bashrc
+OLD_PS1=$PS1
+PS1=" "
+# Hack the interactive flag to get around other .bashrc's
+set -i
+
+source $HOME/.bashrc
```
ethack: I'm having the same issue as @joelillo. My script is erroring out on this line. I'll post more updates if I can track down the cause.

install.sh
```diff
@@ -3,73 +3,116 @@
...
 set -o pipefail
 
-_NAME=$(basename "${0}")
-_INSDIR="/usr/local"
+# Make sure to source the latest .bashrc
+# Hack the PS1 variable to get around ubuntu .bashrc
+OLD_PS1=$PS1
+PS1=" "
+# Hack the interactive flag to get around other .bashrc's
+set -i
```
ethack: Commenting this out lets script execution continue.  And the install seemed to work. @Zalgo2462 do you remember what this hack was for?

install.sh
```diff
@@ -78,151 +121,129 @@ __install() {
...
 
-	printf "[+] Transferring files...\n\n"
-	mkdir $_RITADIR
+	apt-get update > /dev/null & __load "[+] Updating apt"
+	apt-get install -y mongodb-org > /dev/null & __load "[+] Installing MongoDB"
 
-	cp -r etc $_RITADIR/etc
-	cp LICENSE $_RITADIR/LICENSE
+	( # Build RITA
+		go get github.com/ocmdev/rita
```
ethack: Since this defaults to the `master` branch the install fails down below on line 211 when it expects the new `tables.yaml` file. Is there already a method of using this  script to install a different branch?

install.sh
ethack: Not sure if this is it or not but running the install with `sudo` but without this `set -i` line leaves `~/.rita` and `~/go` owned by root still.  Even though there are `chown` commands later on that should switch the ownership.

install.sh
Zalgo2462: Weird. Do you have a .bashrc file? I'll have to do some testing to narrow this down. I remember it worked for me at the time :/  The whole dance seems to be in place just to source the bashrc file automatically (i.e. if the user runs the install script twice in a row, we want to have the most up to date info on the status of the install).

Most bash rc files prevent being sourced from an interactive session. Ubuntu uses PS1 to prevent this, other systems check the interactive flag.

ethack: There is a .bashrc file, but it's just the default one that came with the OS.  I removed it and the script fails for a different reason.  But if I make an empty .bashrc then it goes through the install fine (until it fails later on due to the different branches).

ethack:
I updated the dockerfile to bake in the tables.yaml file.  This is how I tested the split config part of this PR and it worked fine for me there.

install.sh
Zalgo2462: Unfortunately not while using go get. We would have to manually make the path and git clone to use different branches. Alternatively we could use a proxy? maybe?

Zalgo2462:
Fixed the stall and the sudo problems.

ethack:
Note that the installer won't work correctly until this is merged into the master branch. Just leaving a comment here in case anyone runs into the issue.

PR #136, opened 2017-08-17
Fix  Blacklisted URLs False Positives

carrohan:


analysis/blacklist/urls.go
```diff
@@ -115,19 +118,77 @@ func fillBlacklistedURL(blURL *data.BlacklistedURL, longURL, db,
...
-	uniqueConnections := ssn.DB(db).C(uconnCollection).Find(connQuery).Iter()
+	uniqueConnections := ssn.DB(db).C(httpCollection).Pipe(httpPipeline).Iter()
 	var uconn structure.UniqueConnection
 	for uniqueConnections.Next(&uconn) {
 		totalBytes += uconn.TotalBytes
 		totalConnections += uconn.ConnectionCount
 		uniqueConnCount++
 	}
+	if uniqueConnections.Err() != nil {
+		fmt.Println(uniqueConnections.Err())
```
Zalgo2462: just return the error up

analysis/blacklist/urls.go
```diff
@@ -115,19 +118,77 @@ func fillBlacklistedURL(blURL *data.BlacklistedURL, longURL, db,
...
+		{
+			{ "$match", bson.M{"host": host, "uri": resource} },
+		},
+		{
+			{ "$group", bson.M{
+					"_id": bson.D{
+						{ "h", "$host" },
+						{ "u", "$uri" },
+						{ "s", "$id_origin_h" },
+						{ "d", "$id_resp_h" },
```
Zalgo2462: Unfortunately, your solution still grabs connections which may not have been part of the url request, due to reliance on uconn. 

This may be more correct (and possibly faster) to do in parts.
Start a pipeline on http
match on host and uri
project down to just the UID field
lookup from conn on UID
Project the orig_bytes, resp_bytes, and duration fields out of the conn object into the top level and drop the conn object.
From there, run a group on null (to group all the documents). Then run sum and avg commands similar to these: https://github.com/ocmdev/rita/blob/master/analysis/structure/uconn.go#L61

Maybe a better way to do it

carrohan: Fixed!

joelillo:
Ready to merge?

PR #155, opened 2017-12-12
Allow installing from non-master RITA codebases

ethack:


SamuelCarroll:
The installer works very well and will allow the installation of a modified source code. Should be good to merge

Zalgo2462:
Subsumed by #156 and #157

PR #169, opened 2018-01-16
Add Go Dep to the master branch

Zalgo2462:


install.sh
```diff
@@ -320,6 +320,9 @@ __install() {
 	__install_mongodb & __load "[+] Installing MongoDB"
 
 	( # Build RITA
+		# Ensure go dep is installed
+		go get -u github.com/golang/dep/cmd/dep
```
ethack: Dep's readme recommends installing from a versioned release for stability. We'd be versioning our version manager too which seems like a good idea. Thoughts on downloading it this way instead?

```
wget -nv -O $GOPATH/bin/dep https://github.com/golang/dep/releases/download/v0.3.2/dep-linux-amd64
chmod +x $GOPATH/bin/dep
```

Zalgo2462: awesome idea

Zalgo2462:
I need to fix the dockerfile in this pr before this goes through

install.sh
SamuelCarroll: Looks great to me too!

Zalgo2462:
Dockerfile fixed and dep version pinned.

PR #171, opened 2018-01-23
Versioning Update

Zalgo2462:
Added better support for semver in RITA. RITA now stores two version tags, one with the latest tag (latest semver version), and one with exact commit details (ExactVersion). Additionally, RITA will check for incompatibilities at parse time and import time. If RITA attempts to import data into an unanalyzed, incompatible database, it will log out an error (PS we should probably add a log hook for errors to get get printed to the console). Further more, if RITA attempts to analyze a dataset imported with an incompatible version of RITA, the software will warn the user, and skip the analysis.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,9 @@
-VERSION := $(shell git describe --always --long --dirty --tags)
+VERSION := $(shell git describe --abbrev=0 --tags)
+EXACT_VERSION := $(shell git describe --always --long --dirty --tags)
 GOPATH := $(GOPATH)
 BINARY := rita
 
-LDFLAGS=-ldflags="-X github.com/ocmdev/rita/config.VERSION=${VERSION}"
+LDFLAGS=-ldflags="-X github.com/ocmdev/rita/config.Version=${VERSION} -X github.com/ocmdev/rita/config.ExactVersion=${EXACT_VERSION}"
 
 
 default:
@@ -14,4 +15,3 @@ default:
 install:
 	dep ensure
 	go build ${LDFLAGS} -o ${GOPATH}/bin/${BINARY}
-
--- a/commands/analyze.go
+++ b/commands/analyze.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/blang/semver"
 	"github.com/ocmdev/rita/analysis/beacon"
 	"github.com/ocmdev/rita/analysis/blacklist"
 	"github.com/ocmdev/rita/analysis/crossref"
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -1,8 +1,9 @@
-VERSION := $(shell git describe --always --long --dirty --tags)
+VERSION := $(shell git describe --abbrev=0 --tags)
+EXACT_VERSION := $(shell git describe --always --long --dirty --tags)
 GOPATH := $(GOPATH)
 BINARY := rita
 
-LDFLAGS=-ldflags="-X github.com/ocmdev/rita/config.VERSION=${VERSION}"
+LDFLAGS=-ldflags="-X github.com/ocmdev/rita/config.Version=${VERSION} -X github.com/ocmdev/rita/config.ExactVersion=${EXACT_VERSION}"
```
Zalgo2462: VERSION contains the semver
EXACT_VERSION is a full git descriptor for debugging

commands/analyze.go
```diff
@@ -36,26 +37,46 @@ func init() {
...
 	if inDb == "" {
 		res.Log.Info("Running analysis against all databases")
-		toRun = append(toRun, res.MetaDB.GetUnAnalyzedDatabases()...)
+		toRunDirty = append(toRun, res.MetaDB.GetUnAnalyzedDatabases()...)
 	} else {
-		info, err := res.MetaDB.GetDBMetaInfo(inDb)
+		toRunDirty = append(toRun, inDb)
+	}
+
+	// Check for problems
```
Zalgo2462: new logic to check if we should remove a queued db

database/meta.go
```diff
@@ -22,11 +23,12 @@ type (
...
-		Name       string        `bson:"name"`          // Top level name of the database
-		Analyzed   bool          `bson:"analyzed"`      // Has this database been analyzed
-		UsingDates bool          `bson:"dates"`         // Whether this db was created with dates enabled
-		Version    string        `bson:"version"`       // Rita version at import
+		ID             bson.ObjectId `bson:"_id,omitempty"`   // Ident
+		Name           string        `bson:"name"`            // Top level name of the database
+		Analyzed       bool          `bson:"analyzed"`        // Has this database been analyzed
+		UsingDates     bool          `bson:"dates"`           // Whether this db was created with dates enabled
+		ImportVersion  string        `bson:"import_version"`  // Rita version at import
+		AnalyzeVersion string        `bson:"analyze_version"` // Rita version at analyze
```
Zalgo2462: split version tags at analyze and import times

database/meta.go
```diff
@@ -40,10 +42,10 @@ func (m *MetaDBHandle) AddNewDB(name string) error {
...
 	err := ssn.DB(m.DB).C(m.res.Config.T.Meta.DatabasesTable).Insert(
 		DBMetaInfo{
-			Name:       name,
-			Analyzed:   false,
-			UsingDates: m.res.Config.S.Bro.UseDates,
-			Version:    m.res.Config.R.Version,
+			Name:          name,
+			Analyzed:      false,
+			UsingDates:    m.res.Config.S.Bro.UseDates,
+			ImportVersion: m.res.Config.S.Version,
```
Zalgo2462: R: Running config contains a semver object
S: Static config contains a string

database/meta.go
```diff
@@ -128,8 +130,20 @@ func (m *MetaDBHandle) MarkDBAnalyzed(name string, complete bool) error {
 		return err
 	}
 
+	var versionTag string
+	if complete {
+		versionTag = m.res.Config.S.Version
+	} else {
+		versionTag = ""
+	}
```
Zalgo2462: Same method handles setting the db analyzed flag true and false, dependent on the complete variable. if !complete, remove the tag

PR #173, opened 2018-01-25
Adding updated ranges for expected Beacon test results

SamuelCarroll:
Changed because we changed the way we evaluate beacons, may need to find new tests to fill in the ones that we've changed

To run the tests in the Rita project folder type `test ./...`


ethack:
@SamuelCarroll Could you add a new section in https://github.com/ocmdev/rita/blob/dev/Contributing.md that explains there are tests, show any dependencies needed to run them (I don't think there are any?), and then the command to run the tests?

SamuelCarroll:
Absolutely!

PR #178, opened 2018-02-05
Add the ability to split logs into dbs by subfolder, clean up the cli

Zalgo2462:
1.) Clean up the cli so you don't have to specify -d for commands which take a database as an argument
2.) Remove the Bro DirectoryMap (all import mapping is done on the cli)
3.) Change the -i and -d to positional arguments in import
4.) Add a --split option on import for splitting by subfolder and splitting by date

Splitting by subfolder appends the subfolder a bro log is found in to the root db name. This can be used with Bro's default file structure to import to DBNameBase-YYYY-MM-DD. Splitting by date tries to infer the dates when each connection is logged and appends that date to the resultant db. (i.e. the current UseDates option)

### DIFF
--- a/commands/import.go
+++ b/commands/import.go
@@ -2,7 +2,10 @@ package commands
 
 import (
 	"fmt"
+	"os"
+	"strings"
 
+	"github.com/ocmdev/rita/config"
 	"github.com/ocmdev/rita/database"
 	"github.com/ocmdev/rita/parser"
 	"github.com/ocmdev/rita/util"
@@ -11,19 +14,15 @@ import (
 
 func init() {
 	importCommand := cli.Command{
-		Name:  "import",
-		Usage: "Import bro logs into the database",
+		Name:      "import",
+		Usage:     "Import bro logs into a database",
+		ArgsUsage: "<directory to import> <target database>",
 		Flags: []cli.Flag{
 			threadFlag,
 			configFlag,
 			cli.StringFlag{
-				Name:  "import-dir, i",
-				Usage: "Import bro logs from a `directory` into a database. This overides the config file. Must be used with --database, -d",
-				Value: "",
-			},
-			cli.StringFlag{
-				Name:  "database, d",
-				Usage: "Store imported bro logs into a database with the given `name`. This overides the config file. Must be used with --import-dir, -i",
+				Name:  "split, s",
+				Usage: "Split the imported bro logs. Accepted values: \"subfolder\", \"date\"",
 				Value: "",
 			},
 		},
@@ -36,33 +35,35 @@ func init() {
 // doImport runs the importer
 func doImport(c *cli.Context) error {
 	res := database.InitResources(c.String("config"))
-	importDir := c.String("import-dir")
-	databaseName := c.String("database")
+	importDir := c.Args().Get(0)
+	targetDatabase := c.Args().Get(1)
+	splitStrategy := strings.ToLower(c.String("split"))
 	threads := util.Max(c.Int("threads")/2, 1)
 
-	//one flag was set
-	if importDir != "" && databaseName == "" || importDir == "" && databaseName != "" {
-		return cli.NewExitError(
-			"Import failed.\nUse 'rita import' to import the directories "+
-				"specified in the config file or 'rita import -i [import-dir] -d [database-name]' "+
-				"to import bro logs from a given directory.", -1)
+	if importDir == "" || targetDatabase == "" {
+		return cli.NewExitError("Specify a directory to import and a target database", -1)
 	}
 
-	//both flags were set
-	if importDir != "" && databaseName != "" {
-		res.Config.S.Bro.LogPath = importDir
-		res.Config.S.Bro.DBPrefix = ""
-		//Clear out the directory map and set the default database
-		res.Config.S.Bro.DirectoryMap = make(map[string]string)
-		res.Config.S.Bro.DefaultDatabase = databaseName
+	//Remove tailing / when applicable
+	if strings.HasSuffix(importDir, string(os.PathSeparator)) {
+		importDir = importDir[:len(importDir)-len(string(os.PathSeparator))]
 	}
 
-	res.Log.Infof("Importing %s\n", res.Config.S.Bro.LogPath)
-	fmt.Println("[+] Importing " + res.Config.S.Bro.LogPath)
+	res.Config.R.Bro.ImportDirectory = importDir
+	res.Config.R.Bro.TargetDatabase = targetDatabase
+	res.Config.R.Bro.SplitStrategy = config.SplitNone
+	if splitStrategy == "subfolder" {
+		res.Config.R.Bro.SplitStrategy = config.SplitSubfolder
+	} else if splitStrategy == "date" {
+		res.Config.R.Bro.SplitStrategy = config.SplitDate
+	}
+
+	res.Log.Infof("Importing %s\n", res.Config.R.Bro.ImportDirectory)
+	fmt.Println("[+] Importing " + res.Config.R.Bro.ImportDirectory)
 	importer := parser.NewFSImporter(res, threads, threads)
 	datastore := parser.NewMongoDatastore(res.DB.Session, res.MetaDB,
 		res.Config.S.Bro.ImportBuffer, res.Log)
 	importer.Run(datastore)
-	res.Log.Infof("Finished importing %s\n", res.Config.S.Bro.LogPath)
+	res.Log.Infof("Finished importing %s\n", res.Config.R.Bro.ImportDirectory)
 	return nil
 }
--- a/commands/import.go
+++ b/commands/import.go
@@ -3,9 +3,9 @@ package commands
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
-	"github.com/ocmdev/rita/config"
 	"github.com/ocmdev/rita/database"
 	"github.com/ocmdev/rita/parser"
 	"github.com/ocmdev/rita/util"
@@ -14,17 +14,17 @@ import (
 
 func init() {
 	importCommand := cli.Command{
-		Name:      "import",
-		Usage:     "Import bro logs into a database",
-		ArgsUsage: "<directory to import> <target database>",
+		Name:  "import",
+		Usage: "Import bro logs into a target database",
+		UsageText: "rita import [command options] [<import directory> <database root>]\n\n" +
+			"Logs directly in <import directory> will be imported into a database" +
+			" named <database root>. Files in a subfolder of <import directory> will be imported" +
+			" into <database root>-$SUBFOLDER_NAME. <import directory>" +
+			" and <database root> will be loaded from the configuration file unless" +
+			" BOTH arguments are supplied.",
 		Flags: []cli.Flag{
 			threadFlag,
 			configFlag,
-			cli.StringFlag{
-				Name:  "split, s",
-				Usage: "Split the imported bro logs. Accepted values: \"subfolder\", \"date\"",
-				Value: "",
-			},
 		},
 		Action: doImport,
 	}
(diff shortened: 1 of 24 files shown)


ethack:
The pull request satisfies these requirements (I know we never had official requirements written down before now so this is just to try and keep everything straight):

- [x]  Allow users to import new data without editing a config file
- [ ]  Fix or remove buggy UseDates feature
- [x]  Fix or remove buggy DefaultDatabase feature

For the UseDates feature, I think I'm ok getting rid of it. I think splitting multi-day pcap or Bro files is an uncommon enough case that if we run across it again we can come up with a workable solution later. And in the meantime it will remove complexity from RITA. For future reference, here are several tools to look into if we need pcap splitting features (e.g. session, size, time). The first two can split by time.
* https://www.wireshark.org/docs/man-pages/editcap.html
* `tcpdump -G` (rotate by seconds)
* `tcpdump -C` (rotate by size)
* https://github.com/seladb/PcapPlusPlus/tree/master/Examples/PcapSplitter
* http://www.netresec.com/?page=SplitCap

Thoughts on the current PR:
1. It's not clear that the "database" parameter is used for the database prefix with the `--split` option.
2. Removing LogPath and DBPrefix from the config file means the user always has to specify these even if they never change.
3. Removing LogPath and DBPrefix means that when the agent service is installed these parameters are part of that service definition. If they ever need to be changed, the service file needs to be manually edited or the agent needs to be uninstalled/reinstalled instead of editing the config file.
4. If you typo the DBPrefix parameter you would end up re-importing all logs again. This is also possible via config file but seems more likely via CLI if you're forced to type it out every time you run it manually.
5. When using `--split subfolder` nested subfolders are put into different databases as well. The user can avoid this by not splitting the import and just sending it to a single database. I just wanted to note it so we can start documenting the idiosyncrasies of the importer.

Suggestions:
1. Keep the LogPath and DBPrefix options in the config file.
2. Allow command line params to override the LogPath and DBPrefix config values.
3. Don't re-use the database positional parameter to both specify a one-off database name and DBPrefix. Make a specific parameter for DBPrefix.
4. When displaying the help text, show the default values for the LogPath and DBPrefix (grab from the config file)
5. Make the default behavior with no positional arguments split by subfolder. It should get the values for LogPath and DBPrefix using the config file or command line params. This is more similar to the current behavior since you can potentially configure everything you need in the config file and then run `rita import; rita analyze`. It has the benefit of keeping a similar workflow and simplifies what users have to remember to type in.
6. Keep the functionality of specifying a one off import. It's already there, I just want to make it clear it should stay.

Here's how I envision the new help text to look:

```
NAME:
   rita import - Import bro logs into a database. If no database is specified, logs will be split by subfolder.

USAGE:
   rita import [command options] [database]

OPTIONS:
   --threads N, -t N                     Use N threads when executing this command (default: 4)
   --config CONFIG_FILE, -c CONFIG_FILE  Use a given CONFIG_FILE when running this command
   --logpath LOG_PATH, -l LOG_PATH       Import logs from the following path. (default: (value from config file's LogPath))
   --dbprefix PREFIX, -p PREFIX          Use PREFIX on each database name. Has no effect if a database is specified. (default: (value from config file's DBPrefix))
```

Examples:

```
# Imports logs from the config file's LogPath into different databases based on subfolders
rita import

# Imports all new logs from the config file's LogPath into "mydatabase".
rita import mydatabase

# Imports logs from /root/meterpreter-bro/ into "meterpreter"
rita import --logpath /root/meterpreter-bro/ meterpreter

# Imports logs from /root/meterpreter-bro/ into "meterpreter-bro" (Uses the parent directory name if no subdirectories exist and database name isn't specified)
rita import --logpath /root/meterpreter-bro/

# Imports all new logs from /opt/bro/logs/ into different databases based on subfolders
rita import --logpath /opt/bro/logs/

# Imports logs from /root/incident into different databases based on subfolders with the prefix of "incident1234-"
rita import --logpath /root/incident/ --dbprefix incident1234-

# (With no LogPath set or LogPath invalid) Should display an error that LogPath isn't valid.
rita import
```

@SamuelCarroll @Zalgo2462 Thoughts?

Zalgo2462:
## UseDates
I would love to get rid of this feature. However, I've looked at all of the tools listed and none of them can split a pcap into days automatically. The best bet would be using editcap repeatedly to extract the day by day pcaps by specifying start and end timestamps. If this workflow will work for RITA users, I believe we can remove the Use Dates feature.

## Thoughts on current PR
> It's not clear that the "database" parameter is used for the database prefix with the --split option.

This could be added to the help description for the split parameter.

> Removing LogPath and DBPrefix from the config file means the user always has to specify these even if they never change.

My personal taste leads me to prefer this as it is simpler. We have recently felt the push to make sure the user never touches the configuration file. If we remove LogPath and DBPrefix from the config file, we make RITA a bit more transparent without requiring the user to alter the config file.

> Removing LogPath and DBPrefix means that when the agent service is installed these parameters are part of that service definition. If they ever need to be changed, the service file needs to be manually edited or the agent needs to be uninstalled/reinstalled instead of editing the config file.

I'll have to look back at the agent. Aren't we currently specifying this manually anyways to avoid the current usedates issue? However, if we switch to subdirectory importing, we could completely rely on the config file which is a big win. This leads me to agree with you. We should probably put these back in the config file.

> If you typo the DBPrefix parameter you would end up re-importing all logs again. This is also possible via config file but seems more likely via CLI if you're forced to type it out every time you run it manually.

Hopefully you would catch your error, cancel the import, delete the database, and re-import.

>When using --split subfolder nested subfolders are put into different databases as well. The user can avoid this by not splitting the import and just sending it to a single database. I just wanted to note it so we can start documenting the idiosyncrasies of the importer.

YES! Sorry I didn't mention this in the PR.

## Suggestions

> Keep the LogPath and DBPrefix options in the config file.

If we put these back in, should we put an option for whether or not to split the logs? Should we make splitting by subdirectory the only way to import bro logs? (i.e. force the user to move the logs to the top level if thats what they want)

>Allow command line params to override the LogPath and DBPrefix config values.

I, personally, find this confusing. However, I do value the utility. Thoughts? I suppose it could always be removed if users find it confusing.

> Don't re-use the database positional parameter to both specify a one-off database name and DBPrefix. Make a specific parameter for DBPrefix.

I would rather keep them the same parameter. My favorite idea to deal with this is to make splitting by subdirectory the only way to import bro logs and rename the parameter to DBPrefix for the import command.

> When displaying the help text, show the default values for the LogPath and DBPrefix (grab from the config file)

This would require a re-architecturing of RITA. Currently we only load the config file after a command has been run. Additionally, you can specify a config file with the `-c` parameter, which presents data dependency issues.

> Make the default behavior with no positional arguments split by subfolder. It should get the values for LogPath and DBPrefix using the config file or command line params. This is more similar to the current behavior since you can potentially configure everything you need in the config file and then run rita import; rita analyze. It has the benefit of keeping a similar workflow and simplifies what users have to remember to type in.

I argue that we could make splitting by subfolder the only way to import.

> Keep the functionality of specifying a one off import. It's already there, I just want to make it clear it should stay.

SamuelCarroll:
##UseDates
I can't say I've ever used this feature. I'm fine removing it and it sounds like that's the direction we're going

##Thoughts on current PR
> Removing LogPath and DBPrefix from the config file means the user always has to specify these even if they never change.

> My personal taste leads me to prefer this as it is simpler. We have recently felt the push to make sure the user never touches the configuration file. If we remove LogPath and DBPrefix from the config file, we make RITA a bit more transparent without requiring the user to alter the config file.

My view on this is that when the user runs install.sh that we prompt for the LogPath and DBPrefix, then the script will fill this information in. At that point we can use the --split subfolder to get the subfolders from LogPath and we auto use DBPrefix the old way too
Also I think since install.sh will install Bro we should set what the auto-rotation is (if it's set to greater than 1 day)

> Removing LogPath and DBPrefix means that when the agent service is installed these parameters are part of that service definition. If they ever need to be changed, the service file needs to be manually edited or the agent needs to be uninstalled/reinstalled instead of editing the config file.

> I'll have to look back at the agent. Aren't we currently specifying this manually anyways to avoid the current usedates issue? However, if we switch to subdirectory importing, we could completely rely on the config file which is a big win. This leads me to agree with you. We should probably put these back in the config file.

The way `agent install` handles this is to get the Rita path and just run `{{.RitaPath}} import --config {{.ConfigPath}} && {{.RitaPath}} analyze --config {{.ConfigPath}}`

> If you typo the DBPrefix parameter you would end up re-importing all logs again. This is also possible via config file but seems more likely via CLI if you're forced to type it out every time you run it manually.

> Hopefully you would catch your error, cancel the import, delete the database, and re-import.

I believe to depend on a hopefully will put strain on a end-user. This is why we should prompt for the LogPath and DBPrefix in the installer and auto-populate the config file from there


##Suggestions
> Keep the LogPath and DBPrefix options in the config file.

> If we put these back in, should we put an option for whether or not to split the logs? Should we make splitting by subdirectory the only way to import bro logs? (i.e. force the user to move the logs to the top level if thats what they want)

I would say that we make split by subdirectory the only way to import bro logs

> Allow command line params to override the LogPath and DBPrefix config values.

> I, personally, find this confusing. However, I do value the utility. Thoughts? I suppose it could always be removed if users find it confusing.

I value utility here. If the user wants to dig into Rita to find these they can, but it's a documented feature not something a casual user will need to know

> When displaying the help text, show the default values for the LogPath and DBPrefix (grab from the config file)

> This would require a re-architecturing of RITA. Currently we only load the config file after a command has been run. Additionally, you can specify a config file with the -c parameter, which presents data dependency issues.

I think this is done on the first line of doImport in import.go we can access the value using res.System.BroConfig.DBPrefix and res.System.BroConfig.LogPath (since both are strings)

> Make the default behavior with no positional arguments split by subfolder. It should get the values for LogPath and DBPrefix using the config file or command line params. This is more similar to the current behavior since you can potentially configure everything you need in the config file and then run rita import; rita analyze. It has the benefit of keeping a similar workflow and simplifies what users have to remember to type in.

> I argue that we could make splitting by subfolder the only way to import.

> Keep the functionality of specifying a one off import. It's already there, I just want to make it clear it should stay.

I think the user specifying a one-off log to import and analyze may still be something the user *could* want to do.

ethack:
After offline discussion, the current plan is:

Config file options to remove:
* UseDates
* DirectoryMap
* DefaultDatabase

Config file options to keep:
* LogPath
* DBPrefix

The new import command will work like this.  `rita import` will read the LogPath and DBPrefix from the config file and automatically import each subdirectory of LogPath (1 level deep) into its own database. The database names will be `<DBPrefix>-<LogPath subdir name>`.

Instead of using the config file you can run `rita import /path/to/logs/ databasename`.  Which will essentially do the same as above but replace the values LogPath=/path/to/logs/ and DBPrefix=databasename.

If there are valid Bro logs at the top level directory (not inside a subdir) then the database name will just be `databasename` (i.e. whatever is specified on the command line).

```
NAME:
   rita import - Import bro logs into databases based on subdirectories of the directory supplied. If <directory> and <database> are not specified, they are read from the config file from "LogPath" and "DBPrefix", respectively.

USAGE:
   rita import [command options] [<directory> <database>]

OPTIONS:
   --threads N, -t N                     Use N threads when executing this command (default: 4)
   --config CONFIG_FILE, -c CONFIG_FILE  Use a given CONFIG_FILE when running this command
```

Examples:

```
# Imports logs from the config file's LogPath into different databases based on subfolders
rita import

# Imports logs from /root/meterpreter-bro/ into "meterpreter" (assuming no subdirs)
rita import /root/meterpreter-bro/ meterpreter

# Error message requiring both parameters
rita import /root/meterpreter-bro/

# Error message requiring both parameters
rita import meterpreter

# Imports logs from /root/incident into different databases based on subdirs with the prefix of "incident1234-". Any log files in the top level will be in a database named "incident1234"
rita import /root/incident/ incident1234

# (With no LogPath/DBPrefix set or LogPath/DBPrefix invalid) Error message explaining which config values are invalid or unset
rita import
```

Zalgo2462:
Changes:
- Split long command descriptions (analyze; html-report)
- Add description to import command
- Remove split option on import
- (Bug) Caught issue with subfolder import if the root directory was given as a relative path
- Remove BroSplitStrategy
- Added a check to remove a trailing "/" from theimport directory if given in the configuration file
- Removed UsingDates from the MetaDB
- Removed debug prints from MetaDB handler
- Removed logic to remove file records if they were imported using dates (This new version of rita will incorrectly delete old databases if they were imported using the UseDates feature)
- Rename LogPath to ImportDirectory
- Rename DBPrefix to DBRoot
- Removed Dates field from file records in MetaDB

ethack:
> * Added a check to remove a trailing "/" from the import directory if given in the configuration file

This doesn't seem to work for me. I added a trailing slash in the config file and the ImportDirectory was "" in the `test-config` output and `import` didn't pick up any files.  A trailing slash on the command line imported correctly.

commands/import.go
```diff
@@ -36,33 +35,40 @@ func init() {
...
+	//check if the user overrode the config file
+	if importDir != "" && targetDatabase != "" {
+		//expand relative path
+		//nolint: vetshadow
+		importDir, err := filepath.Abs(importDir)
+		if err != nil {
+			return cli.NewExitError(err.Error(), -1)
+		}
+
+		//Remove tailing / when applicable
```
ethack: I think this should be outside the `if` block. This currently only gets executed when the config file is overridden.

ethack: Nvm, I found where this is modifying the config value in `config/static.go`. The bug is there.

commands/import.go
```diff
@@ -36,33 +33,35 @@ func init() {
...
-		res.Config.S.Bro.LogPath = importDir
-		res.Config.S.Bro.DBPrefix = ""
-		//Clear out the directory map and set the default database
-		res.Config.S.Bro.DirectoryMap = make(map[string]string)
-		res.Config.S.Bro.DefaultDatabase = databaseName
+	//check if the user overrode the config file
+	if importDir != "" && targetDatabase != "" {
+		//expand relative path
+		//nolint: vetshadow
+		importDir, err := filepath.Abs(importDir)
```
ethack: `filepath.Abs` automatically calls `Clean` so it's not necessary to remove the trailing slash manually.

config/static.go
```diff
@@ -118,6 +116,11 @@ func loadStaticConfig(cfgPath string) (*StaticCfg, error) {
 	// set the socket time out in hours
 	config.MongoDB.SocketTimeout *= time.Hour
 
+	// clean all filepaths
+	config.Log.RitaLogPath = filepath.Clean(config.Log.RitaLogPath)
+	config.Blacklisted.SafeBrowsing.Database = filepath.Clean(config.Blacklisted.SafeBrowsing.Database)
+	config.Bro.ImportDirectory = filepath.Clean(config.Bro.ImportDirectory)
```
ethack: `filepath.Clean` seems to be a good one to use here. It removes the trailing slash and takes care of some other possible filepath cruft. I went ahead and ran it on the other config values that contain filepaths.

ethack:
This now looks good to me. I made a couple changes and it's a bigger change so another set of eyes would be nice.
@SamuelCarroll @joelillo

SamuelCarroll:
This looks good to me too!

I did find another problem that's unrelated to this issue. In install.sh on lines 137 and 145 we call a function _err. However it looks like we define the function as __err on line 12, should I go ahead and add a hotfix?

PR #182, opened 2018-02-08
New install script

SamuelCarroll:
This install script prompts the user to supply values for Bro Log paths, Mongodb connection string, Database Prefix, MetaDatabase Name, and Google API Key and then it will fill those values in in the config file


joelillo:
Code reviewed and ran this a few times. Looking good!

PR #183, opened 2018-02-09
Updated default install and config locations

ethack:
* Change config file location to `/etc/rita/`
* Change the rita install location to `/opt/rita`
* Adding additional command line parameters for script and developer flexibility. These are meant to be undocumented so I didn't update the rita help text.
  * Disable bro installation
  * Disable mongo installation
  * Change the install location

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,9 +1,7 @@
 package config
 
 import (
-	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


SamuelCarroll:
This looks good to me, I'd say we can merge it in!

PR #185, opened 2018-02-10
WIP: Allow for paths relative to rita executable

ethack:
This needs to be sanity checked, tested, and code reviewed. There are probably better/cleaner ways of doing things.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,7 +2,9 @@ package config
 
 import (
 	"os"
+	"fmt"
 	"reflect"
+	"path/filepath"
 )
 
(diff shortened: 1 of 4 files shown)


Zalgo2462:
After working with the current install location, I've come to realize that fixed paths are better. That way I can still run the application from my GOPATH/bin while developing rita.

SamuelCarroll:
I too like using absolute paths. However I can see users wanting to use relative paths, can we check if the relative path file exists and if it doesn't try using the absolute path (if ../etc/rita/config.yaml doesn't exist we check if /etc/rita/config does exist, before an error)

PR #186, opened 2018-02-14
Major restructure of the installer

ethack:
## Major Changes
* Added ability to detect and install a precompiled RITA binary. This will allow shipping a distribution that contains a RITA binary and will avoid having to install the entire Go build environment. This was my main goal before I got carried away 😄 The new install algorithm is:
  1. Use precompiled `rita` and necessary config files in installer's directory, if available.
  2. Compile rita from source using local code in installer's directory, if available.  (This was the previous default behavior.)
  3. Compile rita from source using remote code from Github. This will currently cause an error because it defaults to the master branch which doesn't have `tables.yaml`.
* Removed uninstaller. This was difficult to keep updated between versions and one uninstaller couldn't uninstall a different rita version if paths changed without getting complex. Replaced with reinstall feature.
* Reinstall feature overwrites previous files and backs up the user's `config.yaml` file first.

## Minor Changes
* Added convenience option for developers to force building rita from source.
* Documented `--disable-bro` and `--disable-mongo` flags.
* Restructured progress output.
* Replaced all uses of `wget` with equivalent `curl` commands to remove wget dependency.
* Fixed bug with using `--prefix` with a relative path.
* Fixed bug if you ran the installer from rita's GOPATH folder.

## Example Usage
```
# Install everything on a blank system, just like before
sudo ./install.sh
# Same thing as above but you already have a previous rita install in the same location. 
# It will prompt you to reinstall or exit, or you can force reinstalling which skips the prompt.
sudo ./install.sh --reinstall

# Output
[+] Ensuring curl, coreutils, realpath, and lsb-release are installed... SUCCESS
[+] Installing Bro IDS... SUCCESS
[+] Installing MongoDB... SUCCESS
[+] Using prebuilt RITA binary.
[+] Backing up configuration to /etc/rita/config.yaml.old... SUCCESS
[+] Installing RITA binary and config files... SUCCESS
```

```
# If you have a rita binary in the installer directory (e.g. from development work) you can force it to build from source instead
sudo ./install.sh --build

# Output
[+] Ensuring curl, coreutils, realpath, and lsb-release are installed... SUCCESS
[+] Installing Bro IDS... SUCCESS
[+] Installing MongoDB... SUCCESS
[+] Installing build environment... SUCCESS
[+] Using locally available RITA source... SUCCESS
[+] Building RITA... SUCCESS
[+] Backing up configuration to /etc/rita/config.yaml.old... SUCCESS
[+] Installing RITA binary and config files... SUCCESS
```

```
# You may also want to disable installing mongo and bro to speed up development and testing or if you have these programs already (these options existed, but were not documented)
sudo ./install.sh --disable-bro --disable-mongo
```

Still to do:

- [ ] Test on CentOS.
- [ ] Make sure Mongo installs.
- [ ] Make sure Bro installs.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -44,12 +44,18 @@ __help() {
 	__title
 
 	cat <<HEREDOC
+This script automatically installs Real Intelligence Threat Analyitics (RITA)
+along with necessary dependencies, including Bro IDS and MongoDB.
+
 Usage:
 	${_NAME} [<arguments>]
 
 Options:
-	-h --help			Show this help message.
-	-u --uninstall			Remove RITA.
+	-h|--help			Show this help message.
+	-r|--reinstall			Force reinstalling RITA.
+	-b|--build			Force building RITA from source code.
+	--disable-bro			Disable automatic installation of Bro IDS.
+	--disable-mongo			Disable automatic installation of MongoDB.
 
 HEREDOC
 }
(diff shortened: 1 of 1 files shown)


ethack:
Most of this PR is obsolete or migrated to #197 

I still think this functionality is worthwhile to migrate.  But before that we should have a discussion if this is the best way to simplify deployments.  Also, I'm working on using Docker to compile a portable binary which would ease creation of releases.  I'd like to complete that as a proof of concept first as well.

> Added ability to detect and install a precompiled RITA binary. This will allow shipping a distribution that contains a RITA binary and will avoid having to install the entire Go build environment. The new install algorithm is:
> 1. Use precompiled rita and necessary config files in installer's directory, if available.
> 2. Compile rita from source using local code in installer's directory, if available. (This was the previous default behavior.)
> 3. Compile rita from source using remote code from Github.

PR #187, opened 2018-02-21
Move Complex Logic from Import into Analysis [Supports bro-rita plugin]

Zalgo2462:
Currently, http records are preprocessed as they are imported in order to clean up request uri's which are absolute urls. This cleaning is important in order to obtain accurate results in the long urls portion of analysis. However, this cleaning relies on go specific libraries. In order to support the bro-rita plugin as a drop in replacement for rita's parser, this logic must be moved to the analysis step. This PR adds a "sanitization" step to analysis, in which offending records are found and updated.

Additionally, this PR changes DNS.RTT's mongo tag to "rtt" from "RTT". This change matches Bro's name for the field. 

With this PR, bro-rita will become a viable drop in replacement for rita's existing parser.

### DIFF
--- a/analysis/sanitization/sanitization.go
+++ b/analysis/sanitization/sanitization.go
@@ -1,0 +1,101 @@
+package sanitization
+
+import (
+	"net/url"
+	"strings"
+
+	mgo "gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+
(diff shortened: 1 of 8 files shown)


SamuelCarroll:
I didn't do any 'rigorous' testing, but it appears that rita function works. Looks good

Zalgo2462:
Tested. LGTM. Merging.

PR #195, opened 2018-03-09
Active CM Rename Part 1

Zalgo2462:
This PR makes the needed changes to move to the new repo https://github.com/activecm/rita from the old repo https://github.com/ocmdev/rita. This is part 1 which depends on the "activecm" branch of rita-bl. Once rita-bl is updated, part 2 will point the dependency back towards the master branch.


Gopkg.toml
```diff
@@ -21,16 +21,16 @@
...
 
 [[constraint]]
-  name = "github.com/ocmdev/mgosec"
+  name = "github.com/activecm/mgosec"
   version = "0.1.0"
 
 [[constraint]]
-  branch = "master"
-  name = "github.com/ocmdev/rita-bl"
+  branch = "activecm"
```
Zalgo2462: This is why this is part 1

Zalgo2462:
I removed the docker installation details due to an impending migration away from docker hub.

Issue #194, opened 2018-03-06
Installer Puts Root Owned Files in Existing GOPATH


Zalgo2462:
If the installer detects an existing GOPATH, it will use it. However, every file the installer writes into the GOPATH will be owned by root. This may create a mess for the user to clean up.

ethack:
Fixed with #197

Issue #192, opened 2018-03-06
Installer Breaks if INSTALL_PREFIX or CONFIG_PATH


Zalgo2462:
If either _INSTALL_PREFIX or _CONFIG_PATH has spaces, the installer will break. Every instance of these variables must be quoted as they are paths that come from the user.

Examples:
Install Prefix: https://github.com/activecm/rita/blob/02b69e10c80518e7c9086fe75bd45e5f904d093e/install.sh#L237

Config Path: https://github.com/activecm/rita/blob/02b69e10c80518e7c9086fe75bd45e5f904d093e/install.sh#L351

ethack:
Fixed with #197

Issue #191, opened 2018-03-06
RITA Installer Does Not Follow the Filesystem Hierarchy Standard


Zalgo2462:
The [Filesystem Hierarchy Standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) is a standard defined by the Linux Foundation as guidance on how to organize files on UNIX-like systems. 

If RITA were to follow the FHS, several files would need to be moved.

Assuming an opt install.
```
GOPATH/bin/rita [the executable] -> /opt/rita/bin/rita
LICENSE -> /opt/rita/LICENSE
config.yaml -> /etc/opt/rita/config.yaml
tables.yaml -> /etc/opt/rita/tables.yaml
safebrowsing -> /var/opt/rita/lib/safebrowsing
RITALogPath -> /var/opt/rita/log/
``` 

Assuming a /usr/local/bin type install
```
GOPATH/bin/rita [the executable] -> /usr/local/bin/rita
LICENSE -> /etc/rita/LICENSE
config.yaml -> /etc/rita/config.yaml
tables.yaml -> /etc/rita/tables.yaml
safebrowsing -> /var/lib/rita/safebrowsing
RITALogPath -> /var/lib/rita/log/
``` 
http://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.pdf

ethack:
Fixed with #197

Issue #190, opened 2018-03-05
Go installation creates separate GOPATHs. (Odd Go Installation)


Zalgo2462:
In the rita installer, go is installed to /usr/local/go if it is not detected. Currently a GOPATH is created in /opt/rita, and, for the duration of the script, this directory is set to the $GOPATH environment variable. However, this environment variable is not added to the user's bashrc file as /opt/rita is not writable without administrator privileges. Despite this, /usr/local/go is added to the user's PATH variable in their bashrc. This means when a user runs go, the default GOPATH at $HOME/go is created. This GOPATH is divorced from the one used to install rita.

Administrator level GOPATH's aren't supported very well by go. Especially as sudo strips most environment variables. 

We can solve this problem in one of two ways while still providing a system wide installation.

Solution A: Make the GOPATH writable by everyone. This removes the need for sudo. However, this prevents using go in the traditional sense (one GOPATH per user). If we go this route, we may want to move the GOPATH to a place other than /opt/rita as setting the GOPATH to will create executables in /opt/rita/bin unrelated to RITA.

Solution B: Use the user's GOPATH and copy the built executable to the system wide location. This is more akin to standard "make install" programs.

I prefer solution B as it follows more industry standards than solution A.

Zalgo2462:
In lieu of #194 solution B is less attractive.

I'm still not a fan of Solution A however.

@ethack proposed another solution (Solution C): Building RITA in a docker container. This would provide reproducible builds and we wouldn't have to muck with GOPATHs at all. Additionally, RITA already has a functioning Dockerfile. However, that may pose its own issues (mainly around linking).

ethack:
Fixed with #197

PR #197, opened 2018-03-20
New Installer 

Zalgo2462:
When we moved to the system wide installs, we created a number of issues. This script is a rewrite of the current installer which aims to fix those issues.

Issues fixed:
https://github.com/activecm/rita/issues/194
https://github.com/activecm/rita/issues/192
https://github.com/activecm/rita/issues/191
https://github.com/activecm/rita/issues/190
https://github.com/activecm/rita/issues/188

Need to check:
https://github.com/activecm/rita/issues/189

The installer now installs rita to /usr/local/bin/rita or $_PREFIX/bin/rita, /var/lib/rita/, and /etc/rita/.
See https://github.com/activecm/rita/issues/191 for more details. (/usr/local/ type install).

While this is in PR, the most useful way to call the script is "./install.sh --version dev"


install.sh
```diff
@@ -3,123 +3,404 @@
...
+	_INSTALL_PREFIX=/usr/local
+	_CONFIG_PATH=/etc/rita
+	_VAR_PATH=/var/lib/rita
 
-# Clean up our hacks
-set +o posix
-set +i
-PS1=$_OLD_PS1
-unset _OLD_PS1
+	# FOR an OPT style installation
```
Zalgo2462: If we want to move to an opt style installation, the path variables can be configured here.

install.sh
```diff
@@ -3,123 +3,404 @@
...
+_IMPORTANT="[!]"
+_QUESTION="[?]"
+_SUBITEM="\t$_ITEM"
+_SUBIMPORTANT="\t$_IMPORTANT"
+_SUBQUESTION="\t$_QUESTION"
 
-#Error handling
-#Kill 0 to kill subshells as well
+
+# ERROR HANDLING
```
Zalgo2462: Print out the line number if the script fails for error reporting.

install.sh
```diff
@@ -3,123 +3,404 @@
...
+			--disable-mongo)
+				_INSTALL_MONGO=false
+				;;
+			--prefix)
+				shift
+				_INSTALL_PREFIX="$1"
+				;;
+			--version)
+				shift
+				_RITA_VERSION="$1"
```
Zalgo2462: Add the ability to install a different branch, and remove the (unrecorded) issue where RITA will fail at runtime if built outside of git.

install.sh
```diff
@@ -3,123 +3,404 @@
...
+				;;
+			--disable-bro)
+				_INSTALL_BRO=false
+				;;
+			--disable-mongo)
+				_INSTALL_MONGO=false
+				;;
+			--prefix)
+				shift
+				_INSTALL_PREFIX="$1"
```
ethack: Did we want to just remove `--prefix` as an option?

install.sh
```diff
@@ -3,123 +3,404 @@
...
+	printf "$_IMPORTANT Uninstalling RITA..."
+	if [ -f "$_BIN_PATH/rita"]; then
+		printf "$_SUBITEM Removing $_BIN_PATH/rita \n"
+		$_ELEVATE rm "$_BIN_PATH/rita"
+	else
+		printf "$_SUBIMPORTANT $_BIN_PATH/rita not found! \n"
+	fi
+	if [ -d "$_CONFIG_PATH" ]; then
+		printf "$_SUBITEM Removing $_CONFIG_PATH \n"
+		$_ELEVATE rm -rf "$_CONFIG_PATH"
```
ethack: For uninstalling it makes sense to remove everything, including the config file. But most of our uninstall use case comes when rita is upgraded.  In that case, I'd really like to avoid wiping out the user's config file and making them start a new one.

To address this issue before, I got rid of the idea of "uninstall" and instead called it "reinstall".  This let me just overwrite files and back them up as needed instead of removing them altogether. Thoughts on doing something like this?
https://github.com/ethack/rita/blob/b651f605bf9cda083d1dcb7b8b2068f33f34cc80/install.sh#L326-L331

I really don't think the standalone uninstall is that useful if we leave behind Go, Mongo, and Bro and I don't think we should try to handle uninstalling those.  Also, the uninstall doesn't work between versions of rita where we've changed the file locations and I don't think we should try to keep track of where all the possible locations could be either to have a proper uninstall.  In practice, I haven't found people very receptive to the instruction of "use the uninstaller from the version of rita you have installed and then download the new version to run the install".  They just jump straight to running the new rita installer, which doesn't uninstall old incompatible versions.

ethack:
There's a syntax error somewhere that I haven't been able to track down yet. Posting here in case someone else can find it.

```
user@7019ae3c9eaa:~$ ./install.sh --version dev
[!] The RITA installer requires root privileges for some tasks.
[!] "sudo" will be used when necessary.
[!] RITA is already installed.
[?] Would you like to erase it and re-install? [y/N] y
[!] Uninstalling RITA..../install.sh: line 151: [: missing `]'
	[!] /usr/local/bin/rita not found!
	[!] /etc/rita not found!
	[!] /var/lib/rita not found!
[!] You may wish to uninstall Go, MongoDB, and Bro IDS if they were installed.
```

ethack:
I've run this in a Ubuntu 16.04 docker container and it seems ready to go there.

Zalgo2462:
I'll give it a run on CentOS and Security Onion today.

Zalgo2462:
On CentOS, I'm getting the "RITA is already installed" prompt when I run it for the first time...

Zalgo2462:
Fixed an error on CentOS. Now running fine, provided --disable-bro is added.

ethack:
Yes.  Here's manually running yum after the packages are already installed.
```
[root@2b63c5ff7363 rita]# yum install bro broctl
Loaded plugins: fastestmirror, ovl
Loading mirror speeds from cached hostfile
 * base: mirror.cs.uwp.edu
 * extras: mirrors.cmich.edu
 * updates: reflector.westga.edu
Package bro-2.5.3-1.1.x86_64 already installed and latest version
Package broctl-2.5.3-1.1.x86_64 already installed and latest version
Nothing to do
```

Zalgo2462:
Awesome. Just ran this through on security onion. Runs fine there. One small thing is that security onion is based on 14.04 so it doesn't have systemd/ systemctl so the help text at the end is wrong.

Zalgo2462:
LGTM

PR #198, opened 2018-03-22
Remove tables config

ethack:
I removed the `tables.yaml` config file and hardcoded the values in the code. This was on the roadmap after we split it into its own file and I don't know of any benefit of having those values in a file. Getting rid of the file means we don't have to keep track of it when installing or upgrading.

This will effect the agent since the agent also uses the `tables.yaml` file. We could hardcode it there as well or I have an idea that might work instead.  `rita test-config` outputs yaml and I'm thinking the agent might just be able to use that and forget about other config files altogether.

This pull request should be merged first:
https://github.com/activecm/rita/pull/197


Zalgo2462:
I'd like to have a solid plan for the agent before we move forward with this.

PR #199, opened 2018-03-22
Allow multiple locations for RITA config

ethack:
This change supersedes https://github.com/activecm/rita/pull/185

It searches multiple places for the RITA's config file.  This allows a user to override a system level config file by placing a config in their home or current directory instead.  The following paths are checked in order:

- `./config.yaml` - Checks the same directory as the rita binary. This makes rita very portable and not dependent on a file structure.
- `$HOME/.rita/config.yaml` - Only included because we've used this location before. I'd be ok not searching here.
- `../etc/rita/config.yaml` - This is useful in installs where `bin` and `etc` are on the same level. But based on this https://github.com/activecm/rita/issues/191 issue that would go against the official standard unless we started rolling out packages for different package managers. I'd be ok not searching here too.
- `/etc/rita/config.yaml` - Fallback to the current config location.

I also moved `etc/rita.yaml` to `config.yaml` in the repo so that you can run rita with a valid config file immediately after building the binary.

Note that the same thing could be accomplished by specifying the `-c /path/to/config.yaml` for every single command, but this PR ups the convenience factor.

This pull request should be merged first so that `tables.yaml` is taken out of the equation.
https://github.com/activecm/rita/pull/198


Zalgo2462:
I could go either way with this feature. On one hand, the config yaml is the only hard coded path in RITA. Opening up the default locations for the config file makes RITA much more flexible as to how the installation is structured. On the other hand, I fear that config files existing in multiple locations may make debugging harder, especially when interacting with users who send us bug reports. Thoughts?

ethack:
I agree on both points.  I was going to suggest removing the two middle search paths which would leave only `./config.yaml` and `/etc/rita/config.yaml`.  I'll have to sit on it for awhile to think about how useful loading a config file from the binary's directory is. It saves having to do `-c ./config.yaml` all the time. 

What about this for an alternative: we get rid of the multiple search paths, still use `/etc/rita/config.yaml` by default, but implement an environment variable that can override the config path (e.g. `export RITA_CONFIG=/home/user/rita/config.yaml`)?  It's introducing a new way of interacting with rita but I think it would be less confusing and less prone to making debugging harder than the multiple search paths.

So RITA would use the config from: (in order of preference)
1. `--config` command line flag
2. `RITA_CONFIG` environment variable
3. `/etc/rita/config.yaml` filepath

ethack:
Or... the more I think about it the more I worry I've got a solution without a problem.  The use cases I have are both possibly obsolete.

* I think I was still trying to make a variable install location work and like you said the config yaml is the only hard coded thing that would prevent it.  But I don't think my original vision for the variable install location (i.e. `install.sh --prefix` option`) is worth it anymore either.
* When I was doing multiple hunts on on the same server I kept having to switch config files and I would just symlink to the one I wanted.  It would be nicer in that case to have the environment variable option.  I haven't run into that scenario again in probably the last 6 months, but I think that's only because we haven't been doing multiple hunts on that single server recently.

Maybe the best path is to close this and come back to it if we ever run into a legit use case again where we really don't want the config in /etc/rita/ and have to type `-c` a lot as a result?

PR #200, opened 2018-03-27
Extending docker capabilities and adding statically linked binary build

ethack:
- Update go version for docker image
- Build out to a statically linked binary (linux 64-bit)
- Move documentation to docs folder
- s/CounterMeasures/Countermeasures

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,28 @@
-#RITA runs in Docker!
-#However, it needs a little help.
-#In order to run rita in Docker, two volume mounts are needed. 
-#One for logs, and another for the config file.
-#Alternatively you may extend this dockerfile and add in these files.
-#Make sure your Dockerized RITA config file points to the correct bro log location.
-#Additionally, make sure that RITA has access to a MongoDB server.
-
-#Ex: docker run -it --rm -v /path/to/bro/logs:/logs/:ro -v /path/to/rita/config.yaml:/root/.rita/config.yaml:ro rita import
-#RITA works best with docker-compose. Docker-compose lets you set these mounts
-#and additionally connect it to MongoDB with ease.
-FROM golang:1.8-alpine as rita-builder
-RUN apk update && apk upgrade && apk add --no-cache git make ca-certificates wget
+FROM golang:1.10-alpine as rita-builder
+RUN apk add --no-cache git make ca-certificates wget build-base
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,25 +1,28 @@
...
-#Alternatively you may extend this dockerfile and add in these files.
-#Make sure your Dockerized RITA config file points to the correct bro log location.
-#Additionally, make sure that RITA has access to a MongoDB server.
-
-#Ex: docker run -it --rm -v /path/to/bro/logs:/logs/:ro -v /path/to/rita/config.yaml:/root/.rita/config.yaml:ro rita import
-#RITA works best with docker-compose. Docker-compose lets you set these mounts
-#and additionally connect it to MongoDB with ease.
-FROM golang:1.8-alpine as rita-builder
-RUN apk update && apk upgrade && apk add --no-cache git make ca-certificates wget
+FROM golang:1.10-alpine as rita-builder
```
ethack: Updated to golang 1.10 because it is the latest, RITA appears to work fine with it, and the previous 1.8 tag isn't displayed on Dockerhub anymore.

Dockerfile
```diff
@@ -1,25 +1,28 @@
...
-FROM alpine:latest
+# Change ARGs with --build-arg to target other architectures
+# Produce a self-contained statically linked binary
+ARG CGO_ENABLED=0
+# Set the build target architecture and OS
+ARG GOARCH=amd64
+ARG GOOS=linux
+# Passing arguments in to make result in them being set as 
+# environment variables for the call to go build
+RUN make CGO_ENABLED=$CGO_ENABLED GOARCH=$GOARCH GOOS=$GOOS
```
ethack: This is what allows static compilation. I did several experiments including a bunch of other flags and options I found online, but `CGO_ENABLED=0` was all that ended up being needed in my testing.

Dockerfile
```diff
@@ -1,25 +1,28 @@
...
+
+FROM scratch
```
ethack: With static compilation of rita, we no longer need to download the alpine image to base this off of.  That being said, alpine is still super tiny and if we want the added convenience of being able to run a shell and install packages inside the container we can switch back to using alpine.

Zalgo2462:
LGTM.

Issue #207, opened 2018-05-01
show-scans --ports option doesn't work.


bekirk:
The option only prints out a bunch of control characters. 
see sample below:

`Source,Destination,Ports Scanned,Ports
10.11.20.9,10.24.1.15,72,^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^`

bekirk:
I went into my code and fixed this, here is what I changed in the showScans and showScansHuman, hope this gets fixed in the next release.
 functions in the show-scans.go:
```
                if showPorts {
                        portSet := make([]byte, scan.PortCount*3)
                        for i, port := range scan.PortSet {
                                if i != 0 {
                                        portSet = append(portSet, " "...)  //changed line
                                }
                                portSet = strconv.AppendInt(portSet, int64(port), 10)  //changed line
                        }
                        data = append(data, string(portSet))
                }

```

bekirk:
FYI, the code I added gives me the output I wanted but there are still a bunch of null characters (^@) printed to my output, and help cleaning this up would be great!!!

Thank you!

Zalgo2462:
Interesting. First off, thank you for your bug report! The suggested changes look good. I think taking out the allocation size from the make() call will remove the null characters you're seeing. We will get this merged in as soon as possible.

bekirk:
Thank you I tested with your suggested change and it works,  Sorry this is my first go, at golang :)
It now makes sense because there were always 3 times the null characters as there were scanned ports.

bekirk:
Should I close this case as it seems to be fixed?

Zalgo2462:
We'll close it when the pull request is merged.

Thank you!

PR #209, opened 2018-05-08
Fix the ports display on show scans

Zalgo2462:
Fix issue #207.

### DIFF
--- a/commands/show-scans.go
+++ b/commands/show-scans.go
@@ -22,7 +22,7 @@ func init() {
 			configFlag,
 			cli.BoolFlag{
 				Name:  "ports, P",
-				Usage: "Show which individual ports were scanned",
+				Usage: "Show which individual ports were scanned. Incompaitble with --human-readable.",
 			},
 		},
 		Action: func(c *cli.Context) error {
@@ -31,6 +31,10 @@ func init() {
 				return cli.NewExitError("Specify a database", -1)
 			}
 			showPorts := c.Bool("ports")
+			humanReadable := c.Bool("human-readable")
+			if showPorts && humanReadable {
+				return cli.NewExitError("--ports and --human-readable are incompatible", -1)
+			}
 
 			res := database.InitResources(c.String("config"))
 
@@ -42,8 +46,8 @@ func init() {
 				return cli.NewExitError("No results were found for "+db, -1)
 			}
 
-			if c.Bool("human-readable") {
-				err := showScansHuman(scans, showPorts)
+			if humanReadable {
+				err := showScansHuman(scans)
 				if err != nil {
 					return cli.NewExitError(err.Error(), -1)
 				}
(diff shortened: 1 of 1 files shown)


Zalgo2462:
@ethack We discussed a possible performance issue regarding `rita show-scans -H -P [database]`. However, I looked through the library code we're using for the table writer, and there doesn't appear to be any faster way to create/ write the table. The AppendBulk method provided by the library, simply calls Append in a loop. 

However, I cannot reproduce the timings you had reported. Could you try another dataset and let me know your results?

ethack:
I've confirmed this with a large dataset on the hunt server.  No `-H` returns in ~1s.  With `-H` I got tired of waiting and killed it after 2m15s.  Both times I piped output to /dev/null to avoid waiting for screen updates.  I confirmed the first command was returning results too by piping output to `head`.

```
$ time ./rita-fix show-scans --ports redacted > /dev/null

real	0m1.028s
user	0m1.116s
sys	0m0.044s
$ time ./rita-fix show-scans --ports -H redacted > /dev/null
^C
real	2m15.930s
user	2m14.664s
sys	0m1.680s
```

For comparison, here is the same test with the master branch rita.  Looks like this fix is an improvement in both speed and correctness for the CSV output at least!

```
$ time ./rita-master show-scans --ports redacted > /dev/null

real	0m20.510s
user	0m39.348s
sys	0m4.528s

$ time ./rita-master show-scans --ports -H redacted > /dev/null

real	0m21.160s
user	0m39.860s
sys	0m4.264s
```

bekirk:
This fix was to address an issue I had is there something you want me to test or run?

I run my show-scans in a scheduled crontab so as long as it finishes by 7 am I don’t mind but that could be several hours since I start it at 1 am, last I checked all my tasks were completing around 4:30 am. But that hasn’t changed from pre Rita 1.0 till now. I can start recording time one it and see compare the results. I also don’t output it in human readable. I prefer csv.

ethack:
@bekirk If you'd like to test out the proposed change and make sure it fixes the issue you were having that would be much appreciated.  Alternatively, you could just wait until we merge it into the master branch and let us know if you have any further issues after updating.

@Zalgo2462 I'm fine with the proposed pull request since it fixes the main issue, especially if there is nothing more to be done to improve the human readable processing time.  Actually, I'd be more inclined to want `--ports` and `-H` be mutually exclusive flags. 1) Because the processing time for building the table seems to be unreasonable and 2) because a gigantic list of ports on a single line kind of defeats the purpose of "human readable" anyway 😄 If that's an easy change feel free to make it.  And if not we can just merge it in as-is and count on folks not calling the two simultaneously.

Zalgo2462:
@ethack I think its reasonable to make the flags mutually exclusive. I'll get right on that.

Zalgo2462:
Let me know if you feel if it needs more documentation.

PR #214, opened 2018-05-17
Set Up Integration Testing. Add Beacon Module Tests

Zalgo2462:
The new Makefile supports "make" "make install" "make test" and "make vendor". Make creates a local rita binary, install creates the binary if it does not exist and moves it into GOPATH/bin, test uses docker to create a Mongo database for integration testing purposes and runs unit/ integration tests, and vendor handles running go dep.

The test recipe is the most complex. The test recipe works by first ensuring the vendor/ directory is up to date. Then the recipe creates the MongoDB docker container and extracts the resulting IP address. The recipe then waits for MongoDB to come up using netcat. Finally, bash is used to run the tests. Bash is used here so the MongoDB container is shut down regardless of the tests' success or failure.

The beacon analysis has changed slightly. First, the main Beacon object was split into three smaller, testable components. The collector, analyzer, and writer. The collector and writer have be tested under integration with MongoDB, while the analyzer can be tested on its own.

The analyzer received a slight tweak which removed a nondeterministic aspect of the algorithm. Previously, the mode of a data set was picked at random from the group of tied, most occurring elements. Now, the mode is selected at the smallest element from the group of tied, most occurring elements. Compare analysis/beacon/analyzer.go:createCountMap to analysis/beacon/beacon.go:createCountMap.

database.Resources has been moved to its own package. The resources package contains the usual resource context provided by InitResources, but now also contains a testing resource context provided by InitIntegrationTestingResources. InitIntegrationTestingResources should be called by any test which needs to access MongoDB. InitIntegrationTestingResources assumes that the MongoDB uri to connect to for testing purposes was passed as the last argument to go test (using the -args flag). Using an environment variable was considered, especially since the RITA config file supports environment variables already. However, environment variables cannot be set using many graphical golang test runners (ie. Atom's go-plus), whereas parameters may be easily configured.

Since database.Resources was moved to its own package, database/control.go and database/meta.go needed to be refactored. Instead of providing a cyclic pointer to the Resources context, DB and MetaDB are passed just what they need.

Make test logs the result of each test and properly returns an error code so CI tools can integrate with the test script.


Zalgo2462:
The new Makefile supports "make" "make install" "make test" and "make vendor". Make creates a local rita binary, install creates the binary if it does not exist and moves it into GOPATH/bin, test uses docker to create a Mongo database for integration testing purposes and runs unit/ integration tests, and vendor handles running go dep.

The test recipe is the most complex. The test recipe works by first ensuring the vendor/ directory is up to date. Then the recipe creates the MongoDB docker container and extracts the resulting IP address. The recipe then waits for MongoDB to come up using netcat. Finally, bash is used to run the tests. Bash is used here so the MongoDB container is shut down regardless of the tests' success or failure.

Zalgo2462:
The beacon analysis has changed slightly. First, the main Beacon object was split into three smaller, testable components. The collector, analyzer, and writer. The collector and writer have be tested under integration with MongoDB, while the analyzer can be tested on its own.

The analyzer received a slight tweak which removed a nondeterministic aspect of the algorithm. Previously, the mode of a data set was picked at random from the group of tied, most occurring elements. Now, the mode is selected at the smallest element from the group of tied, most occurring elements. Compare analysis/beacon/analyzer.go:createCountMap to analysis/beacon/beacon.go:createCountMap.

database.Resources has been moved to its own package. The resources package contains the usual resource context provided by InitResources, but now also contains a testing resource context provided by InitIntegrationTestingResources. InitIntegrationTestingResources should be called by any test which needs to access MongoDB. InitIntegrationTestingResources assumes that the MongoDB uri to connect to for testing purposes was passed as the last argument to go test (using the -args flag). Using an environment variable was considered, especially since the RITA config file supports environment variables already. However, environment variables cannot be set using many graphical golang test runners (ie. Atom's go-plus), whereas parameters may be easily configured.

Since database.Resources was moved to its own package, database/control.go and database/meta.go needed to be refactored. Instead of providing a cyclic pointer to the Resources context, DB and MetaDB are passed just what they need.

util/util.go
```diff
@@ -68,6 +68,25 @@ func RemoveSortedDuplicates(sortedIn []int64) []int64 {
 	return result
 }
 
+func CountAndRemoveSortedDuplicates(sortedIn []int64) ([]int64, map[int64]int64) {
```
ethack: I think `RemoveSortedDuplicates` can be redefined to take advantage of this new function and eliminate some duplicate code.

Something like this:
```
func RemoveSortedDuplicates(sortedIn []int64) []int64 {
 	result, counts := CountAndRemoveSortedDuplicates(sortedIn)
 	return result
 }
```

util/util_test.go
```diff
@@ -68,6 +68,27 @@ func TestRemoveSortedDuplicates(t *testing.T) {
 	assert.True(t, sort.IsSorted(SortableInt64(normalTest)))
 }
 
+func TestCountAndRemoveSortedDuplicates(t *testing.T) {
```
ethack: If the above suggestion is implemented we can remove `TestRemoveSortedDuplicates`

analysis/beacon/analyzer.go
```diff
@@ -0,0 +1,226 @@
...
+		a.analysisWg.Done()
+	}()
+}
+
+// createCountMap returns a distinct data array, data count array, the mode,
+// and the number of times the mode occured
+func createCountMap(sortedIn []int64) ([]int64, []int64, int64, int64) {
+	if !sort.IsSorted(util.SortableInt64(sortedIn)) {
+		// If this could be a compile error, it would be
+		panic(errors.New("Unsorted data passed to beacon/analyzer.go:createCountMap"))
```
ethack: I think I understand why this is here, but I'd like to find a different solution. From a user perspective, this would be a pretty frustrating and obtuse error.  It's also the only place `panic` is used and I'd like to avoid using it at all.

If you're really worried about non-sorted data being passed to the `*SortedDuplicates` functions, is there a reason we couldn't just do something like this inside of `util.CountAndRemoveSortedDuplicates` and related functions to ensure data is sorted?  I think it would be a `O(n)` extra cost for each call but the overall function would remain `O(n)` unless it the data isn't already sorted.
```
if !sort.IsSorted(util.SortableInt64(sortedIn)) {
    sort.Sort(util.SortableInt64(sortedIn))
}
``` 
And then rename to something like `SortAndCountAndRemoveDuplicates` to indicate that the functions will sort the data as well?

@joelillo originally brought this up so he might have some other ideas too.

joelillo: Like Ethan said, I asked for his opinion here.

I'm not a huge fan of having panic statements unless they are 100% needed. I understand the motivation behind this one. In theory it should never be called.  However, I'd prefer to see the code structured in such a way that we don't have to worry about it.

 I would maybe just simplify things by renaming RemoveSortedDuplicates to RemoveConsecutiveDuplicates. Put a comment above it that explains that it will remove duplicates in a list if they are consecutive. Make a note that if the list is sorted all duplicates will be removed. That way, people can use the function and they know exactly what it does. It also scopes that function down to one simple task (instead of including sorting). Then, just don't worry about it in the beacons call stack... make comments saying when arrays must be sorted. If people go in and make changes and don't follow the rules, **hopefully** we can catch it with regression testing 😄

util/util.go
Zalgo2462: While the code is cleaner, doing this would result in poor performance in the beaconing analyzer.

`data.ts = util.RemoveSortedDuplicates(data.ts)`
This function is called to dedup any timestamps that come through. These duplications are rare. Since duplicates are rare, the counts map would end up containing almost as many entries as the `data.ts` array. Maps necessarily require more storage than arrays to hold the same elements. Creating and throwing away that much data in a hot section of the code seems rather costly.

CountAndRemoveSortedDuplicates is used to dedup the timestamp delta times and the connection data sizes. Thankfully, the timestamp delta times tend to clump together, creating fewer records in the count map. Unfortunately, the data sizes are not as clustered and consume more ram in the deduplication process. 

Thoughts?

analysis/beacon/analyzer.go
Zalgo2462: I agree the check should be pushed down into the util function.

However, panicking at runtime when you can't statically verify a property of the function parameters is pretty par for the course in Go. See https://github.com/golang/go/blob/master/src/bytes/bytes.go#L460 etc.

Nonetheless, I understand the sentiment, I am certainly open to changing the panic to a sort. 
The sort would change the complexity of the RemoveDuplicates functions from O(n) to O(n lg n). However, since the data is already sorted coming in, it really wouldn't change anything since the sort would never be fired.

joelillo: Pushing the check to the util function sounds good.

I did a little reading around and I see some people do use panics to prevent programming errors, others reserve them for recovering from things like buffer overflows.

Personally... I'd just get rid of the IsSorted check to speed the function up, explain that people sound probably sort their data before passing to the function in the comments, and let people use it at their own risk 😄 

But I'm cool with keeping the error checking too.

ethack: If we left the error check in I also like moving it to the util.go functions.  I just worry about the user experience if anyone ever hits this error, which they shouldn't.  On the other hand if they do hit it, the user experience is probably better if it gives an error of some sort, they report this error, and we can try to track it down rather than just silently not functioning as intended or exhibiting a difficult to track down bug that only becomes apparent when looking at the end data, if at all.  So that's a point in favor of the error.

I don't really like the idea of sorting silently if it isn't already.  It was a suggestion, but probably not a good one.  Looking closer, I see that the returned value is a new array so it feels pretty ugly to both return a new value and modify the input value.

I like Joe's suggestion best of just not sorting and renaming to `RemoveConsecutiveDuplicates` or similar.  That gives it functionality like the linux `uniq` program and the name indicates to anyone using the function when exactly duplicates are removed.

util/util.go
ethack: That's a good point.  Efficiency is definitely a concern, and with the size of the datasets we're dealing with I'd say it's more important than clean code.

Quick back of the napkin calculation:
A `map[int64]int64` will be at least 16 bytes (8 for each int64).  If n is 100,000 then it's 1.6 MB.  Network bandwidth scales on orders of magnitudes (e.g. 100 Mbit, 1 Gb, 10 Gb, etc) so I'm going to assume connection count could do the same. It's not out of the realm of possibility to deal with 2 or even 3 magnitudes larger in the future which would be 160 Mb and then 1.6 GB.  That does seem quite wasteful on memory.

I agree it's safer to err on the side of performance.  Someday if we have the ability to solidly measure performance benchmarks it will be safer to make changes like this knowing the actual impact.

joelillo:
My testing procedure:
- New ubuntu VM
- install go
- clone rita
- install dep
- install docker
- run **make test**

Result of **make test**: docker spins up mongo container, all tests execute and succeed. 

Tried changing expected output in testwriter:
image

Test fails as expected.

**make** command builds binary as expected

**make install** builds binary and moves to correct directory

**make vendor** runs dep ensure -- as expected

Code review looks great. I think we'll be ready to go when the changes discussed today have been wrapped up!

Makefile
```diff
@@ -3,21 +3,28 @@ EXACT_VERSION := $(shell git describe --always --long --dirty --tags)
...
-install: vendor
-	go build ${LDFLAGS} -o ${GOPATH}/bin/${BINARY}
+install: $(BINARY)
+	mv $(BINARY) $(GOPATH)/bin/$(BINARY)
 
 .PHONY: test
 test: vendor
-	go test ${LDFLAGS} ./...
+test: MONGO_ID := $(shell docker run --rm -d mongo:3.6)
+test: MONGO_IP := $(shell docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' $(MONGO_ID))
```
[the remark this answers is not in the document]
Zalgo2462: Good catch!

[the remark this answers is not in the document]
Zalgo2462: Latest commit df89434 should resolve this issue

analysis/beacon/analyzer.go
Zalgo2462: You make a great point bringing up `uniq` as it does everything were doing here to a T.

Zalgo2462:
I went for the renaming plan rather than including a panic anywhere.

PR #222, opened 2018-06-19
Remove DisableIdIndex as the option is deprecated and is an error in 3.7+

Zalgo2462:
`DisableIdIndex` is currently deprecated. We used it to speed up analysis since we don't make use of the index. However, the `_id` index is required from MongoDB 3.7.5 on. While 3.7.x onward is considered unstable, these changes are coming down the line in 3.8. This commit ensures future compatibility.


ethack:
This looks like it only applies to new collections. So this doesn't affect RITA's ability to read or analyze existing datasets right?

PR #229, opened 2018-07-24
Patch to bypass failed package cache updates

ethack:


ethack:
Turns out this just makes the installer fail with a better error message on unsupported OS's.  It's still an improvement, but not a full fix.

https://github.com/activecm/rita/issues/230

PR #231, opened 2018-07-25
Filter unestablished tcp beacons fixes #117

### LINKED ISSUE #117: Failed TCP connections can give false beacon results
Failed TCP connections can cause new connections to happen on the same second. This appears as a beacon with a delta time of 1.

Suggest using the "state" field in the bro logs to exclude any failed connections and only use successful ones.

Note: strictly speaking, malware could hide data in TCP packets that don't go through the 3-way handshake and this would not be detected.

Zalgo2462:
Issue #117 is a bug which creates a large amount of noise in the beaconing analysis. Failed TCP connections such as those generated by a TCP SYN scan appear with regularity in both time and data size. However, these connections never complete the TCP 3 way handshake. This prevents these connections from carrying meaningful data (outside the realm of abusing the protocol). When these connections are filtered from the beacon results, more important findings make their way to the top.

The patch was tested on [maccdc2012_00013.pcap](https://download.netresec.com/pcap/maccdc-2012/maccdc2012_00013.pcap.gz) which contains a large number of port scans. 

Additionally, the beacon tests were refactored and extended (particularly for the collector). 

### Before 
withouttcppatch

### After
withtcppatch

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -94,7 +94,10 @@
 
 [[projects]]
   name = "github.com/stretchr/testify"
-  packages = ["assert"]
+  packages = [
+    "assert",
+    "require"
+  ]
   revision = "12b6f73e6084dad08a7c6e575284b177ecafbc71"
   version = "v1.2.1"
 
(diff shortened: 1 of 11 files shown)


ethack:
The tests pass for me too.

PR #233, opened 2018-07-30
Documentation updates

ethack:
There were some pull requests merged into dev that never made it into master.  I think they are mostly related to documentation.  However, there are some other changes here to code and the install script that I don't believe we want to keep.  Once those changes are figured out and this is merged in we can delete the dev branch.


ethack:
There doesn't seem to be anything here that isn't already in master. The diff shows otherwise, but when I compare them manually the most important changes are already there.

PR #235, opened 2018-08-02
Hard code version number into installer

ethack:
Hard code the version string in the rita installer and print it out when run.
 
1. Having an older installer install a newer version of RITA will eventually break. We can't always maintain the level of backwards/forwards compatibility to keep that up forever. 
2. It's much easier to troubleshoot bug reports if the installer is deterministic in time and prints out its own version number.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,6 +4,7 @@
 # activecountermeasures.com
 
 # CONSTANTS
+_RITA_VERSION="v1.0.2"
 _NAME=$(basename "${0}")
 _FAILED="\e[91mFAILED\e[0m"
 _SUCCESS="\e[92mSUCCESS\e[0m"
@@ -32,11 +33,6 @@ trap '__int' INT
 set -o errexit
 set -o errtrace
 set -o pipefail
-
-if ! [ $(id -u) = 0 ]; then 
-	echo "You do not have permissions to install RITA!"
-	exit 1
-fi
 
 # ENTRYPOINT
 __entry() {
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -33,11 +34,6 @@ set -o errexit
 set -o errtrace
 set -o pipefail
 
-if ! [ $(id -u) = 0 ]; then
```
ethack: Move this after the arguments processing in order to allow running `./install.sh --help` without requiring sudo.

install.sh
```diff
@@ -77,16 +73,17 @@ __entry() {
 			--disable-mongo)
 				_INSTALL_MONGO=false
 				;;
-			-v|--version)
```
ethack: We no longer want to support installing a different version of rita with a given installer. Each installer is tied one-to-one directly to a version of rita.

install.sh
```diff
@@ -373,7 +363,7 @@ __title() {
 "
  _ \ _ _| __ __|  \\
    /   |     |   _ \\
-_|_\ ___|   _| _/  _\\
+_|_\ ___|   _| _/  _\\  $_RITA_VERSION
```
ethack: Version string should be printed out in the `__title` somewhere so that it is visible no matter how the installer is run.

install.sh
```diff
@@ -230,13 +227,6 @@ __install_mongodb() {
 }
 
 __install_rita() {
-	# Set _RITA_VERSION if it's not set
```
ethack: We no longer want to get the latest version from Github as the version is hard coded.

PR #241, opened 2018-08-28
FIX #237 Allow Analysis While Importing Separate Data (With Refactor)

### LINKED ISSUE #237: Allow Analysis While Importing Separate Data

```
DBMetaInfo struct {
}
```

Zalgo2462:
Currently, RITA will attempt to run analysis on a database that is simultaneously being imported. This leads to major inconsistencies. Additionally, streaming importers cannot work RITA in its current state. Streaming importers hold a database open at all times. Any time `rita analyze` is run, the streaming database will be corrupted. Preventing this corruption is the first step towards supporting nightly analysis with IPFIX-RITA and Bro-RITA.

This PR implements the fixes described in issue #237.

This PR cleans up the MetaDB code by splitting it into RITADatabaseIndex and ImportedFilesIndex objects.

This PR, as is, may break older `dep` compatibility. It was authored with dep v0.5.0.


commands/analyze.go
```diff
@@ -39,109 +39,142 @@ func init() {
 
 func analyze(inDb string, configFile string) error {
```
Zalgo2462: The MetaDB class was refactored into the RITADatabaseIndex and RITADatabase classes. So most of the changes revolve around the refactor.

commands/analyze.go
```diff
@@ -39,109 +39,142 @@ func init() {
...
 	res := resources.InitResources(configFile)
-	var toRunDirty []string
-	var toRun []string
+	var toRunDirty []database.RITADatabase
+	var toRun []database.RITADatabase
 
 	// Check to see if we want to run a full database or just one off the command line
 	if inDb == "" {
-		res.Log.Info("Running analysis against all databases")
-		toRunDirty = append(toRun, res.MetaDB.GetUnAnalyzedDatabases()...)
```
Zalgo2462: A possible error is hidden in the old code as res.MetaDB.GetUnAnalyzedDatabases may fail

commands/analyze.go
```diff
@@ -39,109 +39,142 @@ func init() {
...
 
-	// Check for problems
-	for _, possDB := range toRunDirty {
-		info, err := res.MetaDB.GetDBMetaInfo(possDB)
+		//ensure the databases have finished being imported
+		for i := range unanalyzedDBs {
+			if unanalyzedDBs[i].ImportFinished() {
+				toRunDirty = append(toRunDirty, unanalyzedDBs[i])
+			}
+		}
```
Zalgo2462: This is the big change in this PR. We only analyze databases which have finished importing.

commands/analyze.go
```diff
@@ -39,109 +39,142 @@ func init() {
...
+			res.Log.Error(errStr)
+			return cli.NewExitError(errStr, 255)
 		}
-		semVer, err := semver.ParseTolerant(info.ImportVersion)
+		toRunDirty = append(toRunDirty, specifiedDB)
+	}
+
+	// Check for version problems
+	for i := range toRunDirty {
+		compatible, err := toRunDirty[i].CompatibleImportVersion(res.Config.R.Version)
```
Zalgo2462: Checking for database compatibility has been moved into RITADatabase.CompatibleImportVersion

commands/analyze.go
```diff
@@ -39,109 +39,142 @@ func init() {
...
-		logAnalysisFunc("Scanning", td, res,
+		logAnalysisFunc("Scanning", ritaDB.Name(), res,
 			scanning.BuildScanningCollection,
 		)
-		logAnalysisFunc("Cross Reference", td, res,
+		logAnalysisFunc("Cross Reference", ritaDB.Name(), res,
 			crossref.BuildXRefCollection,
 		)
 
-		res.MetaDB.MarkDBAnalyzed(td, true)
```
Zalgo2462: We use to swallow an error in MarkDBAnalyzed which may fail since it interacts with MongoDB

commands/analyze.go
```diff
@@ -39,109 +39,142 @@ func init() {
...
+		err := ritaDB.SetAnalyzed(res.DB.Session, res.Config.R.Version)
+		if err != nil {
+			errStr := fmt.Sprintf("Error: could not mark %s as analyzed: %s", ritaDB.Name(), err.Error())
+			res.Log.Error(errStr)
+			fmt.Println(errStr)
```
Zalgo2462: This line should be deleted as the cli exit error already displays the string

commands/show-databases.go
```diff
@@ -17,8 +17,12 @@ func init() {
 		},
 		Action: func(c *cli.Context) error {
 			res := resources.InitResources(c.String("config"))
-			for _, name := range res.MetaDB.GetDatabases() {
```
Zalgo2462: Swallowed an error

database/ritadb.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/blang/semver"
+)
+
+type (
+	// RITADatabase provides methods for manipulating an
+	// entry in the RITADatabaseIndex
+	RITADatabase struct {
+		indexDoc            DBMetaInfo
+		metaDatabaseName    string
+		indexCollectionName string
```
Zalgo2462: Bit of data duplication here. Chose this over holding a pointer to the RITADatabaseIndex object which created it.

parser/fsimporter.go
```diff
@@ -61,12 +62,15 @@ func (fs *FSImporter) Run(datastore Datastore) {
 		},
 	).Info("Finished collecting file details. Starting upload.")
 
-	indexedFiles = removeOldFilesFromIndex(indexedFiles, fs.res.MetaDB, fs.res.Log)
+	indexedFiles = removeOldFilesFromIndex(indexedFiles, fs.res.FileIndex, fs.res.Log)
```
Zalgo2462: The importer needed some changes after replacing the metadb class with the RITADatabaseIndex and FileIndex classes

parser/fsimporter.go
```diff
@@ -61,12 +62,15 @@ func (fs *FSImporter) Run(datastore Datastore) {
...
 
 	parseFiles(indexedFiles, fs.parseThreads, datastore, fs.res.Log)
 
-	datastore.Flush()
```
Zalgo2462: Swallowed an error

parser/fsimporter.go
```diff
@@ -75,8 +79,13 @@ func (fs *FSImporter) Run(datastore Datastore) {
 			"total_time":   progTime.Sub(start).String(),
 		},
 	).Info("Finished upload. Starting indexing")
+
 	fmt.Println("\t[-] Indexing log entries. This may take a while.")
-	datastore.Index()
```
Zalgo2462: Swallowed an error

parser/mongodatastore.go
```diff
@@ -91,34 +103,49 @@ func (mongo *MongoDatastore) Flush() {
...
+		//Mark the database as fully imported
+		ritaDB, err := mongo.dbIndex.GetDatabase(database)
+		if err != nil {
+			return err
+		}
+
+		err = ritaDB.SetImportFinished(ssn)
+		if err != nil {
+			return err
+		}
```
Zalgo2462: Here's where the importer sets the new `import_finished` field

resources/logging.go
```diff
@@ -36,21 +37,17 @@ func initLogger(logConfig *config.LogStaticCfg) *log.Logger {
 	case 0:
 		logs.Level = log.ErrorLevel
 	}
-	if logConfig.LogToFile {
-		addFileLogger(logs, logConfig.RitaLogPath)
-	}
```
Zalgo2462: Moved outside of initLogger to addFileLogger

database/dbindex.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return RITADatabase{}, err
+	}
+	return r.newRITADatabase(indexDoc), nil
+}
+
+// GetDatabases returns a RITADatabase object for each
+// registered database or an error if the database index
+// collection cannot be read.
+func (r *RITADatabaseIndex) GetDatabases() ([]RITADatabase, error) {
```
SamuelCarroll: I'm very against single letter variable names, I'd like these changed, but it's not needed

reporting/report.go
```diff
@@ -20,16 +21,15 @@ import (
...
 	if len(dbsIn) == 0 {
 		return errors.New("no analyzed databases to report on")
 	}
 
 	var dbs []string
-	for _, db := range dbsIn {
-		info, err := res.MetaDB.GetDBMetaInfo(db)
-		if err == nil && info.Analyzed {
-			dbs = append(dbs, db)
+	for i := range dbsIn {
```
SamuelCarroll: We should maintain the old looping technique if possible. Indexing doesn't look as nice

commands/analyze.go
```diff
@@ -39,109 +39,142 @@ func init() {
...
+			errStr := fmt.Sprintf("Error: failed to list unanalyzed databases: %s", err.Error())
+			res.Log.Error(errStr)
+			return cli.NewExitError(errStr, 255)
+		}
 
-	// Check for problems
-	for _, possDB := range toRunDirty {
-		info, err := res.MetaDB.GetDBMetaInfo(possDB)
+		//ensure the databases have finished being imported
+		for i := range unanalyzedDBs {
```
SamuelCarroll: Again just like having the for _, checkDB := range unanalyzedDBs approach for looping in GoLang

commands/analyze.go
```diff
@@ -39,109 +39,142 @@ func init() {
...
+			errStr := fmt.Sprintf("Error: %s hasn't finished importing.", specifiedDB.Name())
+			res.Log.Error(errStr)
+			return cli.NewExitError(errStr, 255)
 		}
-		semVer, err := semver.ParseTolerant(info.ImportVersion)
+		toRunDirty = append(toRunDirty, specifiedDB)
+	}
+
+	// Check for version problems
+	for i := range toRunDirty {
```
SamuelCarroll: See above comments

commands/analyze.go
```diff
@@ -39,109 +39,142 @@ func init() {
...
+	if len(toRun) == 0 {
+		return cli.NewExitError("", 255)
 	}
 
 	startAll := time.Now()
 
 	fmt.Println("[+] Analyzing:")
-	for _, db := range toRun {
-		fmt.Println("\t[-] " + db)
+	for i := range toRun {
```
SamuelCarroll: Drop index, in favor of the datatype inside

commands/analyze.go
```diff
@@ -39,109 +39,142 @@ func init() {
...
+		fmt.Println("\t[-] " + toRun[i].Name())
+	}
+
+	var dbNames []string
+	for i := range toRun {
```
SamuelCarroll: Same as above

commands/show-databases.go
```diff
@@ -17,8 +17,12 @@ func init() {
...
-				fmt.Println(name)
+			dbs, err := res.DBIndex.GetDatabases()
+			if err != nil {
+				return cli.NewExitError("Error: could not list databases: "+err.Error(), -1)
+			}
+			for i := range dbs {
```
SamuelCarroll: Use a name instead of index

database/dbindex.go
```diff
@@ -0,0 +1,184 @@
...
+	index := RITADatabaseIndex{
+		dbHandle:            dbHandle.Copy(),
+		metaDatabaseName:    metaDatabaseName,
+		indexCollectionName: indexCollectionName,
+		log:                 logger,
+	}
+	return index, nil
+}
+
+func (r *RITADatabaseIndex) openIndexCollection() *mgo.Collection {
```
SamuelCarroll: As a comment (I wouldn't say it's needed now) we should move away from single character variable names. It makes lines longer but improves readability considerably (even for one line functions like this

database/fileindex.go
```diff
@@ -0,0 +1,131 @@
...
+		dbHandle:            dbHandle.Copy(),
+		metaDatabaseName:    metaDatabaseName,
+		indexCollectionName: indexCollectionName,
+		log:                 logger,
+	}
+
+	return index, nil
+}
+
+func (i *ImportedFilesIndex) openIndexCollection() *mgo.Collection {
```
SamuelCarroll: Same single letter variable issue as above, not needed now but we should work away from these

parser/fsimporter.go
```diff
@@ -37,7 +38,7 @@ func NewFSImporter(resources *resources.Resources,
 }
 
 //Run starts importing a given path into a datastore
-func (fs *FSImporter) Run(datastore Datastore) {
+func (fs *FSImporter) Run(datastore Datastore) error {
```
SamuelCarroll: Love the addition of adding an error for the return type

commands/analyze.go
SamuelCarroll: What's the advantage of the refactor?

database/dbindex.go
Zalgo2462: Almost all of RITA uses the first letter of the class name for the member variable name. We could start moving away from that however.

reporting/report.go
Zalgo2462: I could go either way. The other way is more inefficient since it makes a string copy for each loop iteration, but this isn't hot code so it doesn't matter much.

commands/analyze.go
Zalgo2462: The MetaDB was getting way too complicated, becoming a "god" class for keeping track of which databases have imported, analyzed, etc. as well as keeping track of which files were imported through the main importer. This PR adds significant functionality to the database index side of things. Following single responsibility principle, and the scout's rule, I figured I'd make this section of code a better place by splitting out the components.

cbrenton-acm: This seems like a huge number of changes for a single pull request.

SamuelCarroll: I agree. We should have separate PR's for the refactor and the analyze/import collision issue. We should create a separate issue for the refactor and only go with the changes needed for the analysis import issue for this pull request. When creating the issue for refactor make sure that we explain why each step is necessary and the goals of the overall issue. Since the MetaDB is so critical to RITA functioning and RITA is open source we should have good in line documentation for the community to reference while they might be working on RITA for their stuff

ethack: @Zalgo2462 is there a clean way to break these apart? I know we'd miss a lot of the added error checking benefits, but can the needed changes for setting the import finished flag and checking it be added quickly on top of the existing master?  If so, we can merge that part in fairly quickly without waiting any longer to figure out styling and whatnot on the rest.  The refactor looks like it has plenty of merit, but it's just a lot more to go through and a lot more places to cause holdup in reviews if it's not necessary for the required functionality.  

And just to put on record since it was the first question I asked, the refactor didn't change the database schema in Mongo at all.  The only database change was adding a single boolean flag to support the analyze/import collision.

database/dbindex.go
ethack: That makes sense.  I'm not sure it would add much benefit for a tiny function like this but it's probably easiest to stick with one guideline instead of a guideline with a bunch of exceptions.

Since it sounds like single character variables are fairly common elsewhere in RITA, I agree this doesn't need to be the time to address it.  Though we could do with a style guide that covers these types of things (or better yet some custom linting rules).  A refactor would be much easier to do at one point in time after we've formalized something like that.

database/fileindex.go
```diff
@@ -0,0 +1,131 @@
...
+func (i *ImportedFilesIndex) RegisterFiles(files []*fileparsetypes.IndexedFile) error {
+	indexColl := i.openIndexCollection()
+	defer i.closeIndexCollection(indexColl)
+
+	bulk := indexColl.Bulk()
+	bulk.Unordered()
+
+	//construct the interface slice for bulk
+	interfaceSlice := make([]interface{}, len(files))
+	for i, file := range files {
```
ethack: Here's an example where longer (or just different) variable names would help.  The parameter passed into the function is named `i` and here `i` is redefined in the for loop. I'm guessing that the scope is different here and it doesn't bother anything but it would be clearer to not make someone guess how Go handles this case.

I don't necessarily think that `i` in the for loop needs to be any longer since that's pretty standard convention for an index.

database/dbindex.go
Zalgo2462: As a note for future discussion, the [Effective Go](https://golang.org/doc/effective_go.html) guide and the [Go standard library](https://github.com/golang/go/blob/master/src/compress/zlib/reader.go) use single letter variable names for the method receiver. However, I agree that single letter variable names can be confusing. Especially if you use common indexing variables `i`, `j`, or `n`.

commands/analyze.go
Zalgo2462: @ethack I'll get a version in without the refactor / error checks in by the end of the day. 

The error checks could be added to the existing MetaDB class in a separate PR after that if we are interested.

Unfortunately, this refactor can't build on top of the additions to the MetaDB class as the whole MetaDB go file is removed in this PR.

I don't think there is an easy way to break this PR up as it splits the MetaDB class into three classes RITADBIndex, RITADatabase, and ImportedFilesIndex. 

The new RITADBIndex and RITADatabase objects contain new code to handle the import_finished flag. The ImportedFilesIndex is essentially what is left over after extracting RITADBIndex and RITADatabase out of the MetaDB class.

PR #242, opened 2018-08-31
git and make are no longer required for the install

ethack:
When we moved from compiling rita locally to downloading it pre-compiled, we no longer need git or make as dependencies.


Zalgo2462:
LGTM

PR #243, opened 2018-09-09
Install script update 18.04

roberts-dana:
I updated the install.sh file to account for RITA installation on Ubuntu 18.04. Bro must be installed from the source, I made a note at the end of the installation information about installing it from the source. Mongo is already installed in 18.04, for so I had the install script skip that part of the install.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -166,7 +166,7 @@ __install() {
 	printf "$_IMPORTANT '$_START_MONGO'. You can access the MongoDB shell with \n"
 	printf "$_IMPORTANT 'mongo'. If, at any time, you need to stop MongoDB, \n"
 	printf "$_IMPORTANT run '$_STOP_MONGO'. \n"
-	if [ $_UBUNTU_VERSION = "Release:	18.04"]
+	if [ "$_UBUNTU_VERSION" == "Release:	18.04" ]
 	then
 		printf "$_IMPORTANT Bro does not currently have a package for 18.04 and must be installed manually from the source.\n"
 	fi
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -14,7 +14,7 @@ _QUESTION="[?]"
 _SUBITEM="\t$_ITEM"
 _SUBIMPORTANT="\t$_IMPORTANT"
 _SUBQUESTION="\t$_QUESTION"
-
+_UBUNTU_VERSION="$(lsb_release -r)"
```
ethack: Looks like there is already a `$_OS_CODENAME` variable set elsewhere that accomplishes this same thing.  Could you switch the code to use that instead?

install.sh
```diff
@@ -39,8 +39,14 @@ __entry() {
 	_REINSTALL_RITA=false
 
 	# Optional Dependencies
-	_INSTALL_BRO=true
-	_INSTALL_MONGO=true
+	if [ "$_UBUNTU_VERSION" == "Release:	18.04" ]
+	then
+		_INSTALL_BRO=false
+		_INSTALL_MONGO=false
```
ethack: There may be cases where Mongo is not installed.  For instance, I doubt that the server version of Ubuntu will have it pre-installed or the user may have uninstalled it at some point.

So instead of skipping the Mongo install, could you instead
1. Add the checks (if necessary) [here](https://github.com/activecm/rita/blob/2839b50ae482b133ee2347d98f49ff4b79d18788/install.sh#L348-L353) that will properly set the `_MONGO_INSTALLED` variable on 18.04.
2. Add any special casing [here](https://github.com/activecm/rita/blob/2839b50ae482b133ee2347d98f49ff4b79d18788/install.sh#L224-L235) to handle a proper install routine on 18.04.

PR #248, opened 2018-09-17
Update Mongo version to 3.6

ethack:
Testing:

Mongo 3.6 (new install):
- [ ] ~~Ubuntu 14.04 (Security Onion)~~
- [x] Ubuntu 16.04
- [x] CentOS 7

Mongo 3.6 (upgrade):
- [ ] ~~Ubuntu 14.04 (Security Onion)~~
- [x] Ubuntu 16.04
- [x] CentOS 7

Suggestion for testing:
1. Install RITA v1.0.3 or v1.1.0 on a system.
2. Verify that Mongo version 3.4 is installed: `mongod --version`
3. Update RITA using the installer in this pull request.
4. Verify that Mongo version 3.6 is installed: `mongod --version`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -141,6 +141,8 @@ __install() {
 			__load "$_ITEM Installing MongoDB" __install_mongodb
 		elif ! __satisfies_version "$_MONGO_INSTALLED_VERSION" "$_MONGO_VERSION" ; then
 			__load "$_ITEM Updating MongoDB" __install_mongodb
+			# Need to also install all the components of the mongodb-org metapackage for Ubuntu
+			__install_packages mongodb-org-mongos mongodb-org-server mongodb-org-shell mongodb-org-tools
 		else
 			printf "$_ITEM MongoDB is already installed \n"
 		fi
--- a/install.sh
+++ b/install.sh
@@ -212,13 +212,13 @@ __add_bro_to_path() {
 __install_mongodb() {
 	case "$_OS" in
 		Ubuntu)
-			__add_deb_repo "deb [ arch=$(dpkg --print-architecture) ] http://repo.mongodb.org/apt/ubuntu $(lsb_release -cs)/mongodb-org/3.4 multiverse" \
+			__add_deb_repo "deb [ arch=$(dpkg --print-architecture) ] http://repo.mongodb.org/apt/ubuntu $(lsb_release -cs)/mongodb-org/3.6 multiverse" \
 				"MongoDB" \
-				"https://www.mongodb.org/static/pgp/server-3.4.asc"
+				"https://www.mongodb.org/static/pgp/server-3.6.asc"
 			;;
 		CentOS)
-			if [ ! -s /etc/yum.repos.d/mongodb-org-3.4.repo ]; then
-				echo -e '[mongodb-org-3.4]\nname=MongoDB Repository\nbaseurl=https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/3.4/x86_64/\ngpgcheck=1\nenabled=1\ngpgkey=https://www.mongodb.org/static/pgp/server-3.4.asc' | tee /etc/yum.repos.d/mongodb-org-3.4.repo > /dev/null
+			if [ ! -s /etc/yum.repos.d/mongodb-org-3.6.repo ]; then
+				echo -e '[mongodb-org-3.6]\nname=MongoDB Repository\nbaseurl=https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/3.6/x86_64/\ngpgcheck=1\nenabled=1\ngpgkey=https://www.mongodb.org/static/pgp/server-3.6.asc' | tee /etc/yum.repos.d/mongodb-org-3.6.repo > /dev/null
 			fi
 			;;
 	esac
--- a/install.sh
+++ b/install.sh
@@ -266,9 +266,15 @@ __install_mongodb() {
 				"https://www.mongodb.org/static/pgp/server-${_MONGO_VERSION}.asc"
 			;;
 		CentOS)
-			# __add_rpm_repo "https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/${_MONGO_VERSION}/x86_64/"
 			if [ ! -s /etc/yum.repos.d/mongodb-org-${_MONGO_VERSION}.repo ]; then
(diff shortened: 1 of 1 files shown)


Zalgo2462:
`dpkg-query -W -f '${Version}' mongodb-org` gets the version on debian.

`rpm -qa --queryformat='%{VERSION}' mongodb-org` gets the version on centos.

I think those can be used to get us out of setting MONGO_INSTALLED, but even if we add the new repo, I'm not sure apt/ yum will install the new version over the current one with their `install` commands.

install.sh
```diff
@@ -258,14 +261,16 @@ __add_bro_to_path() {
...
-				"MongoDB" \
-				"https://www.mongodb.org/static/pgp/server-3.4.asc"
+			__add_deb_repo "deb [ arch=$(dpkg --print-architecture) ] http://repo.mongodb.org/apt/ubuntu $(lsb_release -cs)/mongodb-org/${_MONGO_VERSION} multiverse" \
+				"mongodb-org-${_MONGO_VERSION}" \
+				"https://www.mongodb.org/static/pgp/server-${_MONGO_VERSION}.asc"
 			;;
 		CentOS)
-			if [ ! -s /etc/yum.repos.d/mongodb-org-3.4.repo ]; then
-				echo -e '[mongodb-org-3.4]\nname=MongoDB Repository\nbaseurl=https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/3.4/x86_64/\ngpgcheck=1\nenabled=1\ngpgkey=https://www.mongodb.org/static/pgp/server-3.4.asc' | tee /etc/yum.repos.d/mongodb-org-3.4.repo > /dev/null
+			# __add_rpm_repo "https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/${_MONGO_VERSION}/x86_64/"
```
ethack: We have this function and I think we might want to use it, but it runs different commands from what's used here so I haven't tested if it works the same.

install.sh
```diff
@@ -258,14 +261,16 @@ __add_bro_to_path() {
...
+			if [ ! -s /etc/yum.repos.d/mongodb-org-${_MONGO_VERSION}.repo ]; then
+				echo -e '[mongodb-org-${_MONGO_VERSION}]\nname=MongoDB Repository\nbaseurl=https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/${_MONGO_VERSION}/x86_64/\ngpgcheck=1\nenabled=1\ngpgkey=https://www.mongodb.org/static/pgp/server-${_MONGO_VERSION}.asc' | tee /etc/yum.repos.d/mongodb-org-${_MONGO_VERSION}.repo > /dev/null
 			fi
+			__freshen_packages
```
ethack: If we used the `__add_rpm_repo` function we wouldn't need this because that function does this. I'm not actually sure why Mongo installed correctly without this before on CentOS. Ubuntu requires an `apt-get update` after a repo is added, but maybe CentOS does not?

install.sh
```diff
@@ -524,6 +530,59 @@ __package_installed() {
...
+	fi
+
+	if [ $_PKG_MGR -eq 1 ]; then # apt
+		echo $(dpkg-query -W -f '${Version}' "$1")
+	elif [ $_PKG_MGR -eq 2 ]; then # yum and dnf
+		echo $(rpm -qa --queryformat='%{VERSION}' "$1")
+	fi
+}
+
+__satisfies_version() {
```
ethack: `__satisfies_version` seems like a prime candidate for unit testing. Does anyone have ideas on how to actually import this to use in a unit test? Running `source install.sh` in another file will cause issues because there are lines that aren't in functions that are run (including the entrypoint). If we split functions like this out into another script file we lose the benefit of having a user download one installer file in order to install RITA.

install.sh
```diff
@@ -524,6 +530,59 @@ __package_installed() {
...
+	# http://wiki.bash-hackers.org/syntax/pe#assign_a_default_value
+	echo ${installed_major:=0} > /dev/null
+	echo ${installed_minor:=0} > /dev/null
+	echo ${installed_patch:=0} > /dev/null
+	echo ${desired_major:=0} > /dev/null
+	echo ${desired_minor:=0} > /dev/null
+	echo ${desired_patch:=0} > /dev/null
+
+	if [ "$installed_major" -lt "$desired_major" ]; then
+		false; return
```
ethack: With bash functions a return value of 1 represents false and 0 represents true since this is convention for program return codes and what `test` expects. `return 1` isn't really intuitive coming from other languages and something like this just seems like a cruel joke:
```
local true=0
local false=1
...
return $true
```
So `false; return` seemed the like the least confusing option. It could also be written on two lines but I figured it was close enough to `return false;` to be ok as-is.

meljbruno:
Should we still be supporting Ubuntu 14.04 even though it's nearing its end of life and Security Onion has a 16.04 iso as well as instructions for upgrading from 14.04 to 16.04?

meljbruno:
Upgrading from Mongo 3.4 to 3.6 with install.sh was unsuccessful on CentOS 7.5, using the current master branch for the base installation.

meljbruno:
Running install.sh from this branch on CentOS 7.5 with no prior RITA installations does not install any version of Mongo.

install.sh
```diff
@@ -266,9 +266,15 @@ __install_mongodb() {
 				"https://www.mongodb.org/static/pgp/server-${_MONGO_VERSION}.asc"
 			;;
 		CentOS)
-			# __add_rpm_repo "https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/${_MONGO_VERSION}/x86_64/"
 			if [ ! -s /etc/yum.repos.d/mongodb-org-${_MONGO_VERSION}.repo ]; then
-				echo -e '[mongodb-org-${_MONGO_VERSION}]\nname=MongoDB Repository\nbaseurl=https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/${_MONGO_VERSION}/x86_64/\ngpgcheck=1\nenabled=1\ngpgkey=https://www.mongodb.org/static/pgp/server-${_MONGO_VERSION}.asc' | tee /etc/yum.repos.d/mongodb-org-${_MONGO_VERSION}.repo > /dev/null
+				cat << EOF > /etc/yum.repos.d/mongodb-org-${_MONGO_VERSION}.repo
```
ethack: Switched writing this file to a multi-line statement to make it easier to read and to allow parameter expansion. Previously, the single quotes prevented expansion.

ethack:
Thanks @meljbruno!  I pushed up a fix that works on CentOS 7.5.  I re-test an upgrade and did fresh install in 16.04 for good measure.

meljbruno:
The fresh install and upgrade both work on CentOS 7.5 now.

meljbruno:
Tested one more time on CentOS 7.5 with the new changes and everything looks good. Merging in.

Issue #238, opened 2018-08-20
Handle undefined version build


ethack:
This is a common enough occurrence that I think there should be a check in rita.  If the version number is "undefined" print out a helpful message to the user and suggest `git clone` and `make` to ensure the version is compiled in.  Or put this in a documentation and have the error message point to the documentation.

![image](https://user-images.githubusercontent.com/1696711/44367513-a29bc500-a495-11e8-91b6-ace79f29f8d1.png)

PR #253, opened 2018-10-01
Adding error message if there's a problem with the RITA version number

SamuelCarroll:
Addresses issue #238

### DIFF
--- a/config/running.go
+++ b/config/running.go
@@ -61,5 +61,11 @@ func loadRunningConfig(config *StaticCfg) (*RunningCfg, error) {
 	outConfig.MongoDB.AuthMechanismParsed = authMechanism
 
 	outConfig.Version, err = semver.ParseTolerant(config.Version)
+	if err != nil {
+		fmt.Println("[!] Version error: please build using the following commands")
+		fmt.Println("\tgit clone https://github.com/activecm/rita.git")
+		fmt.Println("\tcd rita")
+		fmt.Println("\tmake")
+	}
 	return outConfig, err
 }


config/running.go
```diff
@@ -61,5 +61,11 @@ func loadRunningConfig(config *StaticCfg) (*RunningCfg, error) {
 	outConfig.MongoDB.AuthMechanismParsed = authMechanism
 
 	outConfig.Version, err = semver.ParseTolerant(config.Version)
+	if err != nil {
+		fmt.Println("[!] Version error: please build using the following commands")
+		fmt.Println("\tgit clone https://github.com/activecm/rita.git")
+		fmt.Println("\tcd rita")
+		fmt.Println("\tmake")
```
ethack: These commands makes a few assumptions that I think would be tedious to try and cover in the error message:
1. User is in `$GOPATH/src/github.com/activecm/`
2. `rita` directory doesn't exist
3. `make`, `gcc`(?), `go`, `dep` are installed

What do you think about this: instead of providing exact commands, we make the error message give the enough guidance that a developer familiar with RITA will realize what needs to be done and also link to some more detailed instructions on building RITA?  We may want to revamp our developer documentation a bit too but that could come later.

e.g.
> [!] Version error: please ensure that you cloned the git repo and are using make to build
> See the following resources for further information:
> https://github.com/activecm/rita/blob/master/Contributing.md#common-issues
> https://github.com/activecm/rita/blob/master/docs/Manual%20Installation.md

SamuelCarroll: Sounds good, I'll make the changes

PR #258, opened 2018-10-04
FIX #237 Allow Analysis While Importing Separate Data (With Refactor)

### LINKED ISSUE #237: Allow Analysis While Importing Separate Data

Zalgo2462:
Currently, RITA will attempt to run analysis on a database that is simultaneously being imported. This leads to major inconsistencies. Additionally, streaming importers cannot work RITA in its current state. Streaming importers hold a database open at all times. Any time rita analyze is run, the streaming database will be corrupted. Preventing this corruption is the first step towards supporting nightly analysis with IPFIX-RITA and Bro-RITA.

This PR implements the fixes described in issue #237.

This PR, as is, may break older dep compatibility. It was authored with dep v0.5.0.

This PR implements the same features as PR #241, but it does not refactor the MetaDatabase class or implement bug fixes for swallowed errors.

### DIFF
--- a/database/meta.go
+++ b/database/meta.go
@@ -7,9 +7,9 @@ import (
 	"github.com/activecm/rita/config"
 	fpt "github.com/activecm/rita/parser/fileparsetypes"
 	"github.com/blang/semver"
-	log "github.com/sirupsen/logrus"
 	"github.com/globalsign/mgo"
 	"github.com/globalsign/mgo/bson"
+	log "github.com/sirupsen/logrus"
 )
 
 type (
@@ -26,6 +26,7 @@ type (
 	DBMetaInfo struct {
 		ID             bson.ObjectId `bson:"_id,omitempty"`   // Ident
 		Name           string        `bson:"name"`            // Top level name of the database
+		ImportFinished bool          `bson:"import_finished"` // Has this database been entirely imported
 		Analyzed       bool          `bson:"analyzed"`        // Has this database been analyzed
 		ImportVersion  string        `bson:"import_version"`  // Rita version at import
 		AnalyzeVersion string        `bson:"analyze_version"` // Rita version at analyze
@@ -57,9 +58,10 @@ func (m *MetaDB) AddNewDB(name string) error {
 
 	err := ssn.DB(m.config.S.Bro.MetaDB).C(m.config.T.Meta.DatabasesTable).Insert(
 		DBMetaInfo{
-			Name:          name,
-			Analyzed:      false,
-			ImportVersion: m.config.S.Version,
+			Name:           name,
+			ImportFinished: false,
+			Analyzed:       false,
+			ImportVersion:  m.config.S.Version,
 		},
 	)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


Zalgo2462:
Shoot! This code will break if databases don't have the "import_finished" flag at all. We'll need a bit of code to handle the migration.

database/meta.go
```diff
@@ -230,6 +271,46 @@ func (m *MetaDB) GetUnAnalyzedDatabases() []string {
...
+				"err": err,
+			}).Error("Could not add import_finished field to old databases")
+		}
+	}
+	//end migratoin
+
+	iter = ssn.DB(m.config.S.Bro.MetaDB).C(m.config.T.Meta.DatabasesTable).Find(
+		bson.M{
+			"analyzed":        false,
+			"import_finished": true,
```
ethack: Instead of doing the migration code in lines 284-300 above, the approach I took for a similar scenario in the agent was to change the query here to be `"import_finished": bson.M{"$ne": false},` This is equivalent to the migration above as it assumes that all non-existent keys are true. This method requires less code, and my reasoning essentially boiled down to that this is NoSQL and there are no schemas enforced.  Migrations are a necessity for SQL databases because the schema is rigid, but I didn't have a good reason to impose SQL patterns on a NoSQL database. 

I tried to find anything in the Mongo documentation that would indicate which method is idiomatic or preferred for one reason or another. I didn't come up with much, but I thought this was a great discussion on the subject.
https://stackoverflow.com/questions/49330096/schema-migration-scripts-in-nosql-databases

Bottom line for me is that I prefer the "incremental upgrade" because 1) it requires less code, 2) it's less complex (though the resulting query is a little mind bending) and 3) it doesn't require touching every object.  While #3 is not all that important here because the number of objects will be low, I thought it would be a good precedent to set for handling changes in one of the high-volume collections.  I could be swayed by a good argument in the other direction because there are a couple downsides to this approach as well (namely, you have to keep the code there indefinitely and it could become convoluted if there are more instances where this is needed since you can't ever expect the schema to be a certain way).

However, if there's a better case to be made for writing a migration script then I think it may make more sense to break this (lines 284-300) out into its own file/function.  If we add more migrations in the future it would be easier to keep track of them if various snippets weren't embedded across the codebase.

Thoughts?

parser/mongodatastore.go
```diff
@@ -117,6 +117,14 @@ func (mongo *MongoDatastore) Index() {
 			}
 		}
 		collMap.rwLock.Unlock()
+
+		err := mongo.metaDB.MarkDBImported(database, true)
+		if err != nil {
+			mongo.logger.WithFields(log.Fields{
```
ethack: It looks like this `err` will get logged twice, once inside of the `MarkDBImported` function and once here. Is there a specific reason for that?  I think this is just me not understanding the logging mechanism very well so I thought I'd ask in order to learn something.

database/meta.go
Zalgo2462: I like the approach you suggest. However, I do not believe that will work well here. 

The issue is that there is a field on the returned structs `ImportFinished`. If the field doesn't exist in the database, this field will default to false. So, the method here could return structs with the `ImportFinished` field set to false. In order to fix this, we could set `ImportFinished` to true on all of the returned structs, but we'd have to apply this fix to the methods `GetAnalyzedDatabases` and `GetUnanalyzedDatabases` as well. Unfortunately, the fix cannot be directly applied to these other functions since we won't know if the field was unset or missing. 

I agree that doing the migration in the code here is gross. Would it be better if I came up with a migration layer in the install script?

parser/mongodatastore.go
Zalgo2462: Nah, its exactly the way you explain it. It gets double logged. Good catch

Zalgo2462:
@ethack  I implemented the changes we discussed. I feel the code is getting a bit precarious since the other methods in MetaDB avoid the migration function. `GetMetaDBInfo` is the only function which returns `DBMetaInfo` objects outside of meta.go. All the other functions return database names. 

In order to ensure the other methods see migrated versions of the documents, we would need to add a `RunDBQuery(queryDoc bson.M) ([]DBMetaInfo, err)` function. The other methods could call such a method to run their queries, and the function would do the migration for the caller.

The other methods, as it stands, don't interact with ImportFinished, so they work as is. However, going forward things may get hairy.

Zalgo2462:
I've implemented the method discussed in the previous comment in a new PR. The new PR is the same as this PR with an additional commit. https://github.com/activecm/rita/pull/260/files

I feel this extra commit makes things much less precarious inside the meta db class.

Issue #244, opened 2018-09-10
Consider alternate names for reset-analysis & analyze


ethack:
I propose that we either rename, or create aliases if possible to maintain backwards compatibility, for `reset-analysis` and `analyze` commands.

Reasoning:
* "analysis" and "analyze" are both difficult for some to spell. I often find myself having to concentrate extra hard to type these. I also witnessed another user have difficulty with these and had to type it two or three times to get it correct.
* "analysis" is a noun and "analyze" is a verb.  But since they are similar in spelling and have similar meanings in context (both referring to changing the state of the database) it is not easily intuitive as to which to use.
* `reset-analysis` is most often followed by `analyze` in RITA's workflow.  It would be much more convenient to type `rita reset-analysis long_database_name` and then hit the up arrow and simply delete `reset-` from the command.  But instead you have to delete `analysis` as well and again concentrate on switching to the verb `analyze`.

Suggestions in order of preference.  A combination of these can be considered:
* Rename `analyze` to something easier to spell like `process` and `reset-analysis` to something like `reset`.  (Adding aliases if possible would maintain compatibility for people used to typing the current commands.)
* Remove `reset-analysis` command and have `analyze` prompt to reset if necessary.
* Replace `reset-analysis` with `reanalyze`.  This command would combine the two commands to accommodate the common workflow mentioned above.

ethack:
Here are my latest preferences. Keep old names as aliases for backwards compatibility.
* `reset-analysis` -> `reset`. Add flag `-f|--force` to bypass prompt.
* `analyze` remains the same name. Add flag `-r|--reset` to automatically perform `reset` without prompting followed by `analyze`.
* `delete-database` -> `delete`. This is shorter and more consistent with the other commands. Add `-f|--force` flag to bypass prompt.

PR #259, opened 2018-10-08
Activate bash tab autocomplete

zaowen:
Addresses #244
While the issue states that some commands should be renamed all of the reasoning given speaks to general usability issues that would be solved by tab completion.

If nothing else tab completion is nice.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 VERSION := $(shell git describe --abbrev=0 --tags)
 EXACT_VERSION := $(shell git describe --always --long --dirty --tags)
+AUTOPATH := /etc/bash_completion.d/rita
 GOPATH := $(GOPATH)
 BINARY := rita
 
@@ -18,8 +19,12 @@ cache = $(if $(cached-$1),,$(eval cached-$1 := 1)$(eval cache-$1 := $($1)))$(cac
 $(BINARY): vendor $(SRC)
 	go build ${LDFLAGS}
 
+$(AUTOPATH):
+	mkdir -p /etc/bash_completion.d/
+	sudo cp vendor/github.com/urfave/cli/autocomplete/bash_autocomplete $(AUTOPATH)
+
 .PHONY: install
-install: $(BINARY)
+install: $(BINARY) $(AUTOPATH)
 	mv $(BINARY) $(GOPATH)/bin/$(BINARY)
 
 .PHONY: docker-check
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -18,8 +19,12 @@ cache = $(if $(cached-$1),,$(eval cached-$1 := 1)$(eval cache-$1 := $($1)))$(cac
 $(BINARY): vendor $(SRC)
 	go build ${LDFLAGS}
 
+$(AUTOPATH):
+	mkdir -p /etc/bash_completion.d/
+	sudo cp vendor/github.com/urfave/cli/autocomplete/bash_autocomplete $(AUTOPATH)
```
ethack: This target has a dependency on vendor.  If you try to run the "build" it without vendor first it can fail.
```
make /etc/bash_completion.d/rita
mkdir -p /etc/bash_completion.d/
sudo cp vendor/github.com/urfave/cli/autocomplete/bash_autocomplete /etc/bash_completion.d/rita
cp: cannot stat 'vendor/github.com/urfave/cli/autocomplete/bash_autocomplete': No such file or directory
Makefile:23: recipe for target '/etc/bash_completion.d/rita' failed
make: *** [/etc/bash_completion.d/rita] Error 1
```

I'm not sure if there's a benefit having this be its own build target.  How would you feel about just moving these `mkdir` and `cp` instructions under the existing `install` target?

install.sh
```diff
@@ -339,6 +339,8 @@ __install_rita() {
 	touch "$_VAR_PATH/safebrowsing"
 	chmod 755 "$_VAR_PATH"
 	chmod 666 "$_VAR_PATH/safebrowsing"
+
+	curl -sSL "https://raw.githubusercontent.com/urfave/cli/master/autocomplete/bash_autocomplete" -o "/etc/bash_completion.d/rita"
```
ethack: It's probably safest to put a `mkdir -p /etc/bash_completion.d/` before this (like you have in the makefile).

Makefile
zaowen: I wanted it to be its own target because moving the autocomplete file requires root access. I can just add `vendor` as a prerequisite of `$(AUTOPATH)` and that should fix this.

install.sh
zaowen: Good idea

Makefile
ethack: Makes sense, thanks!

PR #260, opened 2018-10-09
Alternative Migration: FIX #237 Allow Analysis While Importing Separate Data

### LINKED ISSUE #237: Allow Analysis While Importing Separate Data

Zalgo2462:
See PR #258. This PR follows up on the comment [here](https://github.com/activecm/rita/pull/258#issuecomment-428366524) by introducing the runDBInfoQuery method to ensure the data migration happens across all MetaDB database methods.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,19 +2,24 @@
 
 
 [[projects]]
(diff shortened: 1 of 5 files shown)


ethack:
Unit tests pass and Docker image builds.

Issue #275, opened 2018-11-28
Disable blacklist url functionality


ethack:
Blacklist url analysis feature is currently deprecated.

Possibly related:
https://github.com/activecm/rita/issues/266

PR #279, opened 2018-11-28
removed blacklisted urls analysis, reporting, and config

lisaSW:
#275


config/static.go
```diff
@@ -61,7 +61,6 @@ type (
 		SafeBrowsing       SafeBrowsingStaticCfg `yaml:"SafeBrowsing"`
```
ethack: SafeBrowsing is only used for blacklist urls.  Should we remove code surrounding that as well?  There are references in the config file and the readme too.

Issue #280, opened 2018-11-28
Disable Scans Analysis 


lisaSW:
Scans analysis feature is currently deprecated.

PR #281, opened 2018-11-28
removed scans module from analysis, reporting, and config

lisaSW:
#280


ethack:
With the `Scanning` section in the config commented out here is the behavior for an older rita version (v1.1.1).

`rita test-config` shows that `ScanThreshold` is set to 0.
```
Scanning:
  ScanThreshold: 0
```
This causes every connection to be reported in the scanning results (anything greater than 0 = everything).  This isn't exactly an incompatible change, but it will definitely be a detriment to performance as it's basically a copy of the uconn collection into scans.

PR #287, opened 2018-12-06
Removing createMetaDB and adding docs to file indexing function

### LINKED ISSUE #286: MetaDatabase files and databases collections out of sync
This showed up in a user's log file.  
```
level=error msg="newMetaDBHandle failed to build database (aborting)" error="a collection 'MetaDatabase.databases' already exists"
```

I traced the source of this error to this line. This error message means that the `MetaDatabase.databases` collection exists but the code attempted to created it.
https://github.com/activecm/rita/blob/v1.1.1/database/meta.go#L469
Which gets called in the `createMetaDB` function. The most likely candidate for that specific error is when it tries to create the `MetaDatabase.databases` collection on these lines.
https://github.com/activecm/rita/blob/v1.1.1/database/meta.go#L500-L501

The sole condition for entering the `createMetaDB` function is when `isBuilt` returns `false`.
https://github.com/activecm/rita/blob/v1.1.1/database/meta.go#L55-L57

There are two cases where `isBuilt` returns false and the user indicated there were no additional error messages, which means that this line was executed. 
https://github.com/activecm/rita/blob/v1.1.1/database/meta.go#L454
What this means is that the `MetaDatabase.files` collection did not exist, causing `createMetaDB` to be executed. However, for some reason `MetaDatabases.databases` *did* exist (hence the error). The code makes the assumption that `databases` cannot exist without `files` also existing but this assumption is apparently false.

There doesn't seem to be a case in the `createMetaDB` path where the `databases` could exist without `files` existing already. Even scenarios where multiple rita instances were running or a rita run crashed midway through wouldn't result in this state, as far as I can tell.

[globalsign/mgo's documentation](https://godoc.org/github.com/globalsign/mgo#Collection.Create) indicates that any use of a collection will result in it automatically be created. There are several other references to `Meta.DatabaseTable` in the code that could feasibly cause the collection to already exist. However, I did not go down the path to figure out if any of these other instances could be executed in such a way to result in this our of sync condition.

Stepping back to look at the purpose of `createMetaDB` we realized that it may not be necessary. It's purpose seems to be manually creating collections in order to add indexes and uniqueness properties to certain fields, including in the `files` and `databases` collections.  Back of the napkin calculations show that indexes may not be necessary.  In the `files` collection are entries for each Bro log file currently imported into the database.  Currently, we use 3 log files (dns, http, and conn) with one per hour per day (24) on the default log rotation.  Again, these entries only exist if the contents of the log files are stored in the database so a reasonable data retention policy of 90 days would yield: 3 * 24 * 90 = 6480 entries in `files`. This is not enough to require an index. One could envision a worst-case scenario where we might benefit from an index, but if the extra complexity can result in bugs like this I don't believe the tradeoff is worth it. 

Furthermore, it appears that many uses of the `files` collection are simply getting all entries from the database and looping through them in the Go code rather than performing queries.  The only place that the index would be used is when deleting a database and the `database` collection is similar.
https://github.com/activecm/rita/blob/v1.1.1/database/meta.go#L132-L140

ethack:
## Description

This change removes the `createMetaDB` function and related code. The rationale is discussed in the issue this closes.

Additionally, comments are added to a function whose name is easily confused with unrelated Mongo terminology.

Closes #286 

## Manual Testing

First, I started with a clean database by dropping MetaDatabase.
```
# in mongo shell
> db.dropDatabase()
{ "dropped" : "MetaDatabase", "ok" : 1 }
```

Next, I imported a sample dataset and verified that it was populated and MetaDatabase was created.
```
$ ./rita import sample-data/bro/dnscat dnscat
[+] Importing sample-data/bro/dnscat
	[-] Finding files to parse
...
	[-] Indexing log entries. This may take a while.
$ rita show-databases
dnscat
```
```
# in mongo shell
> show dbs
MetaDatabase  0.000GB
admin         0.000GB
config        0.000GB
dnscat        0.122GB
local         0.000GB
rita-bl       0.009GB
> use MetaDatabase
switched to db MetaDatabase
> show collections
databases
files
logs
> db.databases.find()
{ "_id" : ObjectId("5c086bc9ca8a551b0f610044"), "name" : "dnscat", "import_finished" : true, "analyzed" : false, "import_version" : "v1.1.1", "analyze_version" : "" }
> db.files.find()
{ "_id" : ObjectId("5c086bcfca8a551b0f6b8f8e"), "filepath" : "sample-data/bro/dnscat/conn.00:00:00-01:00:00.log.gz", "length" : NumberLong(667023), "modified" : ISODate("2018-11-30T16:50:12.216Z"), "hash" : "907d5fd5921b0f4f565e38c45befdb7f", "collection" : "conn", "database" : "dnscat", "time_complete" : ISODate("2018-12-06T00:22:34.801Z") }
...
```
There were no error messages (only info and normal warnings about unrecognized parse files).

Next, I tried re-importing the same files into the same database.
```
./rita import sample-data/bro/dnscat dnscat
[+] Importing sample-data/bro/dnscat
	[-] Finding files to parse
	[-] Indexing log entries. This may take a while.
```
As expected, this did not find any files.  Additionally, there were warnings that stated the following for each file:
```
time="2018-12-05T18:28:36-06:00" level=warning msg="Refusing to import file into the same database twice" path="sample-data/bro/dnscat/conn.00:00:00-01:00:00.log.gz" target_database=dnscat
```

Deleting the database also functioned as expected.
````
./rita delete-database dnscat
Are you sure you want to delete database dnscat [y/N] y
Deleting database: dnscat
````
```
# in mongo shell
> use MetaDatabase
switched to db MetaDatabase
> db.files.find()
> db.databases.find()
```

Here is a comparison that shows the expected result of the MetaDatabase indexes created with a prior version of RITA vs. the version in this pull request.

# v1.1.1
```
> use MetaDatabase
switched to db MetaDatabase
> db.databases.getIndexes()
[
	{
		"v" : 2,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "MetaDatabase.databases"
	},
	{
		"v" : 2,
		"unique" : true,
		"key" : {
			"name" : 1
		},
		"name" : "nameindex",
		"ns" : "MetaDatabase.databases",
		"background" : true
	}
]
> db.files.getIndexes()
[
	{
		"v" : 2,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "MetaDatabase.files"
	},
	{
		"v" : 2,
		"unique" : true,
		"key" : {
			"hash" : 1,
			"database" : 1
		},
		"name" : "hashindex",
		"ns" : "MetaDatabase.files",
		"background" : true
	}
]
```

# this pull request
```
> use MetaDatabase
switched to db MetaDatabase
> db.databases.getIndexes()
[
	{
		"v" : 2,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "MetaDatabase.databases"
	}
]
> db.files.getIndexes()
[
	{
		"v" : 2,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "MetaDatabase.files"
	}
]
```

### DIFF
--- a/database/meta.go
+++ b/database/meta.go
@@ -1,7 +1,6 @@
 package database
 
 import (
-	"os"
 	"sync"
 	"time"
 
@@ -51,10 +50,7 @@ func NewMetaDB(config *config.Config, dbHandle *mgo.Session,
 		dbHandle: dbHandle,
 		log:      log,
 	}
-	//Build Meta collection
-	if !metaDB.isBuilt() {
-		metaDB.createMetaDB()
-	}
+
 	return metaDB
 }
 
@@ -130,11 +126,6 @@ func (m *MetaDB) DeleteDB(name string) error {
 
 	//delete any parsed file records associated
 	_, err = ssn.DB(m.config.S.Bro.MetaDB).C(m.config.T.Meta.FilesTable).RemoveAll(bson.M{"database": name})
-	if err != nil {
-		return err
-	}
-
-	_, err = ssn.DB(m.config.S.Bro.MetaDB).C("files").RemoveAll(bson.M{"database": name})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: database/meta.go [hunk 1]
```diff
@@ -134,11 +130,6 @@ func (m *MetaDB) DeleteDB(name string) error {
 		return err
 	}
 
-	_, err = ssn.DB(m.config.S.Bro.MetaDB).C("files").RemoveAll(bson.M{"database": name})
```
ethack: This chunk was doing the exact same thing as the few lines above it, with the difference that "files" is hard-coded here instead of using the value from the config as above.

parser/fsimporter.go
```diff
@@ -218,6 +218,10 @@ func parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads int, datastore D
 	parsingWG.Wait()
 }
 
+//removeOldFilesFromIndex checks all indexedFiles passed in to ensure
+//that they have not previously been imported into the same database.
+//The files are compared based on their hashes (md5 of first 15000 bytes)
+//and the database they are slated to be imported into.
 func removeOldFilesFromIndex(indexedFiles []*fpt.IndexedFile,
```
ethack: The name of this function is a little ambiguous since it's not clear what makes files "Old" and "Index" is also used in Mongo terminology but that is not what is being referred to here. The comment hopefully helps clear this up.

Issue #308, opened 2018-12-11
Make hardcoded threshold configurable


ethack:
The following value is hardcoded and should be moved to the config file.

https://github.com/activecm/rita/blob/3e4519c2b2b35df555b24817a2f8003528d127d6/parser/fsimporter.go#L278

PR #311, opened 2018-12-13
Move connection limit to mongoDB conf #308

u5surf:
#308 
I fix hardcoding connLimit to be configurable with `ConnectionLimit` in mongoDBCfg
please review it.

### DIFF
--- a/config/static.go
+++ b/config/static.go
@@ -83,7 +83,7 @@ type (
 
 	//StrobeCfg controls the maximum number of connections between any two given hosts
 	StrobeStaticCfg struct {
-		ConnectionLimit int `yaml:"ConnectionLimit"`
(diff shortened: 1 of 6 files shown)


ethack:
@u5surf We really appreciate the contribution!  Would you be willing to make some revisions if I provide some feedback?

u5surf:
@ethack 
i added new config `Strobe`.

ethack:
This looks great, thank you! 

I'm not going to merge this in immediately because there's another piece I'm working on that will allow for default values in RITA. Without that everyone's connection limit would default to 0 so we'd have to write some sort of migration. But with default values it will be much easier to bring this in. So your contribution will get merged in, but I just wanted to explain why it wasn't happening right away.

Issue #322, opened 2018-12-18
resetting the analysis should not erase the freqConn (strobes) collection


lisaSW:
Since this collection is generated at import time, it should not be erased when resetting analysis for a dataset.

PR #323, opened 2018-12-18
excluded freqconn (strobe) collection from reset analysis

lisaSW:
freqconn is created at import, not excluding it from reset-analysis causes loss of that collection until re-import.


lisaSW:
Closes #322

PR #325, opened 2018-12-18
Add command line option --curl-opts

william-stearns:
Allow the user to provide options to all curl requests in the script with a --curl-opts option.


william-stearns:
Ethan and I found that any options one wished to supply to all curl instances can be more easily done by running:
echo "the_command_line_option" >>$HOME/.curlrc
Do not approve or merge this please.

PR #326, opened 2018-12-18
fixed show databases issue

lisaSW:
A recent change attempted to use the same runDBMetaInfoQuery() function for everything in databases/meta.go, including this one where nil was passed in for the query. The issue with that is that the function then proceeded to try to update stuff as needed for the other callers of the function
that would error out and then return nothing in certain cases.

### DIFF
--- a/commands/show-databases.go
+++ b/commands/show-databases.go
@@ -10,8 +10,9 @@ import (
 func init() {
 
 	databases := cli.Command{
-		Name:  "show-databases",
-		Usage: "Print the databases currently stored",
+		Name:    "list",
+		Aliases: []string{"show-databases"},
+		Usage:   "Print the databases currently stored",
 		Flags: []cli.Flag{
 			configFlag,
 		},
--- a/database/meta.go
+++ b/database/meta.go
@@ -277,12 +277,17 @@ func (m *MetaDB) GetDBMetaInfo(name string) (DBMetaInfo, error) {
 
 // GetDatabases returns a list of databases being tracked in metadb or an empty array on failure
 func (m *MetaDB) GetDatabases() []string {
-	dbs, err := m.runDBMetaInfoQuery(nil)
-	if err != nil {
-		return nil
-	}
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	ssn := m.dbHandle.Copy()
+	defer ssn.Close()
+
+	iter := ssn.DB(m.config.S.Bro.MetaDB).C(m.config.T.Meta.DatabasesTable).Find(nil).Iter()
+
 	var results []string
-	for _, db := range dbs {
+	var db DBMetaInfo
+	for iter.Next(&db) {
 		results = append(results, db.Name)
 	}
 	return results


meljbruno:
Tested this out and it fixed the issue of show-beacons not working on my system.

Zalgo2462:
Do you happen to have the error that was being thrown from `runDBMetaInfoQuery`? 

The error was swallowed at the lines in `GetDatabases` that look like
```
if err != nil {
	return nil
}
```


The error is likely due to a malformed or missing(?!) `ImportVersion` field in the metadatabase records. The only other place for an error to arise is the call to `err := ssn.DB(m.config.S.Bro.MetaDB).C(m.config.T.Meta.DatabasesTable).Find(queryDoc).All(&results)` where `queryDoc` is `nil`. Since the equivalent call in the new code doesn’t fail, this line likely succeeded. 

By going around the `runDBMetaInfoQuery` call, we miss the schema migration step.  The migration step isn’t needed at the moment for obtaining the names of all of the databases, so this code works. However, the code may become difficult to manage if we ever need to migrate the name field. 

Overall, I think the code is good, but there might be a deeper issue with historic metadb records.

lisaSW:
@Zalgo2462  thanks so much for that info! Since you are way more familiar with this, would you mind making an issue so we remember to look into the schema migration step for this, as you mention at at the end?

ethack:
@Zalgo2462 it does look like in some cases there are missing import and analyze version strings. Here's the error:
```
strconv.ParseUint: parsing "": invalid syntax
```

I think this happens in the sample datasets we load with mongodump/mongorestore.  The metadatabase entries in those cases are just mocked in.  I can add in some dummy values for the versions to help, but it's probably good to handle the cases in rita too.

Zalgo2462:
We may be able to fix things up right away by changing 
```
 //migrateDBMetaInfo is used to ensure compatibility with previous database schemas.
 //migrateDBMetaInfo does not migrate any data in the database. Rather,
 //it converts old DBMetaInfo representations into new ones in memory.
 //We follow the reasoning that RITA should be able to read documents from
 //older versions.
 func migrateDBMetaInfo(inInfo DBMetaInfo) (DBMetaInfo, error) {
 	inVersion, err := semver.ParseTolerant(inInfo.ImportVersion)
 	if err != nil {
 		return inInfo, err
 	}
 	if inVersion.LT(semver.Version{Major: 1, Minor: 1, Patch: 0}) {
 		/*
 		*	Before version 1.1.0, database records in the MetaDB lacked
 		* the ImportFinished flag. The flag was introduced to prevent
 		* the simultaneous import and analysis of a database.
 		* See: https://github.com/activecm/rita/blob/9fd7ed84a1bad3aba879e890fad83152266c8156/database/meta.go#L26
 		 */
 		inInfo.ImportFinished = true
 	}
 	return inInfo, nil
}
```
to 
```
//migrateDBMetaInfo is used to ensure compatibility with previous database schemas.
 //migrateDBMetaInfo does not migrate any data in the database. Rather,
 //it converts old DBMetaInfo representations into new ones in memory.
 //We follow the reasoning that RITA should be able to read documents from
 //older versions.
 func migrateDBMetaInfo(inInfo DBMetaInfo) (DBMetaInfo, error) {
        var inVersion semver.Version
        var err error
        if inInfo.ImportVersion != "" {
 	        inVersion, err = semver.ParseTolerant(inInfo.ImportVersion)
 	        if err != nil {
 		        return inInfo, err
 	        }
        } else {
                //The only published version of RITA without the ImportVersion field 
                //is RITA v0.9.1
                inVersion = semver.Version{Major: 0, Minor: 9, Patch: 1}
        }
 	if inVersion.LT(semver.Version{Major: 1, Minor: 1, Patch: 0}) {
 		/*
 		*	Before version 1.1.0, database records in the MetaDB lacked
 		* the ImportFinished flag. The flag was introduced to prevent
 		* the simultaneous import and analysis of a database.
 		* See: https://github.com/activecm/rita/blob/9fd7ed84a1bad3aba879e890fad83152266c8156/database/meta.go#L26
 		 */
 		inInfo.ImportFinished = true
 	}
 	return inInfo, nil
}

Then we would keep the old `GetDatabases` code.

@lisaSW would you be willing to give this a try? If not, I'll make up an issue to handle this case and an issue to re-integrate the migrate step into `GetDatabases`

lisaSW:
@Zalgo2462 sure can!

lisaSW:
@meljbruno @ethack pushed up new version with @Zalgo2462 's suggested changes, please see if this version will work too!

PR #328, opened 2018-12-19
Add NeverInclude to filtering

### LINKED ISSUE #162: Add built-in whitelisting capabilities to Rita
Request to add built in whitelisting capabilities to Rita. This includes ip range\domain and all private ip address space for destination.

### LINKED ISSUE #305: Add NeverInclude section to Filtering
As an extension of https://github.com/activecm/rita/pull/301 there should be a `NeverInclude` section which has the opposite effect of `AlwaysInclude` that already exists.  If hosts match the `NeverInclude` list but would normally be imported due to the `InternalSubnets` section then they should not be imported.

If `NeverInclude` and `AlwaysInclude` lists intersect that behavior is undefined and I think we should print an error and refuse to run the import until the config is fixed.

This should be the last piece to also close https://github.com/activecm/rita/issues/162

### LINKED ISSUE #306: Add documentation for the Filtering section
Document the functionality included with this pull request: https://github.com/activecm/rita/pull/301

Put at least a small blurb in the `Readme`.  If extensive documentation is required to explain it all, then also include a separate document in the `docs/` folder.

I'd also like to see a default section placed in the `etc/rita.yaml` config file with comments explaining the settings.  I believe that duplicate sections in the config file will have the latter override the former, which means that any Filtering section appended to the bottom would override the defaults.  This behavior should be tested though.  Alternatively, the Filtering settings themselves could be commented out.  And any scripts that check for this section should detect whether it is commented out or active.

### LINKED ISSUE #307: Add tests for the Filtering functionality
https://github.com/activecm/rita/pull/301 was added but there was no time to implement unit tests.  There can at least be unit tests for the helper functions that read the config and check if an IP matches the config values.  And integration tests could include checking the database for the correct import decisions.

lisaSW:
Closes #305 
Closes #306 
Closes #162 
Closes #307


parser/filter.go
```diff
@@ -0,0 +1,164 @@
...
+		if err != nil {
+			fmt.Println("Error parsing config file CIDR.")
+			fmt.Println(err)
+		}
+	}
+	return internalIPSubnets
+}
+
+// Get "always included" subnets from the config file
+func getIncludedSubnets(res *resources.Resources) (includedSubnets []*net.IPNet) {
```
ethack: Since the code is largely the same for all three `get*Subnets` functions, could we just combine them to something like this? I realize the internal function doesn't do the add '/32' on the end check but I don't see why it couldn't?
```go
// (I'm not sure whether *[]string or []*string is right)
// (and maybe we should add a logger too?)
 func getSubnets(subnetConfig *[]string) (subnets []*net.IPNet) {
  	for _, entry := range subnetConfig {
 		//try to parse out cidr range
 		_, block, err := net.ParseCIDR(entry)
 
  		//if there was an error, check if entry was an IP not a range
 		if err != nil {
 			// try to parse out IP as range of single host
 			_, block, err = net.ParseCIDR(entry + "/32")
 
  			// if error, report and return
 			if err != nil {
 				fmt.Println("Error parsing config file entry.")
 				fmt.Println(err)
 				return
 			}
 		}
 
  		// add cidr range to list
 		subnets = append(subnets, block)
 	}
 	return
 }
```

And call it like:
```go
internalSubnets := getSubnets(res.Config.S.Filtering.InternalSubnets)
neverSubnets := getSubnets(res.Config.S.Filtering.NeverInclude)
alwaysSubnets := getSubnets(res.Config.S.Filtering.AlwaysInclude)
```

parser/filter.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+
+		// add cidr range to list
+		excludedSubnets = append(excludedSubnets, block)
+	}
+	return
+}
+
+// Check if a single IP address is internal
+func (fs *FSImporter) isInternalAddress(ip net.IP) bool {
```
ethack: I do appreciate the nice names having separate functions gives, but I generally steer away from code duplication.  If you think the current code is way better to read than the more generic suggested code below, these functions are short enough I'd still approve it.

```go
func containsIP(subnets []*net.IPNet, ip net.IP) bool {
 	for _, block := range subnets {
 		if block.Contains(ip) {
 			return true
 		}
 	}
 	return false
 }
```
And calling it doesn't look so bad.
```go
isInternalAddress := containsIP(fs.internal, ip)
isSrcIncluded := containsIP(fs.alwaysIncluded, ip)
isSrcExcluded := containsIP(fs.neverIncluded, ip)
```

etc/rita.yaml
```diff
@@ -89,3 +89,23 @@ Beacon:
...
+    # This functions as a whitelisting setting, and connections involving
+    # ranges entered into this section are filtered out at import time
+    #
+    # Example: InternalSubnets: ["10.0.0.0/8","172.16.0.0/12","192.168.0.0/16"]
+    # This allows a user to identify their internal network, which will result
+    # in any internal to internal and external to external connections being
+    # filtered out at import time.
+    AlwaysInclude: []
+    NeverInclude: []
+    InternalSubnets: ["10.0.0.0/8","172.16.0.0/12","192.168.0.0/16"]
```
ethack: I know we have a script that appends the `Filtering` block in later which is why it wasn't included in the first place.  But I just tested out what happens if there are two `Filtering` blocks and the one defined later (e.g. appended to the end) takes precedence. So it should be safe to leave these in the default config (just maybe a little confusing).

Or would it be better to comment all these out by default in order to preserve RITA's existing functionality with no internal networks defined? Since we're doing actual filtering based on these values and we can't know for sure which ones apply to a given user maybe it would be safer to just not do any filtering by default unless the user (or a script) purposely sets these. Thoughts?

parser/filter.go
```diff
@@ -0,0 +1,164 @@
...
+// todo: Error if a valid CIDR is not provided
+func getInternalSubnets(res *resources.Resources) []*net.IPNet {
+	var internalIPSubnets []*net.IPNet
+
+	internalFilters := res.Config.S.Filtering.InternalSubnets
+	for _, cidr := range internalFilters {
+		_, block, err := net.ParseCIDR(cidr)
+		internalIPSubnets = append(internalIPSubnets, block)
+		if err != nil {
+			fmt.Println("Error parsing config file CIDR.")
```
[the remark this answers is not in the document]
ethack: If we get to this point in all the functions we need to exit RITA and not continue.  Here's what happens if we don't:
```
[+] Importing /opt/bro/logs

        [-] Finding files to parse

Error parsing config file CIDR.

invalid CIDR address: 172.16.31.50

Error parsing config file CIDR.

invalid CIDR address: 172.16.31.51

        [-] Parsing /opt/bro/logs/2018-12-19/conn.15:00:00-15:20:02.log.gz -> RITA-2018-12-19

        [-] Parsing /opt/bro/logs/2018-12-19/conn.13:00:00-14:00:00.log.gz -> RITA-2018-12-19

        [-] Parsing /opt/bro/logs/2018-12-19/conn.12:11:38-13:00:00.log.gz -> RITA-2018-12-19

        [-] Parsing /opt/bro/logs/2018-12-19/conn.14:00:00-15:00:00.log.gz -> RITA-2018-12-19

panic: runtime error: invalid memory address or nil pointer dereference

[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x57d6d6]

 

goroutine 25 [running]:

net.networkNumberAndMask(0x0, 0xc42092ba30, 0x10, 0x10, 0xffffffffffffffff, 0x10, 0x10)

        /usr/local/go/src/net/ip.go:445 +0x26

net.(*IPNet).Contains(0x0, 0xc42092ba00, 0x10, 0x10, 0x938cc0)

        /usr/local/go/src/net/ip.go:469 +0x2f

github.com/activecm/rita/parser.isAlwaysIncluded(0xc4216cd340, 0x2, 0x2, 0xc42092ba00, 0x10, 0x10, 0xc420178640)

        /go/src/github.com/activecm/rita/parser/fsimporter.go:209 +0x6d

github.com/activecm/rita/parser.(*FSImporter).parseFiles.func1(0xc4200c6500, 0xc4216cd340, 0x2, 0x2, 0xc420226f80, 0x1, 0x1, 0xc42092b8e8, 0xc420894060, 0xa9d200, ...)

        /go/src/github.com/activecm/rita/parser/fsimporter.go:302 +0x50f

created by github.com/activecm/rita/parser.(*FSImporter).parseFiles

        /go/src/github.com/activecm/rita/parser/fsimporter.go:240 +0x259

panic: runtime error: invalid memory address or nil pointer dereference

[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x57d6d6]

 

goroutine 23 [running]:

net.networkNumberAndMask(0x0, 0xc42073d800, 0x10, 0x10, 0xffffffffffffffff, 0x10, 0x10)

        /usr/local/go/src/net/ip.go:445 +0x26

net.(*IPNet).Contains(0x0, 0xc42073d7d0, 0x10, 0x10, 0x938cc0)

        /usr/local/go/src/net/ip.go:469 +0x2f

github.com/activecm/rita/parser.isAlwaysIncluded(0xc4216cd340, 0x2, 0x2, 0xc42073d7d0, 0x10, 0x10, 0xc42031a040)

        /go/src/github.com/activecm/rita/parser/fsimporter.go:209 +0x6d

github.com/activecm/rita/parser.(*FSImporter).parseFiles.func1(0xc4200c6500, 0xc4216cd340, 0x2, 0x2, 0xc420226f80, 0x1, 0x1, 0xc42092b8e8, 0xc420894060, 0xa9d200, ...)

        /go/src/github.com/activecm/rita/parser/fsimporter.go:302 +0x50f

created by github.com/activecm/rita/parser.(*FSImporter).parseFiles

        /go/src/github.com/activecm/rita/parser/fsimporter.go:240 +0x259
```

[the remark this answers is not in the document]
lisaSW: fixed

parser/filter.go
lisaSW: done!

parser/filter.go
lisaSW: done!

etc/rita.yaml
lisaSW: commented out

parser/filter.go
```diff
@@ -16,19 +15,25 @@ func (fs *FSImporter) filterConnPair(src string, dst string) (ignore bool) {
...
+	isSrcIncluded := containsIP(fs.alwaysIncluded, srcIP)
+	isDstIncluded := containsIP(fs.alwaysIncluded, dstIP)
 
 	// check if on never included list
-	isSrcExcluded := fs.isNeverIncluded(srcIP)
-	isDstExcluded := fs.isNeverIncluded(dstIP)
+	isSrcExcluded := containsIP(fs.neverIncluded, srcIP)
+	isDstExcluded := containsIP(fs.neverIncluded, dstIP)
 
 	// if a result is on both lists we don't ignore it (need error handling later)
```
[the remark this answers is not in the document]
lisaSW: awesome, thank you so much!

PR #310, opened 2018-12-11
update blacklisted analysis

lisaSW:
Closes #304
Closes #274

PR #329, opened 2018-12-19
Implement default config values

ethack:
This pull request gives default values to every member of the config struct. This allows us to gracefully handle older /etc/rita/config.yaml files. If a section or value is missing we can just define a sane default, which is what most people's config files would contain anyway.  And a user can always override defaults with their config file.

This change allows us to add new fields without worrying about migrating a user's previous config file. These pull requests will immediately benefit from this change.
- https://github.com/activecm/rita/pull/311
- https://github.com/activecm/rita/pull/310
- https://github.com/activecm/rita/pull/328

### DIFF
--- a/commands/test-config.go
+++ b/commands/test-config.go
@@ -33,8 +33,8 @@ func testConfiguration(c *cli.Context) error {
 	// First, print out the config as it was parsed
 	conf, err := config.LoadConfig(c.String("config"))
 	if err != nil {
-		fmt.Fprintf(os.Stdout, "Failed to config, exiting")
-		panic(err)
+		fmt.Fprintf(os.Stdout, "Failed to config: %s\n", err.Error())
+		os.Exit(-1)
 	}
 
 	staticConfig, err := yaml.Marshal(conf.S)
--- a/commands/update-check.go
+++ b/commands/update-check.go
@@ -30,17 +30,10 @@ func GetVersionPrinter() func(*cli.Context) {
 //returns a string indicating the new version if available
 func updateCheck(configFile string) string {
 	res := resources.InitResources(configFile)
-	deltaPtr := res.Config.S.UserConfig.UpdateCheckFrequency
+	delta := res.Config.S.UserConfig.UpdateCheckFrequency
 	var newVersion semver.Version
 	var err error
-	var delta int
 	var timestamp time.Time
-
-	if deltaPtr == nil {
-		delta = 14
-	} else {
-		delta = *deltaPtr
-	}
 
 	if delta <= 0 {
 		return ""
@@ -108,7 +101,6 @@ func getRemoteVersion() (semver.Version, error) {
 // The return value is printed regardless so, "" is returned on errror.
 //func informUser( verStr string, index int ) string {
 func informUser(local semver.Version, remote semver.Version) string {
-
 	return fmt.Sprintf(informFmtStr,
 		versions[versionDiffIndex(remote, local)],
 		fmt.Sprint(remote))
--- a/config/running.go
+++ b/config/running.go
@@ -26,19 +26,19 @@ type (
 	}
 )
 
-// loadRunningConfig attempts deserializes data in the static config
-func loadRunningConfig(config *StaticCfg) (*RunningCfg, error) {
-	var outConfig = new(RunningCfg)
+// initRunningConfig uses data in the static config initialize
+// the passed in running config
+func initRunningConfig(static *StaticCfg, running *RunningCfg) error {
 	var err error
 
 	//parse the tls configuration
-	if config.MongoDB.TLS.Enabled {
+	if static.MongoDB.TLS.Enabled {
 		tlsConf := &tls.Config{}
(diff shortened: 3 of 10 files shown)


commands/test-config.go
```diff
@@ -33,8 +33,8 @@ func testConfiguration(c *cli.Context) error {
 	// First, print out the config as it was parsed
 	conf, err := config.LoadConfig(c.String("config"))
 	if err != nil {
-		fmt.Fprintf(os.Stdout, "Failed to config, exiting")
-		panic(err)
+		fmt.Fprintf(os.Stdout, "Failed to config: %s\n", err.Error())
+		os.Exit(-1)
```
ethack: Changed to match what's done in `resources/resources.go:InitResources()`

commands/update-check.go
```diff
@@ -30,18 +30,11 @@ func GetVersionPrinter() func(*cli.Context) {
 //returns a string indicating the new version if available
 func updateCheck(configFile string) string {
 	res := resources.InitResources(configFile)
-	deltaPtr := res.Config.S.UserConfig.UpdateCheckFrequency
+	delta := res.Config.S.UserConfig.UpdateCheckFrequency
```
ethack: This removed code was to allow UpdateCheckFrequency to have a default value of 14.  This is no longer needed and would have been unsustainable to do for every future addition.

config/running.go
```diff
@@ -26,19 +26,19 @@ type (
 	}
 )
 
-// loadRunningConfig attempts deserializes data in the static config
-func loadRunningConfig(config *StaticCfg) (*RunningCfg, error) {
-	var outConfig = new(RunningCfg)
+// initRunningConfig uses data in the static config initialize
+// the passed in running config
+func initRunningConfig(static *StaticCfg, running *RunningCfg) error {
```
ethack: This change wasn't strictly necessary and was just a refactor to match the usage of the `parseStaticConfig` function.  I can revert it if needed.

config/static.go
```diff
@@ -24,65 +24,66 @@ type (
...
-	UserCfgStaticCfg struct {
-		UpdateCheckFrequency *int `yaml:"UpdateCheckFrequency,omitempty"`
-	}
-
 	//MongoDBStaticCfg contains the means for connecting to MongoDB
 	MongoDBStaticCfg struct {
-		ConnectionString string        `yaml:"ConnectionString"`
-		AuthMechanism    string        `yaml:"AuthenticationMechanism"`
-		SocketTimeout    time.Duration `yaml:"SocketTimeout"`
+		ConnectionString string        `yaml:"ConnectionString" default:"mongodb://localhost:27017"`
```
ethack: Every value going forward should have a new `default` tag. For clarity, even if the default value would be the same (e.g. empty string, 0, or empty array) we should define it explicitly so there's no questioning.  The `github.com/creasty/defaults` package takes care of interpreting these tags.

config/static_test.go
```diff
@@ -26,6 +26,8 @@ Bro:
     DBRoot: "RITA"
     MetaDB: MetaDatabase
     ImportBuffer: 100000
+UserConfig:
+    UpdateCheckFrequency: 14
```
ethack: This should have been added before.  But since I changed code related to this feature this is a good sanity check that nothing is broken.

config/static_test.go
```diff
@@ -35,9 +37,8 @@ BlackListed:
 Beacon:
     DefaultConnectionThresh: 24
 Filtering:
-    AlwaysInclude: ["8.8.8.8/32"]
-    InternalSubnets: ["10.0.0.0/8","172.16.0.0/12","192.168.0.0/16"]
-
+    AlwaysInclude: [8.8.8.8/32]
+    InternalSubnets: [10.0.0.0/8,172.16.0.0/12,192.168.0.0/16]
```
ethack: I was just testing to see if the quotes were necessary and forgot to put them back in.  They are apparently not needed but I can add them back if we want.

config/tables.go
```diff
@@ -14,82 +14,47 @@ type (
...
 	}
 
 	//MetaTableCfg contains the meta db collection names
 	MetaTableCfg struct {
-		FilesTable     string
-		DatabasesTable string
+		FilesTable     string `default:"files"`
+		DatabasesTable string `default:"databases"`
 	}
 )
```
ethack: The entire initialization function below is now replaced with the `default` tags and a single function call in the config.go file.

config/testing.go
```diff
@@ -34,24 +38,32 @@ Filtering:
...
 
 	config.S.MongoDB.ConnectionString = mongoURI
 
-	config.T = *loadTableConfig()
+	// Deserialize the yaml file contents into the static config
+	if err := parseStaticConfig([]byte(testConfig), &config.S); err != nil {
+		return nil, err
+	}
+
+	config.S.Version = "v0.0.0+testing"
```
ethack: It was a little confusing before how `Version` and `ExactVersion` variables were used. They were setting variables in another scope which then led to this happening anyway.  I feel this is a little more explicit.

PR #334, opened 2019-01-03
Adding more recommended default networks to Filtering section

ethack:
This PR adds and documents several networks to the default config file.  [RFC 5735, section 4](https://tools.ietf.org/html/rfc5735#section-4) was used as a reference.

I tested the config file was parsed correctly when uncommented.

```
./rita test-config

...
Filtering:
  AlwaysInclude: []
  NeverInclude:
  - 0.0.0.0/32
  - 127.0.0.0/8
  - 169.254.0.0/16
  - 224.0.0.0/4
  - 255.255.255.255/32
  InternalSubnets:
  - 0.0.0.0/8
  - 10.0.0.0/8
  - 172.16.0.0/12
  - 192.168.0.0/16
...
```

### DIFF
--- a/etc/rita.yaml
+++ b/etc/rita.yaml
@@ -60,9 +60,11 @@ UserConfig:
 #Filtering:
     # These are filters that affect the import of connection logs. They
     # currently do not apply to dns or http logs. 
+    # A good reference for networks you may wish to consider is RFC 5735.
+    # https://tools.ietf.org/html/rfc5735#section-4
     #
     # Example: AlwaysInclude: ["8.8.8.8/32"]
-    # This functionlity overrides the NeverInclude and InternalSubnets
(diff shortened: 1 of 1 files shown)


ethack:
@william-stearns Do these networks look good for default suggestions? They are commented out so wouldn't take effect unless the user specifically uncomments one or more of them.

william-stearns:
0.0.0.0/8 must be removed.  At one point it was a "don't allocate from this space" class A, but now it's actively being allocated from because of the shortage of ipv4 addresses.
Please add fe80:0000:0000:0000:0000:0000:0000:0000/112 .  fe80 is the ipv6 equivalent of 169.254.0.0 .  Based on you placing 169.254 in NeverInclude, I'd put the fe80 network there too (though I'm a little hazy on the difference between NeverInclude and InternalSubnets).

PR #339, opened 2019-01-08
Fix #292: Change Logging directory structure

### LINKED ISSUE #292: Change RITA's log directory structure
Currently, RITA creates a new timestamped directory whenever [`addFileLogger`](https://github.com/activecm/rita/blob/v1.1.1/resources/logging.go#L45) is called and the time has changed. This _generally_ results in one directory per rita command run, but can sometimes result in empty directories created.

RITA also splits out logs into their separate log levels (info.log, warn.log, etc) within each of these directories.

In practice it has been fairly difficult to sort through the logs.  This is a proposal to switch the logging format.

1. Instead of separate log files for each log level, they should all log to the same log file (e.g. rita.log) within the log directory. It will be better to see all messages in chronological order and if a user needs to filter out by severity this is easily accomplished with `grep`.
2. Instead of creating a new directory for every new timestamp/rita command, there should be fewer directories. 
  a. I think the most practical suggestion is to create one directory per day.  I prefer this method over (b) below at least to start with.
  b. Another method could be to forget about creating new directories at all and just log everything to the same file. This is what nearly every other linux utility logging to `/var/log` does. This would result in an ever increasing log file and responsibility would be passed onto the user or a tool like `logrotate` to prevent it from growing too large.
3. To make it easier to separate logs for rita commands, an additional log entry could be added that simply logs the command run. This would be an "info" level and include: the file path to the rita binary, the rita version and exact version, and the parameters passed in.

Changes 1 and 2 can happen in [`addFileLogger()`](https://github.com/activecm/rita/blob/v1.1.1/resources/logging.go#L45) and change 3 would ideally happen in [`commands/commands.go`](https://github.com/activecm/rita/blob/v1.1.1/commands/commands.go) if we can figure out a way to hook into all commands generically. Otherwise, it will have to be added at the beginning of each command individually.

zaowen:
The issue did not specify the format of the logged message so I made one up.
Everything else went swimingly

Closes #292

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -66,10 +66,21 @@ func bootstrapCommands(commands ...cli.Command) {
 			//Get access to the logger
 			configFile := c.String("config")
 			res := resources.InitResources(configFile)
-
-			res.Log.WithFields(log.Fields{
+			derp := log.Fields{
 				"Command arguments": c.Args(),
-			}).Info("Running Command: " + command.Name)
+			}
+			//Display flag info in logs
+			for _, it := range c.GlobalFlagNames() {
+				if c.IsSet( it ) {
+					derp["Global Flag(" + it + ")"] = c.GlobalGeneric(it)
+				}
+			}
+			for _, it := range c.FlagNames() {
+				if c.IsSet( it ) {
+					derp["Flag(" + it + ")"] = c.Generic(it)
+				}
+			}
+			res.Log.WithFields(derp).Info("Running Command: " + command.Name)
 			return nil
 		}
 		allCommands = append(allCommands, command)
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -71,13 +71,13 @@ func bootstrapCommands(commands ...cli.Command) {
 			}
 			//Display flag info in logs
 			for _, it := range c.GlobalFlagNames() {
-				if c.IsSet( it ) {
-					derp["Global Flag(" + it + ")"] = c.GlobalGeneric(it)
+				if c.IsSet(it) {
+					derp["Global Flag("+it+")"] = c.GlobalGeneric(it)
 				}
 			}
 			for _, it := range c.FlagNames() {
-				if c.IsSet( it ) {
-					derp["Flag(" + it + ")"] = c.Generic(it)
+				if c.IsSet(it) {
+					derp["Flag("+it+")"] = c.Generic(it)
 				}
 			}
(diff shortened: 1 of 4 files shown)


commands/commands.go
```diff
@@ -60,6 +64,16 @@ var (
 // bootstrapCommands simply adds a given command to the allCommands array
 func bootstrapCommands(commands ...cli.Command) {
 	for _, command := range commands {
+		command.Before = func(c *cli.Context) error {
+			//Get access to the logger
+			configFile := c.String("config")
+			res := resources.InitResources(configFile)
+
+			res.Log.WithFields(log.Fields{
+				"start_time": time.Now().Format(util.TimeFormat),
```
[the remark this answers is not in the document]
zaowen: Good idea, I got it done.

commands/commands.go
```diff
@@ -60,6 +64,16 @@ var (
...
+			}).Info("Running Command: " + command.Name)
```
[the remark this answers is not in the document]
zaowen: Thanks, I think you are correct. Changes have been made.

commands/commands.go
```diff
@@ -60,6 +62,16 @@ var (
 // bootstrapCommands simply adds a given command to the allCommands array
 func bootstrapCommands(commands ...cli.Command) {
 	for _, command := range commands {
+		command.Before = func(c *cli.Context) error {
+			//Get access to the logger
+			configFile := c.String("config")
+			res := resources.InitResources(configFile)
+
+			res.Log.WithFields(log.Fields{
+				"Command arguments": c.Args(),
```
ethack: Could you make it so it shows both flags and arguments? Basically, we want everything the user types on the command line after `rita`.  For instance, it would be very useful for debugging to know that a user used a custom config file and the `--reset` flag in the following command.

```
rita analyze --config etc/rita.yaml --reset vsagent2
```

But the above command produces the following log entry that only shows the dataset and not the extra flags.

```
time="2019-01-11T13:08:35-06:00" level=info msg="Running Command: analyze" Command arguments="[vsagent2]"
```

zaowen: I'll take a look at this later today. Thanks.

zaowen: Newest patch has the requested functionality, turns out it was simpler than I thought it was making it.

1. Just let me know if you want to workshop the format.
2. I added logging for global flags even though the project contains none, in case you add some in the future.

PR #341, opened 2019-01-09
Refuse to run import if InternalSubnets is not configured

### LINKED ISSUE #268: Use config to label conns entries (orig local, resp local)
~~Depends on #267~~

Use the `Filtering: InternalSubnets` config section to update the values for orig local and resp local values from Bro.  ~~Only do this if the InternalSubnets config is not empty.~~ (InternalSubnets will soon be guaranteed to be set so this should happen with the assumption that it is not empty.)

Additionally, add the local* fields to any other collection that uses them: hosts, conn, uconn, etc.

### LINKED ISSUE #337: Don't filter external to external by default
### Current behavior

If `Filtering: InternalSubnets` is not defined in the config file (which is the default) then the [`parser/filter.go:filterConnPair()` function](https://github.com/activecm/rita/blob/master/parser/filter.go#L48-L50) will assume all IPs are external (i.e. if it's not internal then it must be external) and filter out those connections.

### Desired behavior

If a user has not defined the `Filtering: InternalSubnets` config values then we want to avoid filtering anything based on the internal classification.

There should be a check at [this line](https://github.com/activecm/rita/blob/master/parser/filter.go#L36) (before the internal checks) that special-cases if `InternalSubnets` config array is empty. If it is empty then the function should return false.

Additionally, we need test cases in `filter_test.go` to cover these cases and better define the behavior.

ethack:
Fixes #337 by checking if the `InternalSubnets` list is empty. This avoids classifying all IPs as external and filtering them out by default.

I revised the tests to cover more cases, including the ones this PR fixes, and to print a useful message to identify failing test cases.  Here is the relavant output of the tests before this fix was implemented.

```
=== RUN   TestFilterConnPairWithInternalSubnets
--- PASS: TestFilterConnPairWithInternalSubnets (0.00s)
=== RUN   TestFilterConnPairWithoutInternalSubnets
--- FAIL: TestFilterConnPairWithoutInternalSubnets (0.00s)
    filter_test.go:110: 
        	Error Trace:	filter_test.go:110
        	Error:      	Not equal: 
        	            	expected: false
        	            	actual  : true
        	Test:       	TestFilterConnPairWithoutInternalSubnets
        	Messages:   	internal to internal should not be filtered when InternalSubnets is empty
    filter_test.go:110: 
        	Error Trace:	filter_test.go:110
        	Error:      	Not equal: 
        	            	expected: false
        	            	actual  : true
        	Test:       	TestFilterConnPairWithoutInternalSubnets
        	Messages:   	internal to external should not be filtered when InternalSubnets is empty
    filter_test.go:110: 
        	Error Trace:	filter_test.go:110
        	Error:      	Not equal: 
        	            	expected: false
        	            	actual  : true
        	Test:       	TestFilterConnPairWithoutInternalSubnets
        	Messages:   	external to internal should not be filtered when InternalSubnets is empty
    filter_test.go:110: 
        	Error Trace:	filter_test.go:110
        	Error:      	Not equal: 
        	            	expected: false
        	            	actual  : true
        	Test:       	TestFilterConnPairWithoutInternalSubnets
        	Messages:   	external to external should not be filtered when InternalSubnets is empty
FAIL
```

With the fix in place these tests pass.

Also closes #268

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,19 +2,24 @@
 
 
 [[projects]]
+  digest = "1:5a71fb24ce7cbabaa4e30999bbc93e660ddd843af391b8719fb0d2faf0f81901"
   name = "github.com/activecm/mgorus"
   packages = ["."]
+  pruneopts = ""
   revision = "14fb55db664fcb78819379c2fcc01f917d682c67"
   version = "v0.1.1"
 
 [[projects]]
+  digest = "1:df541133b5b4919a9f1a215283caeaaa24cb4b58bc7358c97b970e7baf930391"
   name = "github.com/activecm/mgosec"
   packages = ["."]
+  pruneopts = ""
   revision = "3940290cb0e63eafbbabeb8c4a0a788f80e6bc7d"
   version = "v0.1.1"
 
 [[projects]]
   branch = "master"
+  digest = "1:150cbde83fd4a0c2620001ce4b2ec6bfc102321e7636c260c25ad66c9e4aa933"
   name = "github.com/activecm/rita-bl"
   packages = [
     ".",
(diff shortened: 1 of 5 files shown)


ethack:
We discussed this some more and decided that filtering out traffic based on InternalSubnets is critical to performance.  So instead of not filtering if InternalSubnets is not defined, we should print an error message to the user at import.  To ease the user experience, we can prompt the user to accept the default networks and make the change to the config file, but stress that accurate configuration is important.

If a user really wants to accept the risks of not filtering they can accomplish the same thing by putting "0.0.0.0/8" in the `AlwaysInclude` list but that should be a conscious decision rather than the default behavior as it will likely only work for small datasets.

ethack:
The latest change prints the following error if the InternalSubnets value isn't defined and exits rita.

```
./rita import
Internal subnets are not defined. Please set the InternalSubnets section of the config file.
```

Prompting the user and changing the default config was going to be too complex, but I would like to add some more documentation to the readme about this since it's a requirement. I'm also open to any suggestions on a better way or different wording for the message.

PR #362, opened 2019-02-05
Finished Logmover

Zalgo2462:
This PR build upon the previous logmover work. 
`logpush.bash` has been renamed and reworked in `log_pusher.bash`
`watcher.sh` has been renamed to `log_watcher.bash`

`logpush.bash` has been dramatically simplified using rsync's `--rsync-path` option. This option allows you to specify a script which will run the server side rsync daemon. Normally this script is simply `rsync`. However, I've taken advantage of this option to run the file locking procedure. Previously, we opened up a second ssh session to handle the file locking which created a bit of a mess. The pipe that needed to be moved to /tmp was a part of this mess. (As mentioned at the bottom of the old README)

This PR has been tested by allowing ssh access to my local machine and running multiple instances of `log_pusher.bash` followed by an instance of `log_watcher.bash`. Everything seemed to work fine.

### DIFF
--- a/logmover/Readme.md
+++ b/logmover/Readme.md
@@ -5,58 +5,60 @@ the scripts in this directory will help you automatically transfer the logs
 collected by Bro IDS/ Zeek to your RITA installation and kick off the
 RITA import and analysis process.
 
-The two scripts work with each other. [Watcher.sh](./watcher.sh) runs on the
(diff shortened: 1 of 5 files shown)


ethack:
Closing as we've decided to deprecate the logmover script.

Issue #421, opened 2019-02-26
Remove Unused Configuration Values in BroConfig in v3


Zalgo2462:
In v3, the `ImportDirectory` and `DBRoot/DBName` are required to be passed in via the cli. 

In v2, these values could be loaded from the rita config file under the Bro header. 

These config values should be removed. However, the fields cannot be simply removed. The fields are still used throughout the parser and metadatabase code. Alternative fields must be referenced throughout these sections of code.

PR #437, opened 2019-03-21
Remove DBName, ImportDirectory Config Settings (Option 2)

Zalgo2462:
This PR addresses issue #421 by preventing the config parser from reading in the DBName and ImportDirectory fields. The corresponding fields are kept around in the config struct as they are used in a couple places by the file parser. 

An alternative PR #438 has been prepared which completely removes the config fields from the config struct in addition to removing the values in the config file.


ethack:
Closed in favor of #438

PR #438, opened 2019-03-21
Remove DBName, ImportDirectory Config Settings (Option 1)

Zalgo2462:
This PR addresses issue #421 by completely removing the DBName and ImportDirectory fields from the config structure and config files. 

Removing these fields necessitates changes in the parser to pass the needed data around. 

An alternative PR #437 has been prepared which simply disables reading the config values from the config file and keeps the fields in the structure.


commands/import.go
```diff
@@ -72,7 +72,13 @@ func (i *Importer) parseArgs() error {
 		return cli.NewExitError("\n\t[!] Both <directory to import> and <database name> are required.", -1)
 	}
 
-	err := i.checkForInvalidDBChars(i.targetDatabase)
+	// check if import directory is okay to read from
+	err := i.checkImportDirExists()
+	if err != nil {
+		return err
```
ethack: Should this return a `cli.NewExitError` like the other checks in this function?

Zalgo2462: checkImportDirExists returns a cli exit error. The wrapping call transforming the normal error into a cli exit error could be moved here however.

PR #465, opened 2019-05-22
Removing Bro config section

ethack:
I removed the `ImportBuffer` config item since it was not being used in the code.  This left only `MetaDB` under the `Bro` config section.  Since this didn't really seem to belong in this section anymore I moved `MetaDB` to the `MongoDB` section.

The `BroStaticCfg` struct in `config/static.go` is still being used to store the rolling analysis command line options so I didn't completely remove this. If we have a better place for these values, we could get rid of this struct too.

I built rita, imported a dataset, and confirmed that there were beacon results. Additionally, `make test` passes.  

Anyone using an old config file will revert to the default of "MetaDatabase" for the `MetaDB` field since they won't have this under the `MongoDB` section. This could cause anyone who changed their default metadatabase value to not see their databases until they migrate their config file to the new format. Because of this, I'd like to hear other thoughts on how to handle this. We could: 
1. Not move the MetaDB config value.  It seems silly to leave it in a place that no longer makes sense indefinitely due to backwards compatibility, though I could go along with holding off until we break backwards compatibility for a better reason in the future.
2. Add code to detect that MetaDB is missing and print a warning/error that they should move their old config value.
3. Add code to fallback on checking the old location if the new location doesn't exist.

### DIFF
--- a/config/static.go
+++ b/config/static.go
@@ -20,7 +20,7 @@ type (
 		Beacon       BeaconStaticCfg      `yaml:"Beacon"`
 		DNS          DNSStaticCfg         `yaml:"DNS"`
 		UserAgent    UserAgentStaticCfg   `yaml:"UserAgent"`
-		Bro          BroStaticCfg         `yaml:"Bro"`
+		Bro          BroStaticCfg         `yaml:"Bro"`		// kept in for MetaDB backwards compatibility
 		Filtering    FilteringStaticCfg   `yaml:"Filtering"`
 		Strobe       StrobeStaticCfg      `yaml:"Strobe"`
 		Version      string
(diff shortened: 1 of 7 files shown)


william-stearns:
IMHO, option 3 (falling back on the old location if there's nothing in the new location) is the one with the fewest side effects.

ethack:
Option 3 is now implemented.

old location: Bro: MetaDB
new location: MongoDB: MetaDB

If the old location exists, isn't empty, and the new location is set to the default value then the old location takes precedent.  So this is what this means for some common scenarios:
- User upgrades rita and doesn't touch config file. Old location exists, new doesn't.  Results in using old location.
- User upgrades both rita and config file. Old location doesn't exist, new does.  Uses new location.
- For some odd reason user ends up with both old and new locations defined.  If new location is set to default value it will use old location.  If new location is set to custom value it will use new location.

PR #468, opened 2019-06-03
Update test runners

ethack:
- Migrate from deprecated gometalinter to recommended golangci-lint
- Enabled several new test utilities and fixed minor issues they reported
- Add test.Dockerfile that can run tests in a docker environment
- Update makefile to include more testing options and be used both in and out of docker
- Update TravisCI file to use docker for testing
- Attempted to get Slack integration working with TravisCI, but have been unsuccessful so far
- Added test status and rita logo to readme


util/util_test.go
```diff
@@ -18,27 +18,27 @@ func TestIsIP(t *testing.T) {
...
-	assert.Nil(t, err)
-	file.Close()
-	exists, err := Exists(filePath)
-	assert.Nil(t, err)
-	assert.True(t, exists)
-	os.Remove(filePath)
-	exists, err = Exists(filePath)
-	assert.Nil(t, err)
-	assert.False(t, exists)
+// func TestFileExists(t *testing.T) {
```
ethack: Commented this test out because it relies on the filesystem and doesn't make a great unit test like the rest of the tests in this file.

pkg/useragent/mongodb_test.go
```diff
@@ -1,3 +1,5 @@
+// +build integration
```
ethack: These tags allow separating out unit tests and integration tests.  `go test` will run unit tests by default and can run integration tests instead by using `go test --tags=integration ...`

PR #470, opened 2019-06-06
Flags for database selection

### LINKED ISSUE #332: Expand delete database command
The current delete-database command should be expanded with the following:

- `-a, --all` flag which indicates that all databases should be removed
- `-m, --match` flag which indicates that the given database string should be used as a substring to match against all databases.  (e.g. "2018-12" would match all databases from December like RITA-2018-12-01, RITA-2018-12-02, etc)
- `-r, --regex` flag which indicates that the given database string should be used as a regex to match against all databases.
- `-n, --dry-run` flag which will only print the databases which would be deleted, but not do any deletions and not give any prompts.

The current behavior of the command remains intact:
- Providing a single database name with no arguments will do an exact match on only one database
- Not specifying the `-f, --force` flag will prompt for confirmation on every database, similar to the linux `rm -i` behavior, and including the `--force` flag will delete without prompting, similar to `rm -f`. (The force flag is pending [this pull request](https://github.com/activecm/rita/pull/324))

This feature request is open for suggestions and discussion.

~~These flags could also be applied to the `reset` and `analyze` commands as well.  The behavior is a little bit different.~~
- ~~`reset` has `--all` the default behavior when no database is specified.~~
- ~~`analyze`  has `--all` the default behavior when no database is specified and always assumes `--force` (never prompts)~~

~~So `--force` and `--all` could be added simply for consistency and so that someone leaving those flags in still does the intended behavior. `rita reset --all` and `rita reset` are equivalent, as are `rita analyze --force` and `rita analyze --all`, etc.~~

Edit: the `reset` and `analyze` commands do not exist in the current version anymore.

bglebrun:
Adds the `-f`, `-a`, `-m`, `-r` and `-n` flags for forced (unprompted), all, match, regex, and dry run respectively to the delete command. Fixes #332.

Concerning the changes outside of `commands/delete-database.go` and `commands/commands.go`, `gofmt` flagged certain parts of these files as incorrect and the changes made to them are the result of running `gofmt -w .` and can be reverted if needed


commands/commands.go
```diff
@@ -32,6 +32,32 @@ var (
...
+	regexFlag = cli.StringFlag{
+		Name:  "regex, r",
+		Usage: "Indicates use `REGEX_PATTERN` regular expression to define databases to be removed",
+		Value: "",
+	}
+
+	// dryRun indicates which databases would be deleted with current options
+	dryRunFlag = cli.BoolFlag{
+		Name:  "dry-run, n",
+		Usage: "Test which databases would be deleted, but does not actually delete, nor ask for prompts",
```
Zalgo2462: Can we reword this?
Maybe "Tests which databases would be deleted. Does not actually delete any data, nor prompt for confirmation".

commands/delete-database.go
```diff
@@ -15,81 +17,176 @@ func init() {
...
+
+	// Different command flags
+	match := c.String("match")
+	regex := c.String("regex")
+	bulk := c.Bool("all")
+	force := c.Bool("force")
+	dryRun := c.Bool("dry-run")
+	var names []string
+	var err error
+
```
Zalgo2462: Do you think we should make the flags exclusive? Should we throw an error if an invalid combination is supplied?
What happens if a user runs `rita delete-database -r -m (foo|bar)`? 
In this implementation, the string `(foo|bar)` would be directly matched against the db list.

@ethack is this something worth dealing with?

commands/delete-database.go
```diff
@@ -15,81 +17,176 @@ func init() {
...
+	} else {
+		// get all database names
+		names = res.MetaDB.GetDatabases()
+	}
 
-	// if no force flag, verify action with the user
-	if !forceFlag {
-		fmt.Print("Are you sure you want to delete database ", database, " [y/N] ")
+	// check if we have databases
+	if err != nil || len(names) == 0 {
```
Zalgo2462: In what case would err not be nil here? Maybe we can remove the check?

commands/delete-database.go
```diff
@@ -15,81 +17,176 @@ func init() {
...
+		for _, db := range dbs {
+			matched := regq.MatchString(db)
+			if matched {
+				names = append(names, db)
+			}
+		}
+
+	} else {
+		// get all database names
+		names = res.MetaDB.GetDatabases()
```
Zalgo2462: Could we reflow the code so that if we're deleting a single db (no flags), we skip getting a list of the database names?

commands/delete-database.go
```diff
@@ -15,81 +17,176 @@ func init() {
...
+	}
 
-		read := bufio.NewReader(os.Stdin)
+	// For single database deletion
+	if !bulk {
 
-		response, _ := read.ReadString('\n')
+		database := c.Args().Get(0)
+		if database == "" {
+			return cli.NewExitError("Specify a database or specify --all or -a for all databases", -1)
```
Zalgo2462: Let's remove mention of the -a flag here since there's others we're not mentioning.

commands/delete-database.go
Zalgo2462: It might simplify things to treat the single (no flags) case as just another bulk case. You'd end up just processing a list with one item.

I think using the print statements for the bulk case would work for the  single case as well.

commands/commands.go
```diff
@@ -32,6 +32,32 @@ var (
...
 	}
 
+	// allFlag indicates all databases in option
+	allFlag = cli.BoolFlag{
+		Name:  "all, a",
+		Usage: "Indicates all databases should be removed",
+	}
+
+	// matchFlag indicates only matched databases to string SEARCH_STRING
+	matchFlag = cli.StringFlag{
```
Zalgo2462: You might think about implementing these as BoolFlags so you only have to deal with reading in the search string once. No need to change it if you don't want to though.

commands/delete-database.go
bglebrun: I can try to run some optimizations, the code flow plan was that it would reach that else statement in case a match or regex flag wasn't set. The code as it was written before gets a list of databases at some point to check if the database exists and this was the adoption of that. I agree though the section between 98 and 138 could be optimized better for this somehow

commands/delete-database.go
```diff
@@ -15,81 +17,177 @@ func init() {
...
+				names = append(names, db)
+			}
+		}
+
+	} else {
+		// Single database deletion
+		tgtDB := c.Args().Get(0)
+		// get all database names
+		dbs := res.MetaDB.GetDatabases()
+		if find(dbs, tgtDB) != len(dbs) {
```
ethack: Can you use [`util.StringInSlice()`](https://github.com/activecm/rita/blob/master/util/util.go#L80) instead of find()?

commands/delete-database.go
```diff
@@ -15,81 +17,177 @@ func init() {
...
+			return cli.NewExitError(dberr.Error, -1)
 		}
 	}
 
-	// check if we have databases
-	if err != nil || len(names) == 0 {
-		return cli.NewExitError("Failed to find any databases", -1)
+	// Dry run warning
+	if dryRun {
+		fmt.Fprintf(os.Stdout, "\t[-] This was a dry run of the delete command, nothing has been changed!\n")
```
ethack: Can you switch all these `fmt.Fprintf(os.Stdout` to just be `fmt.Printf(`?  I think they were already in the file that way, but I don't know of a good reason to keep them and it's inconsistent with the Printf and Println we've been using lately.

commands/delete-database.go
```diff
@@ -15,81 +17,177 @@ func init() {
...
+
+	response = strings.ToLower(strings.TrimSpace(response))
+	if response == "y" || response == "yes" {
+		return true
+	}
+
+	return false
+}
+
+func checkFlags(a bool, b bool, c bool) bool {
```
ethack: It's not clear what this function does.  Can you please add a header that describes its purpose along with which cases it returns true/false?

ethack:
There are a couple things that aren't quite working correctly.

```
$ ./rita delete
Please select a single bulk option
```
If it's run with no arguments, could we print out a different message? Maybe "Please provide a database to delete or use one of the bulk delete options"?  I'm open to "Failed to find any databases" if it makes the logic easier in the code.

```
$ ./rita list
vsagent-strobe
empire-test-files
empire-rolling-files
empire-rolling-test
empire-files
empire-files2
empire-files3
$ ./rita delete --all -n
Failed to find any databases
$ ./rita delete --all   
Failed to find any databases
$ ./rita delete -m empire -n 
	[-] Successfully deleted database empire-test-files.
	[-] Successfully deleted database empire-rolling-files.
	[-] Successfully deleted database empire-rolling-test.
	[-] Successfully deleted database empire-files.
	[-] Successfully deleted database empire-files2.
	[-] Successfully deleted database empire-files3.
	[-] This was a dry run of the delete command, nothing has been changed!
```

There are databases but the --all flag doesn't work even though the match and regex options do.

```
$ ./rita delete empire-files3
Please select a single bulk option
```

Deleting a single database no longer works.

ethack:
--match looks good now but the no argument case and the --all cases are still not quite right.

```
$ ./rita delete -n --all

$ ./rita delete

$ ./rita delete -n

$ ./rita list
vsagent-strobe
corelight-rolling

$ ./rita delete --all

$ ./rita list
vsagent-strobe
corelight-rolling

$ ./rita delete corelight-rolling

$ ./rita list
vsagent-strobe
corelight-rolling
```

- `delete` with no params prints nothing.
- `delete` with param prints nothing and does nothing.
- `--all` prints nothing and does nothing.

ethack:
Manual testing all looks good.  I'd feel more comfortable if we had automated tests too, but that will require a bit of rework and I don't want to hold this up anymore.

PR #471, opened 2019-06-06
Configurable limits on show-* commands

### LINKED ISSUE #466: Add --limit flag to show-* commands
The show-* commands in rita limit the output results to 1000.  This value is hard-coded in and should be made a command line flag.  A value of 0 should indicate no limit (alternatively, we could add a --no-limit flag).  If the flag is not specified it should default to 1000.

```
commands/show-explodedDns.go
33:			data, err := getExplodedDNSResultsView(res, 1000)

commands/show-user-agents.go
43:			data, err := getUseragentResultsView(res, sort, sortDirection, 1000)

commands/show-strobes.go
43:			data, err := getStrobeResultsView(res, sortStr, sortDirection, 1000)

commands/show-bl-hostname.go
43:	data, err := getBlacklistedHostnameResultsView(res, "conn_count", 1000)

commands/show-long-connections.go
38:			data, err := getLongConnsResultsView(res, thresh, sortStr, sortDirection, 1000)

commands/show-bl-ip.go
76:	limit := 1000
117:	limit := 1000
```

Note that the show-beacons command does not have a limit.  However, beacons have a threshold set in the config file so that only beacons above a certain score are displayed.  We should consider either adding the --limit option to the show-beacons command as well for consistency.

Related:
https://github.com/activecm/rita/issues/463

bglebrun:
Implemented `--limit` flag that defaults to 1000 to preserve current behavior and still allow for customization. 
Implemented `--no-limit` flag that will print all results

Closes #466


commands/show-bl-hostname.go
```diff
@@ -155,6 +156,10 @@ func getBlacklistedHostnameResultsView(res *resources.Resources, sort string, li
 		}},
 	}
 
+	if !noLimit {
```
ethack: The limit should still go immediately after the `$sort` step for efficiency reasons.
> When a $sort precedes a $limit and there are no intervening stages that modify the number of documents, the optimizer can coalesce the $limit into the $sort. This allows the $sort operation to only maintain the top n results as it progresses, where n is the specified limit, and ensures that MongoDB only needs to store n items in memory. This optimization still applies when allowDiskUse is true and the n items exceed the aggregation memory limit.
https://docs.mongodb.com/manual/reference/operator/aggregation/limit/

You'll probably have to break up the query and append the `$project` stage afterwards.

commands/show-bl-ip.go
```diff
@@ -234,6 +235,10 @@ func getBlacklistedIPsResultsView(res *resources.Resources, sort string, limit i
 		}},
 	}
 
+	if !noLimit {
```
ethack: Same as above.

commands/show-explodedDns.go
```diff
@@ -105,7 +107,10 @@ func getExplodedDNSResultsView(res *resources.Resources, limit int) ([]explodedd
 		}},
 		bson.M{"$sort": bson.M{"visited": -1}},
 		bson.M{"$sort": bson.M{"subdomain_count": -1}},
-		bson.M{"$limit": limit},
+	}
+
+	if !noLimit {
```
ethack: This one is fine since it's already at the end.

commands/show-user-agents.go
```diff
@@ -109,7 +111,10 @@ func getUseragentResultsView(res *resources.Resources, sort string, sortDirectio
 			"seen":       1,
 		}},
 		bson.M{"$sort": bson.M{sort: sortDirection}},
-		bson.M{"$limit": limit},
+	}
+
+	if !noLimit {
```
ethack: This one is fine since it's already at the end.

ethack:
@bglebrun See my comments above for requested changes.

bglebrun:
@ethack made the requested changes

commands/commands.go
```diff
@@ -62,6 +62,17 @@ var (
 		Usage: "Print a report instead of csv",
 	}
 
+	limitFlag = cli.IntFlag{
+		Name:  "limit, li",
+		Usage: "Limit the outputs of the result to `LIMIT` values",
+		Value: 1000,
+	}
+
+	noLimitFlag = cli.BoolFlag{
```
SamuelCarroll: So random thought check with Ethan, but mongo implements a limit flag with an option of 0. If the limit is equal to zero that indicates no limits. Would it be easier to use that or keep as is. Also can we default the value to False?

commands/show-bl-hostname.go
```diff
@@ -40,7 +42,7 @@ func printBLHostnames(c *cli.Context) error {
 	res := resources.InitResources(c.String("config"))
 	res.DB.SelectDB(db)
 
-	data, err := getBlacklistedHostnameResultsView(res, "conn_count", 1000)
+	data, err := getBlacklistedHostnameResultsView(res, "conn_count", c.Int("limit"), c.Bool("no-limit"))
```
SamuelCarroll: If we have passed the config values into the res value do we need to parse the limit and no limit flags here? Since we should have those initialized could we use the res file below?

commands/show-bl-ip.go
```diff
@@ -71,8 +73,8 @@ func printBLSourceIPs(c *cli.Context) error {
 			bson.M{"blacklisted": true},
 			bson.M{"dat.count_src": bson.M{"$gt": 0}},
 		}}
-	limit := 1000
-	data, err := getBlacklistedIPsResultsView(res, sort, limit, match, "src", "dst")
+	limit := c.Int("limit")
+	data, err := getBlacklistedIPsResultsView(res, sort, c.Bool("no-limit"), limit, match, "src", "dst")
```
SamuelCarroll: It seems a bit odd to me to the line above use limit := c.Int("limit) then on this line call that again

commands/commands.go
bglebrun: Ethan just mentioned [in the issue](https://github.com/activecm/rita/issues/466) that 1000 as a default was a requirement for the pull, it shouldn't be any issue to change it though

SamuelCarroll: I wasn't super clear but 1000 is good.
If a user tries something like `--li 0` they get an error saying the limit must be positive. If we changed the PR so the --nl flag doesn't exist and we make `--li 0` be the no limit flag if that makes sense

commands/show-bl-hostname.go
SamuelCarroll: Sorry about that, I was mistaken on how the command line arguments were handled. Ben pointed it out so this is all good!

commands/commands.go
bglebrun: It's either way, I'll defer to @ethack 's wishes on it, the no limit flag was just another parameter request in the issue

SamuelCarroll: Sounds good! We'll leave it in then

ethack: I'm good either way. The `--no-limit` flag was given as an option equivalent to `--limit 0`. I left it to whoever implemented it to weigh the pros and cons. Since Mongo doesn't allow passing a straight 0 into the limit field in aggregations we'd still have to special case the 0 in code like we do now with the `if !noLimit` checks. So it wouldn't simplify the code at least. But the user experience I'm not sure of which way is better. I'll merge it as-is since this was resolved, but if there's a strong preference either way before we do a new rita release we can change it later.

PR #474, opened 2019-06-24
Rolling params

ethack:
This pull request has the following changes:
- ~~Moves `MetaDatabase.files` into `MetaDatabase.databases` and stores the file hashes in the database entry instead.~~
  - ~~Back of the envelope calculation shows that there is plenty of room for file hash entries (about 46k) and we'd run into problems storing them unsorted long before that.~~
  - ~~Additionally, the files are stored with the chunk ID they were imported with so that they can be removed along with the rest of the data in rolling databases.~~
  - ~~It also makes deleting these file entries much easier as they are all inside the metadatabase databases entry. In the past we've forgotten to remove the files from the separate collection but this takes care of that issue.~~
  - I reverted this change and moved the files back into `MetaDatabase.files`. The reason is that we don't want to remove the file records from the rolling database even after they've been removed because we don't ever want to let the same file get imported again into a rolling database. And a separate collection makes more sense than an array field since it can grow unchecked. We probably don't want this to grow forever though. I thought maybe a capped collection would work but we can't delete entries manually from those.
    - The reason I reverted this change still makes sense, though it was a blocker for me when I reverted it back. Since then I reworked my dependent code so that it doesn't matter to me which way this is implemented. But before I make changes _again_ I wanted to get feedback from someone else.
- Several logging statements were not using the configured logger and thus were printing out log messages to the screen when they should only have been going to the database. This has been fixed.
- All rolling arguments are optional. The command still works the same way it did previously, except in cases where it would complain about missing options before it uses sane defaults instead. 
  - The total number of chunks defaults to 24 and this default value can be changed in the config file.
  - The current chunk ID will default to 0 for new databases. For existing databases, it will read the chunk id from the database and increment by one (wrapping at total chunks), if not specified on the command line.
  - The --rolling argument can still be specified manually. But if left off the database will still be set up to support rolling operations, it just has its metadatabase rolling flag set to false. Any subsequent imports to the same database will automatically convert the non-rolling database to a rolling database. This allows users to import data into an already existing database even if they didn't specify it as --rolling to begin with. If a database gets auto-converted there will be a message printed to the user and logged stating that this was done.
- In addition to the old `rita import <directory> <db name>` syntax, the `<directory>` portion has been extended to take any number of files or directories. This allows the caller to specify individual files to import rather than having all files in a directory automatically be included. This change is mostly to help give more control over which data is imported when dealing with rolling databases.

I wrote unit tests for the `setRolling` function because it was the place I thought most likely to have something go wrong. (And because it was the easiest code to decouple from external dependencies.)


commands/import.go
```diff
@@ -83,76 +88,84 @@ func (i *Importer) parseArgs() error {
 		return cli.NewExitError(err.Error(), -1)
 	}
 
-	// check if rolling flag was passed in
-	if i.rolling {
```
ethack: This logic is either removed altogether or moved into the setRolling function.

commands/import.go
```diff
@@ -83,76 +88,84 @@ func (i *Importer) parseArgs() error {
...
-	}
-
 	i.res = resources.InitResources(i.configFile)
 
 	return nil
 }
 
-func (i *Importer) setTargetDatabase() error {
-	i.res.DB.SelectDB(i.targetDatabase)
+func checkFilesExist(files []string) error {
```
ethack: checkFilesExist replaces checkImportDirExists

commands/import.go
```diff
@@ -83,76 +88,84 @@ func (i *Importer) parseArgs() error {
...
+		}
+	}
 	return nil
 }
 
-func (i *Importer) checkImportDirExists() error {
+// setRolling determines what the current rolling and chunk settings should be
+// based on the values already in the database and the values supplied on the
+// command line by the user.
+func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks int,
```
ethack: This is the function that makes all the rolling arguments optional. It's a bunch of if-statements that make assumptions based on the data we have. The comments should explain well enough what it's doing and the unit tests should give a fair confirmation that it's working as it should.

commands/import.go
```diff
@@ -164,28 +177,41 @@ func (i *Importer) run() error {
...
 	}
 
-	i.res.Log.Infof("Importing %s\n", i.importDir)
-	fmt.Println("\n\t[+] Importing " + i.importDir + " :")
+	i.res.Log.Infof("Importing %v\n", i.importFiles)
+	fmt.Printf("\n\t[+] Importing %v:\n", i.importFiles)
+
+	// print out a message if we're automatically converting this to a rolling database
+	// i.e. it wasn't rolling before and the user didn't specify the --rolling flag
+	if !isRolling && !i.rolling && rollingCfg.Rolling {
```
ethack: @lisaSW had the idea that we should let users know when we're converting a database. There shouldn't be any risk in doing so, but it's different behavior from what we used to do (we used to exit with an error) so I agree it's good to let them know.

commands/import_test.go
```diff
@@ -0,0 +1,209 @@
...
+	// a command line value for --chunk or --numchunks
+	const blank int = -1
+	// these are used to help make the test table below (a little) more readable
+	const exists bool = true
+	const rolling bool = true
+	const returnsError bool = true
+	const default12 int = 12
+	const default24 int = 24
+
+	testCases := []tc{
```
ethack: I tried to balance being concise with readable. I know it's not great but I think it's better than using 10 lines for each test case to fill out the `tc` struct the long way.  I'm more than happy to explain what's going on here if anyone has questions.

Please feel free to suggest test cases you'd like to see that I missed.

database/meta.go
```diff
@@ -64,98 +63,61 @@ func NewMetaDB(config *config.Config, dbHandle *mgo.Session,
...
-//EnsureRollingSettingsMatch ensures that the rolling database settings in the
-//metadb match the provided arguments if the database settings exist at all
-func (m *MetaDB) EnsureRollingSettingsMatch(db string, numchunks int) error {
-	m.lock.Lock()
-	defer m.lock.Unlock()
-	ssn := m.dbHandle.Copy()
-	defer ssn.Close()
-
+//GetRollingSettings gets the current rolling settings
+func (m *MetaDB) GetRollingSettings(db string) (exists bool, isRolling bool, currChunk int, totalChunks int, err error) {
```
ethack: Since we're allow much more flexibility, this function simply returns the values from the database rather than do any logic like `EnsureRollingSettingsMatch` used to do.

database/meta.go
```diff
@@ -64,98 +63,61 @@ func NewMetaDB(config *config.Config, dbHandle *mgo.Session,
...
+	isRolling = result.Rolling
+	currChunk = result.CurrentChunk
+	totalChunks = result.TotalChunks
+	return
 }
 
 //SetRollingSettings ensures that a given db is marked as rolling,
 //ensures that total_chunks matches numchunks, and sets the current_chunk to chunk.
-func (m *MetaDB) SetRollingSettings(db string, numchunks int, chunk int) error {
+func (m *MetaDB) SetRollingSettings(db string, chunk int, numchunks int) error {
```
ethack: This function was updated to allow changing the rolling state and the total number of chunks of an existing database.

database/meta.go
```diff
@@ -178,8 +140,9 @@ func (m *MetaDB) LastCheck() (time.Time, semver.Version) {
 	return time.Time{}, semver.Version{}
 }
 
-// AddNewDB adds a new database to the DBMetaInfo table
-func (m *MetaDB) AddNewDB(name string) error {
+// AddNewDB adds a new database to the DBMetaInfo table. All new databases are
+// ready to be rolling databases.
+func (m *MetaDB) AddNewDB(name string, currentChunk, totalChunks int) error {
```
ethack: This was changed to put all the pieces necessary for a rolling database in right away. That way when converting to a rolling database later all we have to do is toggle the rolling flag from false to true.

parser/fileparser.go
```diff
@@ -35,89 +36,31 @@ func readDir(cpath string, logger *log.Logger) []string {
...
 		}
 	}
 	return toReturn
 }
 
-// readDirRolling recursively reads the directory looking for log and .gz files
-// that match the timestamps of the associated chunk
-func readDirRolling(currentChunk int, totalChunks int, cpath string, logger *log.Logger) []string {
+// readFiles reads the files and directories looking for log and gz files
+func readFiles(paths []string, logger *log.Logger) []string {
```
ethack: By allowing the user to specify individual files on the command line to import we can get rid of this problematic code that deals with parsing times out of filenames. It also lets us use the same code for both rolling and non-rolling imports.

parser/indexedfile.go
```diff
@@ -127,7 +127,7 @@ func indexFiles(files []string, indexingThreads int, res *resources.Resources) [
 					res.Log.WithFields(log.Fields{
 						"file":  files[j],
 						"error": err.Error(),
-					}).Warning("An error was encountered while indexing a file")
+					}).Debug("An error was encountered while indexing a file")
```
ethack: These log messages cluttered up the log files and we've never had much use for them. By switching this to debug it will still get logged to the database log collection but not put in the log files by default.

pkg/beacon/writer.go
```diff
@@ -56,7 +56,7 @@ func (w *writer) start() {
 
 				if err != nil ||
 					((info.Updated == 0) && (info.UpsertedId == nil)) {
-					log.WithFields(log.Fields{
+					w.log.WithFields(log.Fields{
```
ethack: There's a whole bunch of these types of changes. This was causing the messages to be printed out to stdout instead of going to the log file only.

Issue #278, opened 2018-11-28
Remove CentOS Bro install workaround


ethack:
It appears [this workaround](https://github.com/activecm/rita/blob/v1.1.0/install.sh#L184-L187) is no longer needed.  I was able to manually install Bro using the normal instructions without the workaround.

Original issue: https://github.com/activecm/rita/issues/189

PR #480, opened 2019-07-25
Removed workaround for bro install on CentOS

### LINKED ISSUE #189: RITA Installer Spins on CentOS 7
When attempting to install RITA on CentOS 7, the installer spins when trying to install Bro IDS.

It appears there is an issue with Bro's CentOS repo. When running the installation by hand, a message similar to 
```
Downloading packages:                                    
bro-2.5.3-1.1.x86_64.rpm       FAILED
http://download.opensuse.org/repositories/network%3A/bro/CentOS_7/x86_64/bro-2.5.3-1.1.x86_64.rpm: [Errno 12] Timeout on http://provo-mirror.opensuse.org/repositories/network:/bro/CentOS_7/x86_64/bro-2.5.3-1.1.x86_64.rpm: (28, '
Operation too slow. Less than 1 bytes/sec transferred the last 500 seconds')    
```
will be displayed. I have tested this both in a CentOS docker image and in a virtual machine and the same error occurs in both. 

EDIT: I have also tested this on an entirely different network and computer and received the same error.

carrohan:
removed line in code that was initially used to fix #189 because we've been able to manually install Bro on CentOS without that workaround (see #278)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -203,10 +203,6 @@ __install_bro() {
 			;;
 		CentOS|RedHatEnterpriseServer)
 			__add_rpm_repo http://download.opensuse.org/repositories/network:bro/CentOS_7/network:bro.repo
-			# Workaround for https://github.com/activecm/rita/issues/189
-			# Replace the download.opensuse.org link with downloadcontent.opensuse.org link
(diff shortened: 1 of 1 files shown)


lisaSW:
@ethack is this one good to merge or is it depreciated?

Issue #273, opened 2018-11-28
Documentation for RITA output


ethack:
Specifically, the beacon output has tables with obtuse labels, but other modules would benefit as well.

When you run "rita show-beacons", you get a nifty report with the column titles:
Score,Source,Destination,Connections,Avg Bytes,TS Range,DS Range,TS Mode,DS Mode,TS Mode Count,DS Mode Count,TS Skew,DS Skew,TS Dispersion,DS Dispersion,TS Duration

* How are the values generated?
* What values represent good or bad?  (example: score close to 1.0 is a beacon, score close to 0 is not).

carrohan:
Do we want a markdown document in the docs folder that runs through all the show commands, or on the --help for each show command, or something else?

Spriithy:
I would argue in favor of a Markdown document under the `docs` folder. I think that's the whole point of it !

Maybe out of topic here, but one could also expect to find some documentation on the internals / processes of the analyzer part (for beaconing).

PR #482, opened 2019-07-30
added document to explain measurements in RITA

carrohan:
Added a document that explains measurements in RITA
Based on documents I was given I wasn't sure if we wanted a document like this, or one that literally went through and explained the column titles given in show-* commands, if we wanted something else I can rework this into that easily

### DIFF
--- a/docs/Understanding RITA Measurements.md
+++ b/docs/Understanding RITA Measurements.md
@@ -1,0 +1,58 @@
+## Beacons
+
+Compromised systems often send homing signals, known as 'beacons,' back to command and control servers. Attackers use beacons to monitor access to compromised machines as well as to transmit instructions, data, and additional malicious code. Because beacon connections are more uniform than their human-generated counterparts, beacon analysis may be used to identify suspicious patterns caused by malicious activity on a system.
+
+#### Score (Total, Interval, and Data Size)
+
+**Total Score** is the average of interval score and data size score. The closer this score is to 100% the more likely that this is beaconing activity.
+**Interval Score** is the average of interval skew, interval dispersion, and interval duration. The closer this score is to 100%, the more the interval data indicates beaconing activity.
+**Data Size Score** is the average of data size skew, data size dispersion, and data size mode. The closer this score is to 100%, the more the data sizes being received indicates beaconing activity.
+
+#### Skew (Interval and Data Size)
+
+Skew measures how distorted or asymmetric the data is. A value closer to 100% means that the data is very symmetric. This measure is useful in the case of malware that does not try to hide itself, but more importantly will detect malware that tries to hide by adding jitter. This works because malware with jitter most likely uses a random number generator to add or subtract to a mean value (e.g. 30 seconds +/- 5 seconds). The random number generator will uniformly distribute the values which causes the data to be symmetric, and as such this particular measure is hard to beat.
+
+#### Dispersion (Interval and Size)
+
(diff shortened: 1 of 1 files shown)


ethack:
What you have is good too. I don't think the interval score and data size score are actually displayed in the beacon output, but we do store those in the database. I'm not sure if we should explain those in the doc or not.

But I would like something that goes through and literally explains the column titles in the show-* commands.  Maybe something like this could go under each section in the document (e.g. "## Beacons") before each of the in depth explanations. Each one could be a link (if applicable) to the relevant section in the document or have a short description if there is not section. If you need help figuring out how to get the links right let me know. I think there are examples in our other docs though.

The following are the column titles you will see in the output of `rita show-beacons`. You can click them for more information.

- Score (just link to the score section)
- Source IP - IP address that initiated the connections.
- Destination IP - IP address that received the connections.
- Connections - The total number of connections between the source and destination IPs.
- Avg Bytes - etc.
- Intvl Range
- Size Range
- Top Intvl
- Top Size
- Top Intvl Count
- Top Size Count
- Intvl Skew (just link to the section)
- Size Skew (just link to the section)
- Intvl Dispersion (just link to the section)
- Size Dispersion (just link to the section)

Spriithy:
Refers to #273 (just so it shows up in feed)

Spriithy:
Also I fully agree with @ethack I think it would provide more insights to document each column. As I suggested in the issue #273 I think a similar approach could be useful for the analyzers in order to help future / independant maintainers that might not have prior insider knowledge.

Spriithy:
As I dug through the code and tried to make sense of the several indicators and scores used in the analyzer I really wished I had some documentation to back my intuitions mostly regarding the choices made.

Why pick a 30 seconds in the computation of the `tsMadmScore` and use 32 seconds right after for `dsMadmScore` ?

Anyways, I think the indicators are all straightforward. Just the scores might need some explanation.

ethack:
@Spriithy Thanks for the feedback!

To answer your question about these lines,

https://github.com/activecm/rita/blob/d7f7b17928d02b003b9de02095acda33617924a8/pkg/beacon/analyzer.go#L164-L174

`ts` stands for timestamp and refers to the connection interval metrics. `ds` stands for data size and refers to the connection size metrics. In both cases the divisors are setting a value to normalize the score against. So anything with an interval dispersion greater or equal to 30 **seconds** will all have the same `tsMadmScore` of 0. Likewise, anything with a data size dispersion greater or equal to 32 **bytes** will have the same `dsMadmScore` of 0.

 I'm not entirely sure how 30 seconds and 32 bytes were picked. It could be they were just arbitrary choices that tended to work well.

Spriithy:
Thanks for the feedback ! Is there some sort of roadmap for the project ? Anything maybe we could contribute to ?

ethack:
No public roadmap :( But any issue marked ["good first issue"](https://github.com/activecm/rita/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) would be very helpful and contributions would be welcome. If you're interested in any of them just start commenting on the issue with questions or a proposed solution.

carrohan:
I updated the doc I'd written before to hopefully explain some of the headers better, but didn't have a chance to get it proof read. I'm not totally sure if it's what you were hoping for @Spriithy so please feel free to give any feedback (and apologies for taking so long to get back to this!)

thibaultbl:
I think an explanation is still needed to explain what is a good / bad score of beacon. 

- We do not understand immediatly that the score is not a probability.
- We do not understand what threshold (or even if using a threshold make sense) can be used to separate beacon and non-beacon.

PR #484, opened 2019-07-31
WIP: Updated handling of empty databases and out of sync MetaDatabase for html-report

### LINKED ISSUE #159: Gracefully handle out of sync MetaDatabase
One user manually transferred databases between MongoDB instances. They transferred their MetaDatabase, however did not transfer all the databases referenced in the MetaDatabase. This caused rita to error when run, specifically when running `rita html-report`. I'm not certain if this causes issues with `show-*` or `import` commands as well. Below is the sample error output.

```
rita html-report

[-] Writing: /data01/rita/reports/2017/12/03/rita-html-report1/rita-int-09-19
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5c54da]

goroutine 1 [running]:
gopkg.in/mgo%2ev2.(*Iter).Next(0x0, 0x8ae120, 0xc42024e700, 0x8ae120)
    /home/thealy/go/src/gopkg.in/mgo.v2/session.go:3684 +0x3a
gopkg.in/mgo%2ev2.(*Iter).All(0x0, 0x8a9ea0, 0xc4201e08e0, 0x0, 0x0)
    /home/thealy/go/src/gopkg.in/mgo.v2/session.go:3800 +0x2b6
github.com/ocmdev/rita/reporting.printBeacons(0xc4201d0a50, 0xe, 0xc4201cc320, 0x0, 0x0)
    /home/thealy/go/src/github.com/ocmdev/rita/reporting/report-beacons.go:29 +0x370
github.com/ocmdev/rita/reporting.writeDB(0xc4201d0a50, 0xe, 0xc4201a2e80, 0x31, 0xc4201cc320, 0x11, 0x4)
    /home/thealy/go/src/github.com/ocmdev/rita/reporting/report.go:170 +0x291
github.com/ocmdev/rita/reporting.PrintHTML(0xc420066d80, 0x5, 0x8, 0xc4201cc320, 0x0, 0x0)
    /home/thealy/go/src/github.com/ocmdev/rita/reporting/report.go:83 +0x6e2
github.com/ocmdev/rita/commands.init.4.func1(0xc4200c91e0, 0x0, 0xc4200c91e0)
    /home/thealy/go/src/github.com/ocmdev/rita/commands/reporting.go:31 +0x129
github.com/urfave/cli.HandleAction(0x8d7360, 0x99fc90, 0xc4200c91e0, 0xc4200b3000, 0x0)
    /home/thealy/go/src/github.com/urfave/cli/app.go:501 +0xd4
github.com/urfave/cli.Command.Run(0x9831e1, 0xb, 0x0, 0x0, 0x0, 0x0, 0x0, 0x9942aa, 0x29, 0x0, ...)
    /home/thealy/go/src/github.com/urfave/cli/command.go:200 +0x93f
github.com/urfave/cli.(*App).Run(0xc4200e81c0, 0xc42000c3a0, 0x2, 0x2, 0x0, 0x0)
    /home/thealy/go/src/github.com/urfave/cli/app.go:259 +0x7b7
main.main()
    /home/thealy/go/src/github.com/ocmdev/rita/rita.go:26 +0x110
```

The suggestion is to more gracefully handle this situation. Suggestions include:
1. Catch this error or detect this out of sync scenario and print a more useful error message.
2. Ignore "missing" databases even if they are referenced by the Metadatabase.
3. Detect an out of sync Metadatabase and offer to fix it by removing the entries.

Personally, I like option (1) the best. If we're in this situation it's definitely unexpected and likely something the user did to cause it. We should let the user know something is wrong rather than trying to silently handle things. It is up to the user to fix but a more useful error message could at least put them on the right track.

carrohan:
Previously, an out of sync MetaDatabase, or empty database on html-report would just throw a small error, and if multiple databases had an html-report being created no database after would have a report created.
Updated that so now an error message shows on the command line, and the empty/non-existent database page encouraging the user to delete the database.
Closes #159


reporting/report-bl-dest-ips.go
```diff
@@ -35,6 +36,9 @@ func printBLDestIPs(db string, res *resources.Resources) error {
 	if err != nil {
 		return err
 	}
+	if len(w) == 0 {
+		return cli.NewExitError("No results were found for " + db, -1)
```
ethack: Previously, if there were no results it would show a page with the header menu and an empty table like this:

image

Now, it shows a completely empty page with no menu. I could post a screenshot but it would be pure white :)

I'd like to still get the previous page that had the menu at the top so you could keep clicking through to other modules. This applies to every module.

Beacons seems to be the only module that previously would give an empty white page. So beacons should be changed from its previous behavior to have the menu like the others.

reporting/report.go
```diff
@@ -75,7 +76,14 @@ func PrintHTML(dbsIn []string, res *resources.Resources) error {
 	for k := range dbs {
 		err = writeDB(dbs[k], wd, res)
 		if err != nil {
-			return err
+			if err.Error() == ("No results were found for " + dbs[k]) {
+				fmt.Println("[!] " + err.Error())
+				fmt.Println("    This might indicate that this database has been dropped, but the metadatabase hasn't been updated.")
+				fmt.Println("    To update the metadatabase and fix this problem, run:")
+				fmt.Println("    rita delete-database " + dbs[k])
```
ethack: This might indicate that the database has been dropped. However, there could be other reasons. Either of these will also cause this issue:
* if the database exists but is empty
* if the database has data but it's from a previous, incompatible version of rita

Since there could be other reasons, I'm a little worried about recommending to delete the database to fix the problem. It should definitely work, but they may lose something they wanted to keep, especially if there are other situations this can come up that we don't know of. Do you think it would be better to just get rid of the recommendation and the `delete-database` command, or might it be better to expand further and say something like "You can either disregard this message or attempt to fix it by deleting the metadatabase entry. Warning: this will also delete any data in the specified database: <delete command here>"?

carrohan: I figure we could list a few of the potential reasons, and then give the recommendation with warning like you mentioned. Maybe something like:
"This might indicate that this database has been dropped without the metadatabase being updated, the database exists but is empty, or the data in the database is from an incompatible version of rita. You can either disregard this message, or attempt to fix it by deleting the metadatabase entry by running the following command:
rita delete-database <db_name>
WARNING: this will also delete any data in the specified database."

ethack: It's getting wordy but it's not like you will see this often. We can go with that.

Zalgo2462:
I am closing this PR due to inactivity. I will keep the related issue open in case we want to address it in the future.

PR #488, opened 2019-08-14
Allowing databases to increase the number of chunks

### LINKED ISSUE #485: Allow changing the number of chunks of an existing rolling database
At the very least it should be easy to implement increasing the number of chunks. The side effect is that the database may have more data in it than expected.  For instance, if you had a previous database with 12 chunks of 2 hours each and you increased the size to hold 24 chunks and expect each to hold 1 hour.  Then when you increase the size you'll already have (12\*2=) 24 hours of data in the database and with 12 chunks filled and 12 empty. When you fill the remaining 12 chunks you'll have (12\*2 + 12\*1=) 36 hours of data in the database. Thereafter every chunk imported will decrease the number of hours by one until you are left with (24*1=) 24 hours left again.

This change needs to be made in the following locations:
- https://github.com/activecm/rita/blob/v3.0.4/commands/import.go#L124-L130
- https://github.com/activecm/rita/blob/v3.0.4/database/meta.go#L103  - It's quite possible this is not necessary as it was working without this before. But it is probably safer to grow the array explicitly.

As [this comment](https://github.com/activecm/rita/pull/474#discussion_r301869724) suggests, I didn't notice any problems when this was implemented before, but I wanted to avoid any potential issues with changing from the current behavior.

ethack:
Closes #485

### DIFF
--- a/commands/import.go
+++ b/commands/import.go
@@ -121,10 +121,10 @@ func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks 
 	}
 
 	if userTotalChunks != -1 { // user gave the total number of chunks via command line
-		// it's currently an error to try to change the total number of chunks in an existing rolling database
-		if dbExists && dbIsRolling && dbTotalChunks != userTotalChunks {
+		// it's currently an error to try to reduce the total number of chunks in an existing rolling database
+		if dbExists && dbIsRolling && userTotalChunks < dbTotalChunks  {
 			return cfg, fmt.Errorf(
-				"\n\t[!] Cannot modify the total number of chunks [ %d ] in an existing database",
+				"\n\t[!] Cannot modify the total number of chunks in an existing database [ %d ]",
 				dbTotalChunks,
 			)
 		}
--- a/commands/import_test.go
+++ b/commands/import_test.go
@@ -114,10 +114,10 @@ func TestSetRolling(t *testing.T) {
 			exists, rolling, 1, 12, rolling, blank, blank, default12, cfg{0, rolling, 2, 12}, !returnsError},
 
 		tc{"rita import --rolling --chunk 0 --numchunks 24",
-			exists, rolling, 1, 12, rolling, 0, 24, default12, cfg{0, rolling, 0, 24}, returnsError},
+			exists, rolling, 1, 12, rolling, 0, 24, default12, cfg{0, rolling, 0, 24}, !returnsError},
 
 		tc{"rita import --numchunks 24",
-			exists, rolling, 1, 12, !rolling, blank, 24, default12, cfg{0, rolling, 2, 24}, returnsError},
+			exists, rolling, 1, 12, !rolling, blank, 24, default12, cfg{0, rolling, 2, 24}, !returnsError},
 
 		tc{"rita import --chunk 5 (default 12)",
 			exists, rolling, 1, 12, !rolling, 5, blank, default12, cfg{0, rolling, 5, 12}, !returnsError},
@@ -129,7 +129,7 @@ func TestSetRolling(t *testing.T) {
 			exists, rolling, 1, 12, !rolling, 12, blank, default24, cfg{0, rolling, 12, 12}, returnsError},
 
 		tc{"rita import --chunk 12 --numchunks 24",
-			exists, rolling, 1, 12, !rolling, 12, 24, default12, cfg{0, rolling, 12, 24}, returnsError},
+			exists, rolling, 1, 12, !rolling, 12, 24, default12, cfg{0, rolling, 12, 24}, !returnsError},
 
 		tc{"rita import --chunk -2",
 			exists, rolling, 1, 12, !rolling, -2, blank, default12, cfg{0, rolling, -2, 12}, returnsError},
@@ -145,10 +145,10 @@ func TestSetRolling(t *testing.T) {
 			exists, rolling, 11, 12, rolling, blank, blank, default12, cfg{0, rolling, 0, 12}, !returnsError},
 
 		tc{"rita import --rolling --chunk 0 --numchunks 24",
-			exists, rolling, 11, 12, rolling, 0, 24, default12, cfg{0, rolling, 0, 24}, returnsError},
+			exists, rolling, 11, 12, rolling, 0, 24, default12, cfg{0, rolling, 0, 24}, !returnsError},
 
 		tc{"rita import --numchunks 24",
-			exists, rolling, 11, 12, !rolling, blank, 24, default12, cfg{0, rolling, 12, 24}, returnsError},
+			exists, rolling, 11, 12, !rolling, blank, 24, default12, cfg{0, rolling, 12, 24}, !returnsError},
 
 		tc{"rita import --chunk 5 (default 12)",
 			exists, rolling, 11, 12, !rolling, 5, blank, default12, cfg{0, rolling, 5, 12}, !returnsError},
@@ -160,7 +160,7 @@ func TestSetRolling(t *testing.T) {
 			exists, rolling, 11, 12, !rolling, 12, blank, default24, cfg{0, rolling, 12, 12}, returnsError},
 
 		tc{"rita import --chunk 12 --numchunks 24",
-			exists, rolling, 11, 12, !rolling, 12, 24, default12, cfg{0, rolling, 12, 24}, returnsError},
+			exists, rolling, 11, 12, !rolling, 12, 24, default12, cfg{0, rolling, 12, 24}, !returnsError},
 
 		// rolling, current chunk 11, total chunks 24
 		tc{"rita import",
--- a/database/meta.go
+++ b/database/meta.go
@@ -95,27 +95,31 @@ func (m *MetaDB) SetRollingSettings(db string, chunk int, numchunks int) error {
 	ssn := m.dbHandle.Copy()
 	defer ssn.Close()
 
+	selector := bson.M{"name": db}
+	update := bson.M{
+		"rolling": true,
+		"current_chunk": chunk,
+		"total_chunks": numchunks,
+	}
+
 	if !result.Rolling {
 		m.log.Infof("The dataset [ %s ] is being converted to a rolling dataset.", db)
-	} else if result.TotalChunks != numchunks {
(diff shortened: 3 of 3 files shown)


commands/import.go
```diff
@@ -121,10 +121,10 @@ func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks
 	}
 
 	if userTotalChunks != -1 { // user gave the total number of chunks via command line
-		// it's currently an error to try to change the total number of chunks in an existing rolling database
-		if dbExists && dbIsRolling && dbTotalChunks != userTotalChunks {
+		// it's currently an error to try to reduce the total number of chunks in an existing rolling database
+		if dbExists && dbIsRolling && userTotalChunks < dbTotalChunks  {
```
ethack: Now the only thing that isn't allowed is decreasing the number of chunks in an already existing rolling database. This is because we have to decide how to handle the chunks that may already be in the database and need to be removed. Increasing isn't an issue as nothing is in those chunk positions yet.

commands/import_test.go
```diff
@@ -114,10 +114,10 @@ func TestSetRolling(t *testing.T) {
 			exists, rolling, 1, 12, rolling, blank, blank, default12, cfg{0, rolling, 2, 12}, !returnsError},
 
 		tc{"rita import --rolling --chunk 0 --numchunks 24",
-			exists, rolling, 1, 12, rolling, 0, 24, default12, cfg{0, rolling, 0, 24}, returnsError},
+			exists, rolling, 1, 12, rolling, 0, 24, default12, cfg{0, rolling, 0, 24}, !returnsError},
```
ethack: Updated the tests that increase the total chunks so that they no longer expect an error to be returned.

database/meta.go
```diff
@@ -95,27 +95,31 @@ func (m *MetaDB) SetRollingSettings(db string, chunk int, numchunks int) error {
...
 	if !result.Rolling {
 		m.log.Infof("The dataset [ %s ] is being converted to a rolling dataset.", db)
-	} else if result.TotalChunks != numchunks {
-		m.log.Warnf("The total chunk size for existing rolling dataset [ %s ] was set to [ %d ] and is being changed to [ %d ].",
+	} else if result.TotalChunks < numchunks {
+		m.log.Warnf("The total chunk size for existing rolling dataset [ %s ] was set to [ %d ] and is being increased to [ %d ].",
+			db, result.TotalChunks, numchunks)
+
+		// if the total chunks increaase grow cid_list while preserving original entries
+		for i := result.TotalChunks; i < numchunks; i++ {
```
ethack: This is based on the code from `SetChunk` later in this file. I originally tried to just call `SetChunk` in a loop but these both lock the mutex so it was a deadlock. Plus it would have generated a bunch of mongo queries instead of only one.

Technically, I don't think we need to do this. I was using it before without this and it worked fine. In general the `cid_list` gets incremented naturally for every new chunk that is added after the number of chunks are increased. You could feasibly import a chunk that is beyond what has been imported before (e.g. there are 12 existing chunks and you import into chunk 23). Without this for loop what happens is you get this:
```json
"cid_list" : [ { "set" : true }, { "set" : true }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, null, null, null, null, null, null, null, null, null, null, null, { "set" : true } ]
```
Note that Mongo just set the 24th element to the right value and filled in every array value between with `null`s. With the current code this shouldn't break because we're just checking if a chunk `set` = `true`. But I thought it was more explicit and more predictable to have the values be initialized to `{set: false}` here. I'm just trying to do this in the way least likely to break anything now or in the future.

database/meta.go
```diff
@@ -95,27 +95,31 @@ func (m *MetaDB) SetRollingSettings(db string, chunk int, numchunks int) error {
...
+			update["cid_list." + strconv.Itoa(i) + ".set"] = false
+		}
+	} else if result.TotalChunks > numchunks {
```
ethack: We still don't allow the total chunks to decrease. That's why there is still a TODO comment in this block. But the `setRolling` function in the import command prevents this function from being called with a decrease in total chunks.

ethack:
I first imported some files into chunk 1/12.
```bash
./rita import --numchunks 12 ignore/empire-raw/*.00:* chunk-test 

	[+] Importing [ignore/empire-raw/capture_loss.00:00:00-01:00:00.log ignore/empire-raw/communication.00:00:00-01:00:00.log ignore/empire-raw/conn.00:00:00-01:00:00.log ignore/empire-raw/conn-summary.00:00:00-01:00:00.log ignore/empire-raw/dns.00:00:00-01:00:00.log ignore/empire-raw/files.00:00:00-01:00:00.log ignore/empire-raw/http.00:00:00-01:00:00.log ignore/empire-raw/notice.00:00:00-01:00:00.log ignore/empire-raw/pe.00:00:00-01:00:00.log ignore/empire-raw/software.00:00:00-01:00:00.log ignore/empire-raw/ssl.00:00:00-01:00:00.log ignore/empire-raw/stats.00:00:00-01:00:00.log ignore/empire-raw/weird.00:00:00-01:00:00.log ignore/empire-raw/x509.00:00:00-01:00:00.log]:
	[+] Non-rolling database chunk-test will be converted to rolling
	[-] Parsing logs to: chunk-test ... 
	[-] Parsing ignore/empire-raw/http.00:00:00-01:00:00.log -> chunk-test
	[-] Parsing ignore/empire-raw/ssl.00:00:00-01:00:00.log -> chunk-test
	[-] Parsing ignore/empire-raw/conn.00:00:00-01:00:00.log -> chunk-test
	[-] Parsing ignore/empire-raw/dns.00:00:00-01:00:00.log -> chunk-test
	[-] Host Analysis:            1601 / 1601  [==================] 100 %
	[-] Uconn Analysis:           2883 / 2883  [==================] 100 %
	[-] Exploded DNS Analysis:    1359 / 1359  [==================] 100 %
	[-] Hostname Analysis:        1359 / 1359  [==================] 100 %
	[-] Beacon Analysis:          2883 / 2883  [==================] 100 %
	[-] UserAgent Analysis:       9 / 9  [==================] 100 %
	[-] Invalid Cert Analysis:    5 / 5  [==================] 100 %
	[-] Updating blacklisted peers ...
	[-] Indexing log entries ... 
	[-] Updating metadatabase ... 
	[-] Done!
```

Here's the database entry at this point. The `cid_list` has 12 entries.
```json
> db.databases.find({name:"chunk-test"})
{ "_id" : ObjectId("5d5358cd7c4b3e0ca166e74a"), "name" : "chunk-test", "analyzed" : true, "analyze_version" : "v3.0.4", "rolling" : true, "total_chunks" : 12, "current_chunk" : 0, "ts_range" : { "min" : NumberLong(1520834214), "max" : NumberLong(1520837991) }, "cid_list" : [ { "set" : true }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false } ] }
```

Next I imported into the next chunk while increasing the number of chunks (chunk 2/24).
```bash
$ ./rita import --numchunks 24 ignore/empire-raw/*.01:* chunk-test 

	[+] Importing [ignore/empire-raw/capture_loss.01:00:00-02:00:00.log ignore/empire-raw/communication.01:00:00-02:00:00.log ignore/empire-raw/conn.01:00:00-02:00:00.log ignore/empire-raw/conn-summary.01:00:00-02:00:00.log ignore/empire-raw/dns.01:00:00-02:00:00.log ignore/empire-raw/files.01:00:00-02:00:00.log ignore/empire-raw/http.01:00:00-02:00:00.log ignore/empire-raw/notice.01:00:00-02:00:00.log ignore/empire-raw/software.01:00:00-02:00:00.log ignore/empire-raw/ssl.01:00:00-02:00:00.log ignore/empire-raw/stats.01:00:00-02:00:00.log ignore/empire-raw/weird.01:00:00-02:00:00.log ignore/empire-raw/x509.01:00:00-02:00:00.log]:
	[-] Parsing logs to: chunk-test ... 
	[-] Parsing ignore/empire-raw/conn.01:00:00-02:00:00.log -> chunk-test
	[-] Parsing ignore/empire-raw/http.01:00:00-02:00:00.log -> chunk-test
	[-] Parsing ignore/empire-raw/dns.01:00:00-02:00:00.log -> chunk-test
	[-] Parsing ignore/empire-raw/ssl.01:00:00-02:00:00.log -> chunk-test
	[-] Host Analysis:            1471 / 1471  [==================] 100 %
	[-] Uconn Analysis:           2764 / 2764  [==================] 100 %
	[-] Exploded DNS Analysis:    1313 / 1313  [==================] 100 %
	[-] Hostname Analysis:        1313 / 1313  [==================] 100 %
	[-] Beacon Analysis:          2764 / 2764  [==================] 100 %
	[-] UserAgent Analysis:       9 / 9  [==================] 100 %
	[-] Invalid Cert Analysis:    11 / 11  [==================] 100 %
	[-] Updating blacklisted peers ...
	[-] Indexing log entries ... 
	[-] Updating metadatabase ... 
	[-] Done!
```

Here's the new metadatabse entry. Note that the `cid_list` has 24 entries now.
```json
> db.databases.find({name:"chunk-test"})
{ "_id" : ObjectId("5d5358cd7c4b3e0ca166e74a"), "name" : "chunk-test", "analyzed" : true, "analyze_version" : "v3.0.4", "rolling" : true, "total_chunks" : 24, "current_chunk" : 1, "ts_range" : { "min" : NumberLong(1520834214), "max" : NumberLong(1520841583) }, "cid_list" : [ { "set" : true }, { "set" : true }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false }, { "set" : false } ] }
```

And here's the log entry generated.
```
time="2019-08-13T19:43:28-05:00" level=warning msg="The total chunk size for existing rolling dataset [ chunk-test ] was set to [ 12 ] and is being increased to [ 24 ]."
```

PR #494, opened 2019-09-04
Update Security Onion link in documentation

meljbruno:
Update the link to documentation about Security Onion with RITA to point to the new version of Security Onion's wiki.


ethack:
Thanks!

PR #510, opened 2019-11-05
Add install.sh support for Ubuntu 18.04

bkonsela:
Only small changes here.

Added a conditional in the function `__gather_OS()` to add a new variable, `_MONGO_OS_CODENAME`. I then changed the `__install_mongodb()` function's reference from `$(lsb_release -cs)` to `${_MONGO_OS_CODENAME}`.

Hopefully this all looks good!

Thanks.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -410,6 +410,9 @@ __gather_OS() {
 		exit 1
 	fi
 
+	# There is no official repository for Mongo3.6 on Ubuntu 18.04 Bionic.
+	# Documentation recommends using the xenial keyword in repos to install Mongo 3.6:
+	# https://linuxconfig.org/how-to-install-latest-mongodb-on-ubuntu-18-04-bionic-beaver-linux
 	if [ "$_MONGO_OS_CODENAME" == "bionic" ]; then
 		_MONGO_OS_CODENAME="xenial"
 	fi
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -402,11 +402,17 @@ __install_rita() {
...
+	_MONGO_OS_CODENAME="$(lsb_release -cs)"
+
 	if [ "$_OS" != "Ubuntu" -a "$_OS" != "CentOS" -a "$_OS" != "RedHatEnterpriseServer" ]; then
 		printf "$_ITEM This installer supports Ubuntu, CentOS, and RHEL. \n"
 		printf "$_IMPORTANT Your operating system is unsupported."
 		exit 1
 	fi
+
+	if [ "$_MONGO_OS_CODENAME" == "bionic" ]; then
+		_MONGO_OS_CODENAME="xenial"
```
ethack: Could you add a comment as to why we need this? I think you said it's because we need Mongo 3.6 which isn't available in bionic but is in xenial, but otherwise reading the code it isn't obvious.

bkonsela: Thanks for the feedback. Yes, I can add comments. I will do so in a moment.

bkonsela: Hows that?

bkonsela: I should change "keyword" to "codename", my apologies. I should change that real quick.

RuneBergh:
Would be nice to have this working on 18.04 indeed :)

PR #511, opened 2019-11-06
Allow Users to Re-Import Logs

Zalgo2462:
This PR introduces the `--delete` flag to the `rita import` command. This flag deletes any existing data in the target import location before importing new data.

If the target database is not a rolling database, the flag is used as:
```
rita import --delete /path/to/file_or_dir_1 [/path/to/file_or_dir_n ...] target_database
```

If the target database is a rolling database, the CID for the chunk to be replaced ~must~ may be specified:
```
rita import --delete -CC N /path/to/file_or_dir_1 [/path/to/file_or_dir_n ...] target_database
```

If the CID is not specified and the target database is a rolling database, the latest chunk is replaced:
```
rita import --delete --rolling /path/to/file_or_dir_1 [/path/to/file_or_dir_n ...] target_database
```

The command will fail on existing rolling databases. This is because the CID for each indexed file in the files collection is not currently tracked. However, going forward this will not be a problem.

I tested this PR using the `rita-diff.py` script in the old ipfix-rita repository. I created a rolling database with four chunks, copied the database to a clean, backup database, and re-imported each chunk. Finally, I ran `rita-diff.py` against the datasets. The script revealed minor inconsistencies in the beacons results in the `ts.conns_score` field, but this inconsistency was tracked down to code outside of this PR. Once those inconsistencies were accounted for, there were none.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -55,7 +55,7 @@ var (
 	// before importing the target data
 	deleteFlag = cli.BoolFlag{
 		Name:  "delete, D",
-		Usage: "Indicates that the existing dataset should be deleted before re-importing. If the dataset is a rolling dataset, --chunk is required.",
+		Usage: "Indicates that the existing dataset should be deleted before re-importing. If the dataset is a rolling dataset and --chunk is not specified, the latest chunk will be replaced.",
 	}
 
 	rollingFlag = cli.BoolFlag{
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -51,6 +51,13 @@ var (
 		Usage: "Tests which databases would be deleted. Does not actually delete any data, nor prompt for confirmation",
 	}
 
+	// deleteFlag indicates whether any matching, existing data should be deleted
+	// before importing the target data
+	deleteFlag = cli.BoolFlag{
+		Name:  "delete, D",
+		Usage: "Indicates that the existing dataset should be deleted before re-importing. If the dataset is a rolling dataset, --chunk is required.",
+	}
+
 	rollingFlag = cli.BoolFlag{
 		Name:  "rolling, R",
 		Usage: "Indicates rolling import, which builds on and removes data to maintain a fixed length of time",
--- a/commands/delete-database.go
+++ b/commands/delete-database.go
@@ -110,7 +110,7 @@ func deleteDatabase(c *cli.Context) error {
 
 	// Iterate through databases to delete and delete them one by one
 	for _, database := range names {
-		dberr := deleteSingleDatabase(res, names, database, dryRun)
+		dberr := deleteSingleDatabase(res, database, dryRun)
 		if dberr != nil {
 			return cli.NewExitError(dberr.Error, -1)
 		}
@@ -124,10 +124,13 @@ func deleteDatabase(c *cli.Context) error {
 	return nil
 }
 
-func deleteSingleDatabase(res *resources.Resources, dbnames []string, db string, dryRun bool) error {
+func deleteSingleDatabase(res *resources.Resources, db string, dryRun bool) error {
 	// check if database exists
-	dbExists := util.StringInSlice(db, dbnames)
-
+	collNames, err := res.DB.Session.DB(db).CollectionNames()
+	if err != nil {
+		return err
+	}
+	dbExists := len(collNames) != 0
 	// check if metadatabase record for database exists
 	mDBExists := util.StringInSlice(db, res.MetaDB.GetDatabases())
 
--- a/commands/import.go
+++ b/commands/import.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -9,6 +10,7 @@ import (
 	"github.com/activecm/rita/pkg/remover"
 	"github.com/activecm/rita/resources"
 	"github.com/activecm/rita/util"
+	log "github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
@@ -67,6 +69,7 @@ func NewImporter(c *cli.Context) *Importer {
 	}
 }
 
+//parseArgs handles parsing the positional import arguments
 func (i *Importer) parseArgs() error {
 	if len(i.args) < 2 {
 		return cli.NewExitError("\n\t[!] Both <files/directory to import> and <database name> are required.", -1)
@@ -91,8 +94,6 @@ func (i *Importer) parseArgs() error {
 		return cli.NewExitError(err.Error(), -1)
 	}
 
-	i.res = resources.InitResources(i.configFile)
-
 	return nil
 }
 
@@ -105,11 +106,12 @@ func checkFilesExist(files []string) error {
 	return nil
 }
 
-// setRolling determines what the current rolling and chunk settings should be
-// based on the values already in the database and the values supplied on the
-// command line by the user.
-func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks int,
-	userIsRolling bool, userCurrChunk int, userTotalChunks int, cfgDefaultChunks int) (config.RollingStaticCfg, error) {
+// parseFlags validates the user supplied flags against the current state of
+// the target database and determines what settings should be set in the
+// rolling configuration
+func parseFlags(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks int,
+	userIsRolling bool, userCurrChunk int, userTotalChunks int, cfgDefaultChunks int,
+	deleteOldData bool) (config.RollingStaticCfg, error) {
 
 	cfg := config.RollingStaticCfg{}
 
@@ -118,11 +120,21 @@ func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks 
 		userIsRolling = true
 	}
 
+	// ensure the user specifies a rolling import if the database exists
+	// and is not a rolling database.
+	if !deleteOldData && (dbExists && !dbIsRolling) && !userIsRolling {
+		return cfg, errors.New(
+			"\t[!] New data cannot be imported into a non-rolling database. " +
+				"Run with --rolling to convert this database into a rolling database.",
+		)
+	}
+
+	// set cfg.TotalChunks
 	if userTotalChunks != -1 { // user gave the total number of chunks via command line
 		// it's currently an error to try to reduce the total number of chunks in an existing rolling database
 		if dbExists && dbIsRolling && userTotalChunks < dbTotalChunks {
 			return cfg, fmt.Errorf(
-				"\n\t[!] Cannot modify the total number of chunks in an existing database [ %d ]",
+				"\t[!] Cannot modify the total number of chunks in an existing database [ %d ]",
 				dbTotalChunks,
 			)
 		}
@@ -130,8 +142,13 @@ func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks 
 		// use the user-provided value
 		cfg.TotalChunks = userTotalChunks
 	} else { // user didn't specify the total number of chunks
+
 		if !dbExists && !userIsRolling {
 			// if the database doesn't exist and wasn't specified to be rolling
+			// then assume only one chunk
+			cfg.TotalChunks = 1
+		} else if deleteOldData && dbExists && !dbIsRolling && !userIsRolling {
+			// if the user is re-importing in to a non-rolling database
 			// then assume only one chunk
 			cfg.TotalChunks = 1
 		} else if dbExists && dbIsRolling {
@@ -144,13 +161,21 @@ func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks 
 		}
 	}
 
+	//set cfg.CurrentChunk
 	if userCurrChunk != -1 { // user gave the current chunk via command line
 		// use the user-provided value
 		cfg.CurrentChunk = userCurrChunk
 	} else { // user didn't specify the current chunk
 		if !dbExists {
-			// if the databse doesn't exist, then assume this is the first chunk
+			// if the database doesn't exist then assume this is the first chunk
 			cfg.CurrentChunk = 0
+		} else if deleteOldData && dbExists && !dbIsRolling {
+			// if the user wants to re-import into a non-rolling database
+			// then assume we want to replace the first chunk
+			cfg.CurrentChunk = 0
+		} else if deleteOldData && dbIsRolling {
+			// replace the last chunk if the user specified --delete but not --chunk
+			cfg.CurrentChunk = dbCurrChunk
 		} else {
 			// otherwise increment tne current value, wrapping back to 0 when needed
 			cfg.CurrentChunk = (dbCurrChunk + 1) % cfg.TotalChunks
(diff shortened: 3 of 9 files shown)


commands/commands.go
```diff
@@ -51,6 +51,13 @@ var (
 		Usage: "Tests which databases would be deleted. Does not actually delete any data, nor prompt for confirmation",
 	}
 
+	// deleteFlag indicates whether any matching, existing data should be deleted
+	// before importing the target data
+	deleteFlag = cli.BoolFlag{
+		Name:  "delete, D",
+		Usage: "Indicates that the existing dataset should be deleted before re-importing. If the dataset is a rolling dataset, --chunk is required.",
+	}
```
Zalgo2462: New flag definition.

commands/delete-database.go
```diff
@@ -124,10 +124,13 @@ func deleteDatabase(c *cli.Context) error {
 	return nil
 }
 
-func deleteSingleDatabase(res *resources.Resources, dbnames []string, db string, dryRun bool) error {
+func deleteSingleDatabase(res *resources.Resources, db string, dryRun bool) error {
 	// check if database exists
-	dbExists := util.StringInSlice(db, dbnames)
```
Zalgo2462: This was an unnecessary parameter. In each call site, the value passed as db was obtained by the iterating over the value passed as dbnames. I've replaced the ineffectual check with one that will check if there is any data in the given db.

commands/import.go
```diff
@@ -40,27 +41,29 @@ func init() {
...
-		threads        int
+		res             *resources.Resources
+		configFile      string
+		args            cli.Args
+		importFiles     []string
+		targetDatabase  string
+		deleteOldData   bool
+		userRolling     bool
+		userTotalChunks int
+		userCurrChunk   int
```
Zalgo2462: deleteOldData is a new bool which contains the deleteFlag result.

rolling, totalChunks, and currentChunk were prefixed with the tag `user` to help distinguish their values between the values loaded from the metadb later on.

commands/import.go
```diff
@@ -159,20 +157,21 @@ func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks
 		}
 	}
 
-	// already existing db is converted to rolling if it wasn't already
-	// or if the user specified that a new database is rolling
-	cfg.Rolling = dbExists || userIsRolling
+	cfg.Rolling = dbIsRolling || userIsRolling
```
Zalgo2462: This was an odd behavior that would convert any existing database to a rolling database.
This meant that if you ran an import command without any rolling parameters against an existing database, the cfg.Rolling flag would be set. This created problems with the `--delete` flag. 

If we run `rita import --delete existing_database` we do not want to be in rolling mode.

commands/import.go
```diff
@@ -159,20 +157,21 @@ func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks
...
+		cfg.CurrentChunk >= cfg.TotalChunks {
 		return cfg, fmt.Errorf(
 			"\n\t[!] Current chunk number [ %d ] must be 0 or greater and less than the total number of chunks [ %d ]",
 			cfg.CurrentChunk,
 			cfg.TotalChunks,
 		)
 	}
 
+	// preserve the default chunks setting (even though we don't use it after this currently)
+	cfg.DefaultChunks = cfgDefaultChunks
```
Zalgo2462: This was set outside of this function, cluttering up the main function. I moved it here since this function is the only one which writes to the config. However, I had to change the test cases to match this.

commands/import.go
```diff
@@ -187,21 +186,31 @@ func (i *Importer) run() error {
...
+	// set up the rolling configuration
 	// get settings from an existing database
 	exists, isRolling, currChunk, totalChunks, err := i.res.MetaDB.GetRollingSettings(i.targetDatabase)
 	if err != nil {
 		return cli.NewExitError(fmt.Errorf("\n\t[!] Error while reading existing database settings: %v", err.Error()), -1)
 	}
+	// if the user wants to re-import a chunk, make them specify which chunk
+	if i.deleteOldData && isRolling && i.userCurrChunk == -1 {
+		return cli.NewExitError("--chunk is required when 'import --delete' is ran against a rolling database", -1)
+	}
```
Zalgo2462: Enforce the description of the delete flag

commands/import.go
```diff
@@ -212,7 +221,7 @@ func (i *Importer) run() error {
 
 	// print out a message if we're automatically converting this to a rolling database
 	// i.e. it wasn't rolling before and the user didn't specify the --rolling flag
-	if !isRolling && !i.rolling && rollingCfg.Rolling {
+	if exists && !isRolling && rollingCfg.Rolling {
```
Zalgo2462: This would display at an inopportune time. I can't exactly remember what the issue was to be honest.

But this should really only display if we are converting an existing dataset which is what the change reflects.

commands/import.go
```diff
@@ -224,6 +233,31 @@ func (i *Importer) run() error {
 	return nil
 }
 
+func (i *Importer) handleDeleteOldData() error {
```
Zalgo2462: Actually run the pre import deletion.

commands/import_test.go
```diff
@@ -44,149 +44,149 @@ func TestSetRolling(t *testing.T) {
...
 		tc{"rita import --numchunks -2 (default 12)",
-			!exists, !rolling, 0, 0, !rolling, blank, -2, default12, cfg{0, rolling, 0, -2}, returnsError},
+			!exists, !rolling, 0, 0, !rolling, blank, -2, default12, cfg{12, rolling, 0, -2}, returnsError},
 
 		// existing database scenarios
 
 		// non-rolling, current chunk 0, total chunks 1
 		tc{"rita import",
-			exists, !rolling, 0, 1, !rolling, blank, blank, default12, cfg{0, rolling, 1, 12}, !returnsError},
+			exists, !rolling, 0, 1, !rolling, blank, blank, default12, cfg{12, !rolling, 1, 12}, !returnsError},
```
Zalgo2462: We don't want to default to rolling if we don't specify it.

Zalgo2462:
I forgot to include the whitelisting bit in this PR so I've marked it as a WIP. The changes will be small so feel free to start in on a review.

The whitelisting bit will be two small changes to delete-database.go:deleteSingleDatabase() and import.go:handleDeleteOldData() to remove the database from the whitelist collection if it exists.

commands/import.go
```diff
@@ -88,8 +94,6 @@ func (i *Importer) parseArgs() error {
 		return cli.NewExitError(err.Error(), -1)
 	}
 
-	i.res = resources.InitResources(i.configFile)
```
Zalgo2462: Moved into the main run function since this isn't parsing. This will end up dialing the database, etc.

commands/import.go
```diff
@@ -102,40 +106,51 @@ func checkFilesExist(files []string) error {
...
-	// }
-
 	// a user-provided value for either of the chunk options implies rolling
 	if userTotalChunks != -1 || userCurrChunk != -1 {
 		userIsRolling = true
 	}
 
+	// ensure the user specifies a rolling import if the database exists
+	// and is not a rolling database.
+	if !deleteOldData && (dbExists && !dbIsRolling) && !userIsRolling {
```
Zalgo2462: Force the user to specify the rolling flag or a chunk flag to convert an existing non-rolling dataset to a rolling dataset.

commands/import.go
```diff
@@ -102,40 +106,51 @@ func checkFilesExist(files []string) error {
...
 
 		// use the user-provided value
 		cfg.TotalChunks = userTotalChunks
 	} else { // user didn't specify the total number of chunks
+
 		if !dbExists && !userIsRolling {
 			// if the database doesn't exist and wasn't specified to be rolling
 			// then assume only one chunk
 			cfg.TotalChunks = 1
+		} else if deleteOldData && dbExists && !dbIsRolling && !userIsRolling {
```
Zalgo2462: Missed this in the old version. I'm not sure whether total chunks and current chunk matter if cfg.Rolling isn't set, but we might as well have a coherent config even if the values aren't currently used.

commands/import.go
```diff
@@ -146,28 +161,37 @@ func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks
...
+	//set cfg.CurrentChunk
 	if userCurrChunk != -1 { // user gave the current chunk via command line
 		// use the user-provided value
 		cfg.CurrentChunk = userCurrChunk
 	} else { // user didn't specify the current chunk
 		if !dbExists {
-			// if the databse doesn't exist, then assume this is the first chunk
+			// if the database doesn't exist then assume this is the first chunk
+			cfg.CurrentChunk = 0
+		} else if deleteOldData && dbExists && !dbIsRolling {
```
Zalgo2462: This covers the cases of `rita import --delete --rolling` and `rita import --delete` running against a non-rolling database.

commands/import.go
```diff
@@ -146,28 +161,37 @@ func setRolling(dbExists bool, dbIsRolling bool, dbCurrChunk int, dbTotalChunks
...
+			// if the user wants to re-import into a non-rolling database
+			// then assume we want to replace the first chunk
 			cfg.CurrentChunk = 0
+		} else if deleteOldData && dbIsRolling {
```
Zalgo2462: In this last iteration I changed the code so `--chunk` isn't required by `--delete`. If `--chunk` isn't specified, but `--rolling` or `-NC` are, the last chunk is replaced.

commands/import.go
Zalgo2462: I've updated the code to prevent upgrades unless the user specifies one of the rolling flags.  Since the user must set one of the rolling flags, `userIsRolling` will be true solving the issue here.

commands/import.go
Zalgo2462: Comment changed.

commands/import.go
```diff
@@ -224,6 +260,46 @@ func (i *Importer) run() error {
 	return nil
 }
 
+func (i *Importer) handleDeleteOldData() error {
+	if !i.res.Config.S.Rolling.Rolling {
+		fmt.Printf("\t[+] Removing database: %s\n", i.targetDatabase)
```
Zalgo2462: Added some print statements and some error handling to pretty this section up a bit.

In particular the call to `removerRepo.Remove` prints to the screen, so I wanted to make sure the something was printed for non rolling databases as well.

commands/import_test.go
```diff
@@ -1,33 +1,34 @@
...
+		dbIsRolling      bool
+		dbCurrChunk      int
+		dbTotalChunks    int
+		userIsRolling    bool
+		userCurrChunk    int
+		userTotalChunks  int
 		cfgDefaultChunks int
-		expected cfg
-		err bool
+		deleteOldData    bool
```
Zalgo2462: Added deleteOldData and 20 ish test cases

commands/import.go
```diff
@@ -184,24 +208,37 @@ func (i *Importer) run() error {
...
-	// preserve the default chunks setting (even though we don't use it after this currently)
-	rollingCfg.DefaultChunks = i.res.Config.S.Rolling.DefaultChunks
 	i.res.Config.S.Rolling = rollingCfg
 
+	if i.deleteOldData {
+		err := i.handleDeleteOldData()
+		if err != nil {
+			return cli.NewExitError(fmt.Errorf("Error deleting old data: %v", err.Error()), -1)
+		}
+	}
```
ethack: When the `--delete` flag is specified I'd like to have the database be deleted as late as possible. Right now, there are several places where errors can happen after the database is deleted that cause the user to have their database removed and no new one to replace it.

```bash
# I recreated the database rtest before running each of these

# 1) trying to import when InternalSubnets is not defined
$ ./rita import --delete --config etc/rita.yaml ignore/dnscat-tsv/conn.log rtest
	[+] Removing database: rtest
	[-] Successfully deleted database rtest.

Internal subnets are not defined. Please set the InternalSubnets section of the config file.

# 2) trying to import a directory that exists but is empty, or only contains unsupported log files
$ ./rita import --delete ignore/empty rtest
	[+] Removing database: rtest
	[-] Successfully deleted database rtest.

	[+] Importing [ignore/empty]:

	[!] No compatible log files found

# 3) similar to the last one, trying to import a log file that is unsupported
$ ./rita import --delete ignore/dnscat-tsv/weird.log rtest
	[+] Removing database: rtest
	[-] Successfully deleted database rtest.

	[+] Importing [ignore/dnscat-tsv/weird.log]:
	[!] All files in this directory have already been parsed into database:  rtest
```

Fixing 1 should be as easy as moving this block below the check for internal subnets on lines 243-245.

Fixing 2 and 3 is trickier but doable. I think if we move [these lines](https://github.com/activecm/rita/blob/508-reimport/parser/fsimporter.go#L82-L108) to their own function and call it before we delete the database/chunk it would fix those.

commands/import.go
```diff
@@ -184,46 +208,104 @@ func (i *Importer) run() error {
...
 		return cli.NewExitError(err.Error(), -1)
 	}
-	// preserve the default chunks setting (even though we don't use it after this currently)
-	rollingCfg.DefaultChunks = i.res.Config.S.Rolling.DefaultChunks
 	i.res.Config.S.Rolling = rollingCfg
 
 	importer := parser.NewFSImporter(i.res, i.threads, i.threads, i.importFiles)
 	if len(importer.GetInternalSubnets()) == 0 {
 		return cli.NewExitError("Internal subnets are not defined. Please set the InternalSubnets section of the config file.", -1)
 	}
```
ethack: Moved InternalSubnets check ahead of the delete handler to fix 1.
```
$ ./rita import --delete --config etc/rita.yaml ignore/dnscat-tsv/conn.log rtest

Internal subnets are not defined. Please set the InternalSubnets section of the config file.
```

commands/import.go
```diff
@@ -184,46 +208,105 @@ func (i *Importer) run() error {
...
 
+	indexedFiles := importer.CollectFileDetails()
+	// if no compatible files for import were found, exit
+	if len(indexedFiles) == 0 {
+		return cli.NewExitError("No compatible log files found", -1)
+	}
```
ethack: Added check here to make sure some compatible files are actually found. This fixes 2 and 3 (along with a piece elsewhere).
```
$ ./rita import --delete ignore/empty rtest

No compatible log files found

$ ./rita import --delete ignore/unsupported/weird.log rtest

No compatible log files found

$ ./rita import --delete ignore/unsupported rtest

No compatible log files found
```

parser/fsimporter.go
```diff
@@ -77,35 +77,17 @@ func (fs *FSImporter) GetInternalSubnets() []*net.IPNet {
...
 
-//Run starts the importing
-func (fs *FSImporter) Run() {
-	// track the time spent parsing
-	start := time.Now()
-	fs.res.Log.WithFields(
-		log.Fields{
-			"start_time": start.Format(util.TimeFormat),
-		},
-	).Info("Starting filesystem import. Collecting file details.")
```
ethack: I ended up removing these log statements. I had them in but they cluttered up the calling function and I started thinking that these particular ones aren't all that valuable currently anyway. We haven't been concerned about the performance of this code in quite some time.

parser/fsimporter.go
```diff
@@ -77,35 +77,17 @@ func (fs *FSImporter) GetInternalSubnets() []*net.IPNet {
...
 	// hash the files and get their stats
-	indexedFiles := indexFiles(files, fs.indexingThreads, fs.res)
-
-	// if no compatible files for import were found, handle error
-	if !(len(indexedFiles) > 0) {
-		fmt.Println("\n\t[!] No compatible log files found")
-		return
-	}
+	return indexFiles(files, fs.indexingThreads, fs.res)
+}
```
ethack: After the logging statements are gone and the actual check for files returned was moved to the calling function this ended up being very short. Almost not worth having a separate function for... almost.

parser/indexedfile.go
```diff
@@ -138,7 +139,16 @@ func indexFiles(files []string, indexingThreads int, res *resources.Resources) [
...
 	indexingWG.Wait()
-	return output
+
+	// remove all nil values from the slice
+	indexedFiles := make([]*fpt.IndexedFile, 0, len(output))
+	for _, file := range output {
+		if file != nil {
+			indexedFiles = append(indexedFiles, file)
+		}
+	}
```
ethack: This is the piece that fixed 3. The code above pre-allocates a slice in order to make it work with goroutines. But some of the routines would end up leaving nil in the slice. Previously, this slice, nils and all, would get returned and handled elsewhere. Now the nils get stripped out, leaving an empty slice if no valid files were found. This makes it easier to check if there were files we could import or not earlier.

parser/indexedfile.go
```diff
@@ -156,11 +166,6 @@ func removeOldFilesFromIndex(indexedFiles []*fpt.IndexedFile,
 	}
 
 	for _, newFile := range indexedFiles {
-		if newFile == nil {
-			//this file was errored on earlier, i.e. we didn't find a tgtDB etc.
-			continue
-		}
```
ethack: This is where the nils were handled previously, which is no longer needed.

PR #513, opened 2019-11-13
Add support for Zeek JSON logs

### LINKED ISSUE #146: Add support for Bro logs in JSON format
We log Bro files directly into JSON format for easier downstream parsing by logstash.  Add a parser to import bro logs that are logged natively in JSON format.

ethack:
This change implements parsing JSON logs in rita.

I've tested this with a single dataset so far but the results look promising. It is a bit slower than import TSV but I'm not sure there's much we can do about that. Here are the results from my comparison of importing the same dataset in TSV format vs JSON. Note that all the number of items being imported are the same. Additionally, I checked the log file after the runs and didn't notice any errors.

```bash
$ time ./rita import ignore/dnscat-tsv dnscat-tsv

	[+] Importing [ignore/dnscat-tsv]:
	[-] Parsing logs to: dnscat-tsv ... 
	[-] Parsing ignore/dnscat-tsv/conn.log -> dnscat-tsv
	[-] Parsing ignore/dnscat-tsv/ssl.log -> dnscat-tsv
	[-] Parsing ignore/dnscat-tsv/http.log -> dnscat-tsv
	[-] Parsing ignore/dnscat-tsv/dns.log -> dnscat-tsv
	[-] Host Analysis:            4432 / 4432  [==================] 100 %
	[-] Uconn Analysis:           14764 / 14764  [==================] 100 %
	[-] Exploded DNS Analysis:    65529 / 65529  [==================] 100 %
	[-] Hostname Analysis:        65529 / 65529  [==================] 100 %
	[-] Beacon Analysis:          14764 / 14764  [==================] 100 %
	[-] UserAgent Analysis:       16 / 16  [==================] 100 %
	[-] Invalid Cert Analysis:    8 / 8  [==================] 100 %
	[-] Updating blacklisted peers ...
	[-] Indexing log entries ... 
	[-] Updating metadatabase ... 
	[-] Done!

./rita import ignore/dnscat-tsv dnscat-tsv  53.11s user 19.66s system 263% cpu 27.578 total
```

```bash
$ time ./rita import ignore/dnscat-json dnscat-json

	[+] Importing [ignore/dnscat-json]:
	[-] Parsing logs to: dnscat-json ... 
	[-] Parsing ignore/dnscat-json/http.log -> dnscat-json
	[-] Parsing ignore/dnscat-json/ssl.log -> dnscat-json
	[-] Parsing ignore/dnscat-json/conn.log -> dnscat-json
	[-] Parsing ignore/dnscat-json/dns.log -> dnscat-json
	[-] Host Analysis:            4432 / 4432  [==================] 100 %
	[-] Uconn Analysis:           14764 / 14764  [==================] 100 %
	[-] Exploded DNS Analysis:    65529 / 65529  [==================] 100 %
	[-] Hostname Analysis:        65529 / 65529  [==================] 100 %
	[-] Beacon Analysis:          14764 / 14764  [==================] 100 %
	[-] UserAgent Analysis:       16 / 16  [==================] 100 %
	[-] Invalid Cert Analysis:    8 / 8  [==================] 100 %
	[-] Updating blacklisted peers ...
	[-] Indexing log entries ... 
	[-] Updating metadatabase ... 
	[-] Done!

./rita import ignore/dnscat-json dnscat-json  64.06s user 19.01s system 261% cpu 31.819 total
```

There are some additional things I want to test before removing the WIP.
- [x] Real world data from Security Onion
- [ ] Real world data from Corelight. The json log contains an extra field that indicates what type of log file it is. I'm not sure if the filename still has this too.
  - https://github.com/corelight/json-streaming-logs
- [x] Mixed logs where some are TSV and some are JSON
- [x] Make sure the `TimeStampFloat` field isn't being added into Mongo anywhere
- [ ] Test using the rita-diff.py script in the old ipfix-rita repository to compare imports from different file formats.

Closes #146


parser/indexedfile.go
```diff
@@ -55,22 +57,35 @@ func newIndexedFile(filePath string, res *resources.Resources) (*fpt.IndexedFile
...
+	var broDataFactory func() pt.BroData
+	if header.ObjType != "" {
+		// TSV log files have the type in a header
+		broDataFactory = pt.NewBroDataFactory(header.ObjType)
+	} else if scanner.Err() == nil && len(scanner.Text()) > 0 && // no error and there is text
+		json.Valid(scanner.Bytes()) {
+		// JSON log files only have the type in the filename
+		broDataFactory = pt.NewBroDataFactory(filepath.Base(toReturn.Path))
+		toReturn.SetJSON()
+	}
```
ethack: This is one of the main changes. Basically if the `scanHeader` call on line 53 above returns "" for the header type it *could* be because the log is in json. There are other reasons but that's why there is a check here to see if it is actually json. Next, json logs don't include any sort of header or indication on what type of log they are (e.g. conn, dns, http) so the only thing we have to go on is the filename.

parser/indexedfile.go
```diff
@@ -55,22 +57,35 @@ func newIndexedFile(filePath string, res *resources.Resources) (*fpt.IndexedFile
...
-		return toReturn, err
+	var fieldMap fpt.BroHeaderIndexMap
+	// there is no need for the fieldMap with JSON
+	if !toReturn.IsJSON() {
+		fieldMap, err = mapBroHeaderToParserType(header, broDataFactory, res.Log)
+		if err != nil {
+			fileHandle.Close()
+			return toReturn, err
+		}
+		toReturn.SetFieldMap(fieldMap)
```
ethack: Since json files don't have a header this block is unnecessary and causes an error.

parser/parsetypes/conn.go
```diff
@@ -5,58 +5,57 @@ import (
...
-		OrigIPBytes int64 `bson:"orig_ip_bytes" bro:"orig_ip_bytes" brotype:"count"`
-		// RespPkts counts response packets
-		RespPkts int64 `bson:"resp_pkts" bro:"resp_pkts" brotype:"count"`
-		// RespIpBytes gives the bytecount of response data
-		RespIPBytes int64 `bson:"resp_ip_bytes" bro:"resp_ip_bytes" brotype:"count"`
-		// TunnelParents lists tunnel parents
-		TunnelParents []string `bson:"tunnel_parents" bro:"tunnel_parents" brotype:"set[string]"`
-	}
-)
+// Conn provides a data structure for bro's connection data
```
ethack: Sorry for the large diff :frowning_face:   I reformatted the type definition in a couple of the files to match the others. 

The main change here is adding the `json` struct field tags and the `TimeStampFloat` field. Basically, we can't unmarshall json logs directly into the `TimeStamp` field because it is an `int64` and the json library calls `strconv.ParseInt()` on it, but it's a string with a decimal in it and the library returns an error similar to this:
```
time="2019-11-12T18:18:35-06:00" level=error msg="Encountered unparsable JSON in log" error="json: cannot unmarshal number 1517366387.201695 into Go struct field Conn.ts of type int64"
```

To get around this I use the `json:"-"` tag on the original `TimeStamp` field and the `bson:"-"` tag on the new `TimeStampFloat` field to have them ignore the respective fields when marshalling and unmarshalling.

parser/parsetypes/conn.go
```diff
@@ -65,3 +64,8 @@ func (in *Conn) TargetCollection(config *config.StructureTableCfg) string {
 func (in *Conn) Indices() []string {
 	return []string{"$hashed:id_orig_h", "$hashed:id_resp_h", "-duration", "ts", "uid"}
 }
+
+//ConvertFromJSON performs any extra conversions necessary when reading from JSON
+func (in *Conn) ConvertFromJSON() {
+	in.TimeStamp = int64(in.TimeStampFloat)
+}
```
ethack: The remaining piece is this function, which is called after unmarshalling the json. This takes the float field that `ts` was read into and typecasts it to the original `TimeStamp` field. If we have more issues like this in the future we can add more conversion steps to this function.

parser/fileparser.go
```diff
@@ -205,7 +206,21 @@ func mapBroHeaderToParserType(header *fpt.BroHeader, broDataFactory func() pt.Br
...
+	if isJSON {
+		err := json.Unmarshal([]byte(lineString), dat)
+		if err != nil {
+			logger.WithFields(log.Fields{
+				"error": err.Error(),
+			}).Error("Encountered unparsable JSON in log")
+		}
+		dat.ConvertFromJSON()
+		return dat
+	}
```
ethack: If the line is coming from a json file, we short circuit the tsv parsing logic and use the `encoding/json` library to unmarshal the string to an object. Then to take care of the timestamp discrepancy we call the new `ConvertFromJSON()` function.

ethack:
I had a chance to run this on some Security Onion logs. The only problem is that Security Onion changes the format of the timestamp field from unix to ISO8601.

So instead of this: `1517336042.279652` they look like this: `2019-11-13T09:00:01.932360Z`

Currently, this leads to similar errors I encountered before.
image

I'd say it's straightforward to switch this and import the new timestamp. What's not is supporting both date formats in the same type of file. So to summarize we can support unix timestamps in TSV and ISO8601 timestamps in JSON files. But supporting JSON with both date formats is not doable at the moment. I think that's ok as the majority use case for JSON support is because of Security Onion and that's a good starting point.

parser/parsetypes/conn.go
Zalgo2462: I wish we didn't have to include this function with the same definition in each parse type. It would be nice to have one method that could handle this for all of the types. We'd either have to use reflection or make sure to maintain a large typeswitch to accomplish that though. What are your thoughts?

parser/indexedfile.go
ethack: I can definitely rename the function. Is `scanTSVHeader` ok with you? The definition of scanHeader already has a note about advancing the file scanner, though I can edit it to match what you suggested. Do you want an additional note when it is called here?

parser/parsetypes/conn.go
ethack: I like that idea. I think I'll make that a separate issue/change though.

parser/parsetypes/conn.go
ethack: I wish it wasn't like this either. I explored some different options before settling on this one.

I'm not a huge fan of reflection and my impression is that it reduces performance. Though we already use it for the Bro log parsing and I think the json library uses it too. If we could get it to work I don't think it would be too convoluted or noticeably impact performance. And the `convertTimestamp` function already uses a type switch which I could argue as being a type of reflection.

It's not often we have to add a new type and my preference would be to leave it how it is. But be open to switching if we ever refactor the parser in the future.

parser/fileparser.go
ethack: I like the idea in theory. An issue I haven't worked out is how coupled fsimporter is with everything else in the parser package. Let me give it a shot and I'll come back if I think it will cause too many sweeping changes.

parser/parsetypes/conn.go
lisaSW: Agreed... Currently the affect on import time seems to be negligible, even with large datasets. This is the most consistent option we've found so far... if we refactor later maybe a new one will be available, but for now this is what works.

parser/parsetypes/http.go
```diff
@@ -10,73 +10,79 @@ type HTTP struct {
...
 //TargetCollection returns the mongo collection this entry should be inserted
-//into
-func (line *HTTP) TargetCollection(config *config.StructureTableCfg) string {
+func (in *HTTP) TargetCollection(config *config.StructureTableCfg) string {
 	return config.HTTPTable
 }
 
 //Indices gives MongoDB indices that should be used with the collection
-func (line *HTTP) Indices() []string {
+func (in *HTTP) Indices() []string {
```
lisaSW: Did these get changed to "in" to match with the others parsetypes? What if we just change all of them to "line", it seems to be clearer to me...

parser/parsetypes/parsetypes.go
```diff
@@ -1,14 +1,16 @@
...
 	"strings"
+
+	"github.com/activecm/rita/config"
 )
 
 //BroData holds a line of a bro log
 type BroData interface {
 	TargetCollection(*config.StructureTableCfg) string
 	Indices() []string
+	ConvertFromJSON()
```
lisaSW: Sorry if I'm being dense, but does this gets called for everything, regardless if the file was json or a bro log? Could we just split this into a different type? Maybe at least a clarifying comment on it would help for future development, I understand that it might be too difficult to split up.

parser/parsetypes/http.go
ethack: Yep, and can do!

parser/parsetypes/parsetypes.go
ethack: It's defined on the interface that is used for both tsv and json logs. But it's only called when a json log is imported, not a tsv one (https://github.com/activecm/rita/blob/146-json-support/parser/fileparser.go#L221).

I can certainly add a comment. Would `// ConvertFromJSON is called when importing from JSON logs` work, or did you have something in mind?

lisaSW: yup, that would be great, and could we add some comments to the parseline function if statements just to visually separate and clarify the two choices (json and bro)?

ethack:
@Zalgo2462 @lisaSW thanks for all the great feedback and the sample logs to test with!

I think I've taken care of most of the suggestions with the latest changes.

I agree that there could be some refactoring with the bro import part of the code. My goal was to add functionality without impacting the current code too much. I split up the `parseLine` body into two separate functions which I think should partially address @Zalgo2462's request and clarify the intent even better than comments as per @lisaSW's suggestion, as well as meet my desire to avoid unnecessary changes.

I looked into adding a `ParseLine` function to `IndexedFile` as @Zalgo2462 suggested but quickly found that `IndexedFile` was in a different package than the `indexedfile.go` file where `parseLine` is currently defined. Thinking through the dependencies going back and forth left me with the opinion it would be better left as-is rather than risk messing it up trying to patch it together. I'd much rather see us overhaul the package structure and interface for parsing Bro logs at a future date.

I also added in a check that should handle Corelight logs, at least as far as we can test with the json-streaming-logs plugin and the logs @Zalgo2462 provided.

```
$ ./rita import ignore/gmail-streaming-logs gmail-streaming

	[+] Importing [ignore/gmail-streaming-logs]:
	[-] Parsing logs to: gmail-streaming ... 
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_dns.1.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_conn.3.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_conn.1.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_conn.4.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_conn.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_conn.2.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_http.1.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_dns.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_http.2.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_dns.4.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_dns.2.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_dns.3.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_http.3.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_http.4.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_ssl.3.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_ssl.2.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_ssl.4.log -> gmail-streaming
	[-] Parsing ignore/gmail-streaming-logs/json_streaming_ssl.1.log -> gmail-streaming
	[-] Host Analysis:            17 / 17  [==================] 100 %
	[-] Uconn Analysis:           16 / 16  [==================] 100 %
	[-] Exploded DNS Analysis:    19 / 19  [==================] 100 %
	[-] Hostname Analysis:        19 / 19  [==================] 100 %
	[-] Beacon Analysis:          16 / 16  [==================] 100 %
	[-] UserAgent Analysis:       4 / 4  [==================] 100 %
	[-] Invalid Cert Analysis:    1 / 1  [==================] 100 %
	[-] Updating blacklisted peers ...
	[-] Indexing log entries ... 
	[-] Updating metadatabase ... 
	[-] Done!
```

PR #531, opened 2020-01-24
Support /var/log/bro/

william-stearns:


install.sh
```diff
@@ -210,7 +210,16 @@ __install_bro() {
...
 	__install_packages bro broctl
-	chmod 2755 /opt/bro/logs
+	if [ -d /opt/bro/logs/ ]; then
+		chmod 2755 /opt/bro/logs
+	elif [ -d /var/log/bro/ ]; then
+		mkdir -p /opt/bro/logs/
+		chmod 2755 /opt/bro/logs
+		mv -f /var/log/bro /var/log/bro.orig
+		cd /var/log
+		ln -s /opt/bro/logs bro
```
ethack: Could this just be `ln -s /opt/bro/logs /var/log/bro` without the `cd`ing? Or is there some nuance that I'm missing?

install.sh
```diff
@@ -210,7 +210,16 @@ __install_bro() {
 			;;
 	esac
 	__install_packages bro broctl
-	chmod 2755 /opt/bro/logs
+	if [ -d /opt/bro/logs/ ]; then
+		chmod 2755 /opt/bro/logs
+	elif [ -d /var/log/bro/ ]; then
+		mkdir -p /opt/bro/logs/
```
ethack: Could you add some comments as to why this is needed please? e.g. Something like (if accurate) `# RHEL and Ubuntu OS bro packages store logs in /var/log/bro. Link them to /opt/bro/logs to standardize the location.`

install.sh
william-stearns: I've had bad luck with that in the past.  I absolutely agree that it should do that, but sometimes it just doesn't make the link we both agree it should.  I've learned to not risk it, especially with a production shell script.

install.sh
william-stearns: Agreed, will add a comment to that effect.

PR #535, opened 2020-01-30
Terminal width human reporting (for exploded dns)

joswr1ght:
Wrap long domain names in human readable table view for show-exploded-dns.

This PR is a hack, shared to support a feature request. The human readable output of RITA is very difficult to parse on terminals where width is less than the total table size. I've changed the max domain name length per line to an arbitrary 80 characters, wrapping longer values to a new line. This change also requires the addition of a row separator line, to avoid confusion where a wrapped line could appear as a new record.

If there's a better way to represent or implement this, I'm all ears! 
before
after2
after1

### DIFF
--- a/commands/show-explodedDns.go
+++ b/commands/show-explodedDns.go
@@ -62,7 +62,8 @@ func init() {
 	bootstrapCommands(command)
 }
 
-func SplitSubN(s string, n int) []string {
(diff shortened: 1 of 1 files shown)


ethack:
Great idea for a change! The output looks really nice. I fixed the commits so only the one relevant change was included.

PR #536, opened 2020-01-31
Human readable duration for show-long-connections output

joswr1ght:
Change the duration value in `show-long-connections` to a human-readable duration.

This change introduces human readable duration to the output of `show-long-connections`. The old display was a number of seconds, which were sometimes represented as an exponential value, which is kind of hard for most users to parse easily. This change modified the duration to display time in a readable format. See attached screenshots.
old
new

### DIFF
--- a/commands/show-long-connections.go
+++ b/commands/show-long-connections.go
@@ -102,7 +102,7 @@ func showConns(connResults []uconn.LongConnAnalysisView) error {
 			result.Src,
 			result.Dst,
 			strings.Join(result.Tuples, " "),
-			duration(time.Duration(int(result.MaxDuration*float64(time.Second)))),
+			f(result.MaxDuration),
 		})
 	}
 	csvWriter.Flush()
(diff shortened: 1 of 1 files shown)


joswr1ght:
Sorry, I wasn't sure how to make the PRs independent. I thought branching off of RITA master would be enough. My bad!

lisaSW:
It usually is, but somehow the install script changes from your other branch must have slipped through on this one. Totally okay, I somehow manage to mess that up all the time :) We're just happy to have your help.

commands/show-long-connections.go
```diff
@@ -74,7 +102,7 @@ func showConns(connResults []uconn.LongConnAnalysisView) error {
 			result.Src,
 			result.Dst,
 			strings.Join(result.Tuples, " "),
-			f(result.MaxDuration),
+			duration(time.Duration(int(result.MaxDuration*float64(time.Second)))),
```
ethack: I'm not sure we want this in the csv output. It would make sorting or filtering difficult both on the command line or in a spreadsheet program.

lisaSW: Don’t most people use the regular non human readable output for parsing?

ethack: I would think so. I'm not sure what you're getting at, sorry.

lisaSW: Yep, you’re right, I missed that the csv output was being edited too

lisaSW: Yep, you’re right, I missed that the csv output was being edited too

joswr1ght:
Thank you!

PR #554, opened 2020-03-17
Install bro 2.5.5 for ubuntu xenial; Use ACM managed repos

Zalgo2462:
Bro 2.6 will not compile under Ubuntu Xenial in the OpenSuse Build Service. This breaks the bro installation on Xenial. I've copied the build files for Bro 2.6 and edited them to build Bro 2.5.5 for Xenial in OBS. 

~I have the script download the debs and install them manually so this one off repo isn't installed into RITA users' machines. This prevents the need to migrate users' repo configs once RITA migrates to using Zeek.~

This PR changes the installer to use ACM managed repos for both Bro 2.6 and Bro 2.5.5.

I've also removed a couple usages of sudo that were unnecessary since the script requires super user permissions up front. These sudo calls broke the script while I was testing in docker.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -197,42 +197,26 @@ __install_installer_deps() {
 __install_bro() {
 
 
-	if [ "$_OS" != "Ubuntu" -o "$_OS_CODENAME" != "xenial" ]; then
+	if  [ "$_OS" == "Ubuntu" ] && [ "$_OS_CODENAME" == "xenial" -o "$_OS_CODENAME" == "trusty" ] ; then
+		# Bro 2.6 will not compile under Debian 8/ Ubuntu Trusty/ Xenial in the OpenSuse Build Service
+		# Install Bro 2.5.5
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -195,17 +195,45 @@ __install_installer_deps() {
...
+				;;
+			CentOS|RedHatEnterprise|RedHatEnterpriseServer)
+				__add_rpm_repo http://download.opensuse.org/repositories/network:bro/CentOS_7/network:bro.repo
+				;;
+		esac
+		__install_packages bro broctl
+	else  # "$_OS" = "Ubuntu" -a "$_OS_CODENAME" = "xenial"
+		# Bro 2.6 will not compile under Ubuntu Xenial in the OpenSuse Build Service
+		# Manually download and install Bro 2.5.5 for Ubuntu Xenial
+		local tmpdir=`mktemp -d -q "$HOME/rita-install.XXXXXXXX" < /dev/null`
```
Zalgo2462: I copied this mktemp line from elsewhere in the install script, but I find it a bit weird to make the tmp directory in $HOME.

PR #564, opened 2020-05-11
Switch to Go modules

ethack:
Go modules were introduced in 1.11 and made the default versioning solution in 1.13. Now that [1.14 is out](https://blog.golang.org/go1.14) and says go modules are ready for production use we might as well switch over.

https://blog.golang.org/using-go-modules

Because we use makefiles anyway I don't think our workflow should be any different*. One main benefit is that we can have the source code live anywhere (no more messing with $GOPATH), which makes development easier, especially for community contributors not familiar with Go.

* But I could be missing things that will break people's existing dev workflows. So please try this out and report problems so we can either fix them or document the differences.


Makefile
```diff
@@ -16,15 +16,9 @@ cache = $(if $(cached-$1),,$(eval cached-$1 := 1)$(eval cache-$1 := $($1)))$(cac
 
 # force rita to be rebuilt even if it's up to date
 .PHONY: rita
-rita: vendor $(SRC)
+rita: $(SRC)
 	go build ${LDFLAGS}
 
-vendor: Gopkg.lock
```
ethack, line 22: From what I can tell go's builtin tooling takes care of the vendoring behind the scenes. So just running `go build` will automatically download missing dependencies rather than producing an error. This means our Makefile can be simpler.

docs/Manual Installation.md
```diff
@@ -31,23 +31,14 @@ You have a few options for installing RITA.
 
 ##### Installing Golang
 
-In order to compile RITA manually you will need to install both [Golang](https://golang.org) and [Dep](https://github.com/golang/dep).
```
ethack, line 34: Using go modules means we can remove a bunch of the steps that used to be necessary. This should get new contributors up to speed much easier and with fewer problems.

parser/indexedfile.go
```diff
@@ -155,7 +155,7 @@ func indexFiles(files []string, indexingThreads int, res *resources.Resources) [
 						"error": err.Error(),
 					}).Debug("An error was encountered while indexing a file")
 					//errored on files will be nil
-					fmt.Printf("\t[!] An error occured while indexing %v. Perhaps this log is empty?", files[j])
```
ethack, line 158: This is an unrelated change that caused the linter to fail. I fixed the spelling but the error will go away when/if https://github.com/activecm/rita/pull/563 is pulled in as well.

test.Dockerfile
```diff
@@ -1,21 +1,18 @@
...
 RUN apt-get update && apt-get install -y git make ca-certificates wget build-essential
 WORKDIR /go
-# install dep for vendoring
-RUN wget -q -O bin/dep https://github.com/golang/dep/releases/download/v0.5.3/dep-linux-amd64 \
-	&& chmod +x bin/dep \
 # install testing dependencies
-	&& wget -O - -q https://install.goreleaser.com/github.com/golangci/golangci-lint.sh \
-	| sh -s v1.16.0
+RUN wget -O - -q https://install.goreleaser.com/github.com/golangci/golangci-lint.sh \
+	| sh -s v1.26.0
```
ethack, line 9: I updated golangci-lint to the newest release while I was at it.

Zalgo2462:
Do we still need to set the GOPATH variable here?
https://github.com/activecm/rita/pull/564/files#diff-b67911656ef5d18c4ae36cb6741b7965R3

ethack:
Good catch! I don't believe we need GOPATH anywhere anymore so I'll remove it.

Zalgo2462:
The integration-test target is broken since it depends on vendor. We can fix it by just removing the dependency line 33 in the Makefile at https://github.com/activecm/rita/pull/564/files#diff-b67911656ef5d18c4ae36cb6741b7965R33

Zalgo2462:
`make install` fails due to the lack of the vendor directory. We can't easily copy the bash autocomplete file on install anymore. We can grab it from `$GOPATH/pkg/mod/urfave/cli@v1.20.0`, but that introduces a dependency on the GOPATH again and we'd have to grep the version out of the mod file.

Dockerfile
```diff
@@ -1,13 +1,12 @@
...
+FROM golang:1.14-alpine as rita-builder
 
 RUN apk add --no-cache git make ca-certificates wget build-base
-RUN wget -q -O /go/bin/dep https://github.com/golang/dep/releases/download/v0.5.3/dep-linux-amd64 && chmod +x /go/bin/dep
 
 WORKDIR /go/src/github.com/activecm/rita
 
 # cache dependencies
-COPY Gopkg.lock Gopkg.toml Makefile ./
-RUN make vendor
```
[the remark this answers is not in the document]
Zalgo2462: I trust that if it works with the current dependency set, that it would work with another one added. The docker build seems to work fine for me. If the dependencies weren't downloaded according the mod file, it would have broken. It helps that we always start from scratch in Docker, so there's less  to worry about.

Makefile
lisaSW: I updated to go 14.2 before testing this branch, and got a whole bunch of `go: inconsistent vendoring` errors. Do we need to figure out how to correctly transition existing go users to do this update?

parser/indexedfile.go
lisaSW: pulled into master and changed merged into this branch.

Makefile
ethack: I think I had the same thing. For me it was due to already having a `vendor` directory. Once I deleted that the issues went away. Do you think it's worth putting a `rm -rf vendor` statement in the makefile to automate this for people (assuming it also fixes it for you)? `go mod vendor` can also create a vendor directory but I don't see us having to use that so it should be safe to just delete it.

lisaSW: yes, that sounds like a perfect idea, lets add the removal step.

lisaSW: and it did fix it for me, thank you!

PR #573, opened 2020-07-19
Specify Output Delimiter

### LINKED ISSUE #572: Add Option to Specify Delimiter in Non-human-readable Output
Sometimes user agent strings contain commas which can make parsing the CSV output tricky; while the human-readable output can be parsed instead, it is needlessly annoying to do so. My suggestion is a `-d` flag that would allow the user to specify a string to use as a delimiter instead of a comma. A comma would still be the default delimiter if the `-d` flag was not provided. Although I wouldn't expect any of the other `show` commands to have this kind of issue, it might still be a nice option for them as well.

This seems like a relatively simple implementation, so I'll try to get it going myself; just opening the issue here in accordance with the [Contributing Guidelines](https://github.com/activecm/rita/blob/master/docs/RITA%20Gittiquette.md).

null-default:
Each of the `show` commands now supports a `-d` option that allows the user to specify a delimiter for the command's output. A couple of examples:
- `rita show-beacons -d ~ dataset_name` - Delimits the output by `~` instead of by `,`.
- `rita show-beacons -d "@@@" dataset_name` - Delimits the output by `@@@` instead of by `,`.

Under the hood, each `show` command's file now uses `fmt` to output instead of `encoding/csv`, and each output function now takes an additional `string` argument representing the delimiter. _This is still a comma by default._

The main `Readme.md` has been slightly updated to reflect this new functionality.

Closes #572

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -117,8 +117,11 @@ RITA cycles data into and out of rolling databases in "chunks". You can think of
       * `show-long-connections`: Print long connections and relevant information
       * `show-strobes`: Print connections which occurred with excessive frequency
       * `show-useragents`: Print user agent information
-  * By default RITA displays data in CSV format
+  * By default, RITA displays data in CSV format
+      * `-d [DELIM]` delimits the data by `[DELIM]` instead of a comma
+          * Strings can be provided instead of single characters if desired, e.g. `rita show-beacons -d "---" dataset_name`
       * `-H` displays the data in a human readable format
+          * This takes precedence over the `-d` option
       * Piping the human readable results through `less -S` prevents word wrapping
           * Ex: `rita show-beacons dataset_name -H | less -S`
   * Create a html report with `html-report`
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -111,6 +111,14 @@ var (
 		Name:  "connected, C",
 		Usage: "Show hosts which were connected to this blacklisted entry",
 	}
+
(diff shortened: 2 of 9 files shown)


null-default:
Additionally, in `show-beacons.go`, the following function names were changed in order to more closely resemble the other files' functions:
- `showBeaconReport() --> showBeaconsHuman()`
- `showBeaconCsv() --> showBeaconsDelim()`

ethack:
Looks great, thanks for the contribution!

All the output was identical when the delimiter was left as a comma. 
I also ran a script to compare output performance. The performance was slightly worse with the string printing, but only meaningful when I was doing loops of 100 on my dataset. The worst difference was the show-exploded-dns with 19.72s vs 14.86. I don't see this as anything worth optimizing but if it's needed in the future we have the option to use a string.Builder.

```
+ time bash -c 'for i in {1..100}; do ./rita-master show-exploded-dns --no-limit perf-test >/dev/null 2>&1; done'
14.86user 2.42system 1:02.90elapsed 27%CPU (0avgtext+0avgdata 42188maxresident)k
0inputs+0outputs (0major+731540minor)pagefaults 0swaps
+ time bash -c 'for i in {1..100}; do ./rita-null show-exploded-dns --no-limit perf-test >/dev/null 2>&1; done'
19.72user 3.80system 1:09.61elapsed 33%CPU (0avgtext+0avgdata 42720maxresident)k
0inputs+0outputs (0major+744929minor)pagefaults 0swaps
```

Issue #501, opened 2019-09-16
rita-bl collection isn't updated with custom blacklists


ethack:
When a custom blacklist is enabled [here](https://github.com/activecm/rita/blob/master/etc/rita.yaml#L107-L110) rita doesn't apply this to new data imported. If the `rita-bl` database is dropped then it does.

IPs added to an existing custom blacklist file are also not applied correctly. It's uncertain whether changes to the normal blacklist feeds are updated and applied correctly.

ethack:
Update from our call today:
- Adding a new custom blacklist to the rita config and then analyzing a new dataset successfully blacklisted the IP
- Adding a new IP to an existing custom blacklist and then analyzing a new dataset did not blacklist the IP
- Adding an additional custom blacklist and then analyzing a new dataset caused the IPs from the existing blacklist to not be blacklisted

PR #575, opened 2020-07-30
Always Update Custom Blacklists

Zalgo2462:
Currently a custom blacklist will only be updated a day after it was last updated. If a user creates a custom blacklist, runs rita with it, adds another item to the custom blacklist, and runs rita again within the day, the new item will not be checked against in the blacklist module see #501.

This PR sets the cache time for custom blacklists to 0, causing them to be reloaded on every run. This PR is supported by a [rita-bl PR](https://github.com/activecm/rita-bl/pull/8) which allows changing the cache time. *It must be pulled in first*, and then the go module file can be updated to point to the new version. 

It will take 1 day of running RITA for the fix to propagate itself on user machines as the current cache timeout will remain in place until the user's next blacklist update.

Long term, down the road we probably want to add some sort of change detection for the lists based on modification time or file hash. 

To test this PR:
- import a dataset with the current version of RITA with a *new* custom blacklist.
- clone https://github.com/activecm/rita-bl/pull/8 and adjust the redirect in the go.mod file to point to your local copy 
- build this version of RITA
- use mongo shell/ robo 3T to set the lastupdate field of the list in rita-bl.lists to 0 (or generally anything less than a day before the current time)
  - Note the cacheTime should be 86400 as that was the old default
  - Alternatively let a day pass IRL or adjust the system clock
- add a known value to the custom blacklist (search through hostname collection for a hostname, or grab a beacon ip for an ip)
- import the same dataset (under a new name) with the new version of RITA
- the new item should be in the blacklist results, and the cacheTime for the list in mongo should be 0

Using this PR I am able to add new ips/ hostnames to the custom blacklist files and have them show up in the rita results on my next run.


go.mod
```diff
@@ -6,39 +6,31 @@ go 1.14
 // should be updated in etc/bash_completion.d/rita
 // https://github.com/urfave/cli/blob/master/autocomplete/bash_autocomplete
 
+replace github.com/activecm/rita-bl => /home/logan/Development/ACM/rita-bl
```
Zalgo2462, line 9: I used this while working on both projects since code needed to be changed in both places.
This needs to be replaced/ removed after the rita-bl PR goes through.

PR #581, opened 2020-08-10
Update threat intel feeds

ethack:
See https://github.com/activecm/rita-bl/pull/9 for a full description.

### DIFF
--- a/config/static_test.go
+++ b/config/static_test.go
@@ -25,9 +25,8 @@ LogConfig:
 UserConfig:
     UpdateCheckFrequency: 14
 BlackListed:
-    myIP.ms: true
     MalwareDomains.com: true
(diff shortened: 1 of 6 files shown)


config/static_test.go
```diff
@@ -65,8 +64,7 @@ var testConfigFullExp = StaticCfg{
 		UpdateCheckFrequency: 14,
 	},
 	Blacklisted: BlacklistedStaticCfg{
-		UseIPms:            true,
-		UseMDL:             true,
```
Zalgo2462, line 69: Will the config parser crash if the user still has these fields defined? Would that force users to upgrade their config files?

ethack: Extra fields are ignored and doesn't cause any problems.

PR #586, opened 2020-09-09
Replace reflect with type assertions in import

ethack:
There's a slight speedup when testing with a small dataset. This savings should scale with the size of the dataset. In the output below `rita-reflect` represents the master branch and `rita-assert` represents these changes. I ran each import a couple times just to make sure the times were close.

```
./rita-reflect import ignore/empire-raw reflect  16.08s user 3.68s system 191% cpu 10.308 total
./rita-reflect import ignore/empire-raw reflect2  15.79s user 3.64s system 192% cpu 10.106 total
./rita-assert import ignore/empire-raw assert  14.80s user 3.50s system 192% cpu 9.505 total
./rita-assert import ignore/empire-raw assert2  14.62s user 3.72s system 181% cpu 10.080 total
```

But that's just the cherry on the top. The biggest benefit is increased code readability.

Thanks @Zalgo2462 for finding this!

### DIFF
--- a/parser/fsimporter.go
+++ b/parser/fsimporter.go
@@ -6,7 +6,6 @@ import (
 	"math"
 	"net"
 	"os"
-	"reflect"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


Zalgo2462 (approved):
LGTM

PR #591, opened 2020-10-20
Support Parsing Zeek Logs Collected By Multiple Remote Agents

Zalgo2462:
- Support parsing Zeek log entries with AgentUUID and AgentName fields as produced by Espy.

- Implement UniqueIP objects throughout RITA wich tag IP addresses with 128 bit UUIDs and human readable network names. This prevents local IP collisions when two computers share the same IP on distinct network segments.

	- newUniqueIP handles assigning a Network UUID and Network Name to a given IP address.

	- If the IP is publicly routable, the IP is assigned to the Public network.

	- Otherwise, if the log entry has agent fields, the IP is assigned to a network based on those fields. Currently, we directly assign the agent's hostname to the network name and the agent's UUID to the network UUID. However, a mapping might be used in the future to provide better control of the network assignments.

	- Finally, if there are no agent fields, the IP address is assigned to the Unknown Private network.

- When the `-network-names` flag is supplied, the `show-xxxx` and `html-report` commands display the name of the network associated with each IP address.

- Deduplicated the `getXXXXResultsView` logic in each of the `show-xxxx.go` and `reporting/*.go` files. The replacement methods now live in the `pkg` packages associated with each module.

- Changed minimum MongoDB version to 3.6.

- Implemented Zeek log separator parsing. (Needed for espy/zeek-data-mock)

- Remove dead code


pkg/uconn/analyzer.go
```diff
@@ -140,15 +144,20 @@ func (a *analyzer) hostMaxDurQuery(maxDur float64, localIP string, externalIP st
 	// create query
 	query := bson.M{}
 
+	//TODO: Why do we update old chunks with new max durations??? (https://github.com/activecm/rita/pull/512/files) -LL
```
Zalgo2462, line 147: This query updates max duration entries for a given ip pair, but it updates the entries within all of the available chunks. TBH I'm not sure how this query really works or where the consuming view logic is at.

pkg/useragent/analyzer.go
```diff
@@ -110,38 +114,36 @@ func (a *analyzer) start() {
...
-					} else {
+					_ = ssn.DB(a.db.GetSelectedDB()).C(a.conf.T.Structure.HostTable).Find(entryHostQuery).All(&hostEntries)
 
-						if res2[0].CID != a.chunk {
-							// fmt.Println("host existing", a.chunk, res2, data.Host)
-							newRecordFlag = true
-						}
+					//TODO: I think there might be a bug here. From what I can tell, if we set the new record flag
+					// when there is a rare signature record from a previous chunk, then we will end up pushing a duplicate
+					// record for each chunk. We should investigate removing this chunk check here. -LL
```
Zalgo2462, line 141: I suppose this is somewhat the same issue I have with the max duration query. In the max duration query, we update old chunks with new measurements. Here, we always push a new entry per chunk. How does the consuming logic work for this data, and could we standardize on method or the other?

ethack:
Fixing the integration tests involved re-organizing a little bit in 44f7ac593b321582eb26fb5892ad24bd9e3ba953. There were circular imports that I fixed but removing  `parser/ip.go` and merging it with `util/ip.go`. I think this makes sense structurally too as they are all similar functions that work with IP addresses. The only exception is that `NewUniqueIP()` went to `pkg/data/repository.go` because that's where the `UniqueIP` struct is defined and where `NewUniqueIPPair()` was already.

I also removed the `TestCreateIndexes` functions first because I was having problems that didn't end up being from those. But I left them out because I don't really see the value in them. If we go through and document a list of all the indexes we for sure need it would make some sense to have tests that verify the correct indexes are created. Even better would be if we made the list of indexes with a program that analyzes the queries made and recommends indexes and we test against that recommendation list. Anyway, I can put the functions back in if someone else thinks they have value. 

I was tempted to take the integration tests out as well as they don't have much value right now either. But I left them in to at least be examples for filling out the datatypes. 

I had problems with the `make integration-test` and had to modify it in order to get it to run. But I'm not sure if it was something just on my system so I didn't commit my changes there as they meant adding the manual step of starting a database.

commands/show-bl-ip.go
```diff
@@ -50,35 +51,30 @@ func init() {
...
-func parseBLArgs(c *cli.Context) (string, string, bool, bool, error) {
+func parseBLArgs(c *cli.Context) (string, string, bool, bool, bool, error) {
 	db := c.Args().Get(0)
 	sort := c.String("sort")
 	connected := c.Bool("connected")
 	human := c.Bool("human-readable")
+	showNetNames := c.Bool("network-names")
 	if db == "" {
-		return db, sort, connected, human, cli.NewExitError("Specify a database", -1)
+		return db, sort, connected, human, showNetNames, cli.NewExitError("Specify a database", -1)
```
fullmetalcache, line 61: Just a very minor suggestion but can we maybe consider having an errVal (or whatever you want to call it) in this particular case rather than the multiple returns? Just since we have so many params in each return. Will make future changes easier by only needing to modify the params in one place rather than three. So like:

`var errVal error = nil

if db == "" { 
   errVal = cli.NewExitError("Specify a database", -1)
} else if sort != "conn_count" && sort != "total_bytes" {
   errVal = cli.NewExitError("Invalid option passed to sort flag", -1)
}

return db, sort, connected, human, errVal`
}

It's not a big deal but just a suggestion

commands/show-bl-ip.go
```diff
@@ -177,9 +196,17 @@ func showBLIPs(ips []host.AnalysisView, connectedHosts, source bool, delim strin
...
 	table := tablewriter.NewWriter(os.Stdout)
-	headers := []string{"IP", "Connections", "Unique Connections", "Total Bytes"}
+	var headers []string
+	if showNetNames {
+		headers = []string{"IP", "Network"}
+	} else {
+		headers = []string{"IP"}
+	}
+
+	headers = append(headers, "Connections", "Unique Connections", "Total Bytes")
```
fullmetalcache, line 208: Some places we use "headers" and others we use "headerFields".  Might want to consider keeping them consistent, unless I am missing a fundamental difference between the two?

commands/show-bl-hostname.go
```diff
@@ -111,65 +125,25 @@ func showBLHostnamesHuman(hostnames []hostname.AnalysisView) error {
 			strconv.Itoa(entry.TotalBytes),
 		}
 
-		sort.Strings(entry.ConnectedHosts)
-		serialized = append(serialized, strings.Join(entry.ConnectedHosts, " "))
+		var sourceIPs []string
```
fullmetalcache, line 128: The logic here appears to essentially be the same as what is used in the showBLIPs function in the show-bl-ip.go function (around line 159).  The logic here uses an if/else with a for loop in each branch.  The logic in showBLIPs uses a single for loop with an if/else statement inside.  The former is likely faster as it doesn't need to evaluate the if/else at each loop.  The latter is only marginally more concise, imo...so I would lean towards using the former for consistency and a small speed boost. Your call though and not a big deal.

commands/show-long-connections.go
```diff
@@ -51,13 +48,13 @@ func init() {
 			}
 
 			if c.Bool("human-readable") {
-				err := showConnsHuman(data)
+				err := showConnsHuman(data, c.Bool("network-names"))
```
fullmetalcache, line 51: There are a lot of places where we are using the same hardcoded strings. We might want to consider declaring these somewhere as constants for use throughout the program.

commands/show-bl-ip.go
```diff
@@ -130,21 +120,29 @@ func printBLDestIPs(c *cli.Context) error {
...
 			return cli.NewExitError(err.Error(), -1)
 		}
 	}
 	return nil
 }
 
-func showBLIPs(ips []host.AnalysisView, connectedHosts, source bool, delim string) error {
-	headers := []string{"IP", "Connections", "Unique Connections", "Total Bytes"}
+func showBLIPs(ips []blacklist.IPResult, connectedHosts, showNetNames, source bool, delim string) error {
+	var headers []string
```
fullmetalcache, line 137: Consider assigning the values to the headers variable all in one shot rather than assigning some of the values and then appending on the rest in a second statement. For instance, look at line 96 of the showConns function in show-long.connections.go file. The former method has the advantage that most of the header values only need to be modified in a single place and might reduce some trouble-shooting headaches later, the latter method is a bit faster as the program doesn't have to allocate space for a new array and then copy the elements due to the append function.  I saw the mismatch in logic-approach in a few locations.

parser/fileparser.go
```diff
@@ -103,22 +103,23 @@ func scanTSVHeader(fileScanner *bufio.Scanner) (*fpt.BroHeader, error) {
 		//On the comment lines
 		if fileScanner.Text()[0] == '#' {
 			line := strings.Fields(fileScanner.Text())
-			if strings.Contains(line[0], "separator") {
-				//TODO: Figure out how to escape read in string
-				if line[1] == "\\x09" {
-					toReturn.Separator = "\x09"
+			if line[0][1:] == "separator" {
```
fullmetalcache, line 106: Maybe consider collapsing this whole tree into a switch statement.  Might clean it up a little bit but that might just be personal preference too.

parser/filter_test.go
```diff
@@ -19,9 +21,9 @@ func TestFilterConnPairWithInternalSubnets(t *testing.T) {
 		res:             nil,
 		indexingThreads: 1,
 		parseThreads:    1,
-		internal:        getParsedSubnets([]string{"10.0.0.0/8"}),
-		alwaysIncluded:  getParsedSubnets([]string{"10.0.0.1/32", "10.0.0.3/32", "1.1.1.1/32", "1.1.1.3/32"}),
-		neverIncluded:   getParsedSubnets([]string{"10.0.0.2/32", "10.0.0.3/32", "1.1.1.2/32", "1.1.1.3/32"}),
+		internal:        util.ParseSubnets([]string{"10.0.0.0/8"}),
+		alwaysIncluded:  util.ParseSubnets([]string{"10.0.0.1/32", "10.0.0.3/32", "1.1.1.1/32", "1.1.1.3/32"}),
```
fullmetalcache, line 25: I don't know what the usage of these are but just a note that there are subnets that appears in both alwaysIncluded and neverIncluded (10.0.0.3/32, 1.1.1.3/32)

parser/filter_test.go
```diff
@@ -79,8 +81,8 @@ func TestFilterConnPairWithoutInternalSubnets(t *testing.T) {
 		indexingThreads: 1,
 		parseThreads:    1,
 		// purposely omitting internal subnet definition
-		alwaysIncluded: getParsedSubnets([]string{"10.0.0.1/32", "10.0.0.3/32", "1.1.1.1/32", "1.1.1.3/32"}),
-		neverIncluded:  getParsedSubnets([]string{"10.0.0.2/32", "10.0.0.3/32", "1.1.1.2/32", "1.1.1.3/32"}),
+		alwaysIncluded: util.ParseSubnets([]string{"10.0.0.1/32", "10.0.0.3/32", "1.1.1.1/32", "1.1.1.3/32"}),
```
fullmetalcache, line 84: Same comment here about subnets in both alwaysIncluded and neverIncluded

pkg/beacon/analyzer.go
```diff
@@ -273,15 +274,14 @@ func (a *analyzer) hostIcertQuery(icert bool, src string, dst string) updateInfo
...
+		var resList []interface{}
 
-		var resList []hostRes
+		hostSelector := src.BSONKey()
+		hostSelector["dat.icdst"] = dst.BSONKey()
 
-		_ = ssn.DB(a.db.GetSelectedDB()).C(a.conf.T.Structure.HostTable).Find(bson.M{"ip": src, "dat.icdst": dst}).All(&resList)
+		_ = ssn.DB(a.db.GetSelectedDB()).C(a.conf.T.Structure.HostTable).Find(hostSelector).All(&resList)
 
-		if !(len(resList) > 0) {
```
fullmetalcache, line 284: Good call on changing this one...the previous way of stating "less than or equal to zero" was...strange...

parser/fsimporter.go
```diff
@@ -402,120 +415,108 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
...
 								}
 
 								// Increment the connection count for the src-dst pair
-								uconnMap[srcDst].ConnectionCount++
-								hostMap[src].ConnectionCount++
-								hostMap[dst].ConnectionCount++
-
-								if stringInSlice(dst, hostMap[src].ConnectedDstHosts) == false {
-									hostMap[src].ConnectedDstHosts = append(hostMap[src].ConnectedDstHosts, dst)
-								}
```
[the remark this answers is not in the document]
ethack: There's a `ConnectedHosts` field in `pkg/host/repository.go#AnalysisView` that seems related. That has a bson tag so should get written to Mongo, though it says it's only for reading and only in blacklisted. I don't see `ConnectedHosts` being set anywhere in the code either. I checked a live Mongo dataset and don't see any of the values from that struct in the `host` collection. I think that's the view you moved to the blacklist module anyway.

ethack (changes requested):
I think most of my comments can be tabled for a future cleanup. There are a few questions I had for sanity checks though.

parser/fsimporter.go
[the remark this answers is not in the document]
Zalgo2462: Just checked the `HostnameResults()` function in `pkg/blacklist/results.go`. The other `ConnectedHosts` is read from the `sources` field coming out of the aggregation. The `sources` field gets built up by grabbing the IPs for a domain and running a join on the uconn table to see which IPs connected to the domain's IPs.

pkg/host/repository.go
```diff
@@ -26,12 +31,11 @@ type IP struct {
 	UntrustedAppConnCount int64
 	MaxTS                 int64
 	MinTS                 int64
-	ConnectedSrcHosts     []string
-	ConnectedDstHosts     []string
 	IP4                   bool
 	IP4Bin                int64
 }
 
+//TODO: Remove
```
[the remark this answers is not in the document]
Zalgo2462: Removed :+1:

commands/show-long-connections.go
ethack: `goconst` should be checking these. I didn't realize it was already enabled. We may need to investigate why it's not reporting them and decide if we should tweak its logic.
https://github.com/activecm/rita/blob/master/.golangci.yml#L8
I'll make an issue for this.

parser/filter_test.go
ethack: These are intentional for tests that check which list takes precedent.

parser/filter_test.go
ethack: These are intentional for tests that check which list takes precedent.

pkg/beacon/analyzer.go
ethack: There were other places in the code I noticed this hadn't yet been changed. We should make an effort to switch them all eventually.

pkg/blacklist/analyzer.go
```diff
@@ -155,54 +154,88 @@ func hasBlacklistedSrcQuery(chunk int, ip string, entry uconnRes, newFlag bool)
...
+//getUniqueConnsforBLSource returns the IP addresses that a given blacklisted IP contacted along with the number
+//of connections and bytes sent
+func (a *analyzer) getUniqueConnsforBLSource(blSourceIP data.UniqueIP) []connectionPeer {
+	ssn := a.db.Session.Copy()
+	defer ssn.Close()
+
+	var blIPs []connectionPeer
 
+	blIPQuery := []bson.M{
+		bson.M{"$match": bson.M{
```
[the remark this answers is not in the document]
ethack: @Zalgo2462

pkg/hostname/mongodb.go
```diff
@@ -85,19 +85,15 @@ func (r *repo) Upsert(hostnameMap map[string]*Input) {
 	)
 
 	// loop over map entries
-	for entry, ipLists := range hostnameMap {
+	for _, entry := range hostnameMap {
```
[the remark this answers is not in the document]
ethack: @Zalgo2462

pkg/useragent/mongodb.go
```diff
@@ -87,16 +88,15 @@ func (r *repo) Upsert(userAgentMap map[string]*Input) {
 	)
 
 	// loop over map entries
-	for key, value := range userAgentMap {
+	for _, entry := range userAgentMap {
```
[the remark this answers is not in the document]
ethack: @Zalgo2462

pkg/hostname/analyzer.go
```diff
@@ -77,14 +77,15 @@ func (a *analyzer) start() {
 			// set up writer output
 			var output update
 
+			//TODO[AGENT]: Use UniqueIP/ NetworkID info in ips/ src_ips in hostnames collection
```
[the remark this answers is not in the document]
Zalgo2462: Removed comment

commands/show-bl-ip.go
Zalgo2462: Changed. :+1:

commands/show-bl-ip.go
Zalgo2462: Looks like only this file used `headers`. Changed to `headerFields.`

commands/show-bl-hostname.go
Zalgo2462: Changed to match showBLIPs for consistency. Since this is printing logic, the speed shouldn't matter too much given it has to aggregate over the database first. However, I like keeping the code patterned and predictable.

commands/show-long-connections.go
Zalgo2462: Agreed, we should probably spin this off as a separate tech-debt PR. Our other projects need this as well.

commands/show-bl-ip.go
Zalgo2462: Changed, although it turned into a 6-way if statement.

parser/fileparser.go
Zalgo2462: cleaned up

pkg/blacklist/analyzer.go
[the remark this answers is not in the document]
Zalgo2462: Unfortunately, no, because we need to match on "src"/ "src_network_uuid" rather than on "ip"/ "network_uuid". 

The UniqueIPPair translates down to these fields but also includes "dst"/"dst_network_uuid".

One could imagine adding `AsSrcBSONKey` and `AsDstBSONKey` methods, but I didn't think of that while writing the code initially.

pkg/blacklist/results.go
```diff
@@ -0,0 +1,268 @@
...
+		}},
+		bson.M{"$unwind": "$ips"},
+		// find out which IPs connected to each hostname via uconn
+		bson.M{"$lookup": bson.M{
+			"from": "uconn",
+			"let":  bson.M{"ip": "$ips.ip", "network_uuid": "$ips.network_uuid"},
+			"pipeline": []bson.M{{"$match": bson.M{"$expr": bson.M{
+				"$and": []bson.M{
+					{"$eq": []string{"$dst", "$$ip"}},
+					{"$eq": []string{"$dst_network_uuid", "$$network_uuid"}},
```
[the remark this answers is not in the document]
Zalgo2462: Twas the first time I had dealt with it too! haha

pkg/hostname/mongodb.go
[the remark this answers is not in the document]
Zalgo2462: Yep, just cleaning things up.

ethack (approved):

PR #592, opened 2020-10-24
Add fqdn command

ethack:


commands/fqdn.go
```diff
@@ -0,0 +1,164 @@
...
+	// setting up to query mongo stuff (aka don't worry about it)
+	ssn := res.DB.Session.Copy()
+	defer ssn.Close()
+
+	// declare your result variable
+	var fqdn []FqdnResult
+
+	fqdnQuery := []bson.M{
+		bson.M{"$match": bson.M{
+			"dat.ips": bson.RegEx{ip, ""},
```
Zalgo2462, line 131: This will need to change to `data.ips.ip` after https://github.com/activecm/rita/pull/591

PR #608, opened 2021-02-12
Remove connection filter for dns entries

### LINKED ISSUE #607: Potential C2s over DNS not added to hosts table
There are certain cases where a host's ip for a c2 channel running over dns won't be added to the hosts table due to internal to internal connections being filtered out. Removing this filter for only the dns data will increase visibility for potential c2 over dns.

NahomieG:
Closes #607
Removed the connection filter for only the dns entries so that those host ip's can get added to the hosts table.

[dns-test (1).tar.gz](https://github.com/activecm/rita/files/5974284/dns-test.1.tar.gz)

For this dataset that previously would not import the host creating the dns queries, rita now imports said host.
image


Zalgo2462 (approved):
LGTM

Issue #609, opened 2021-02-12
Change TXT query count to DNS query count


NahomieG:
Currently the TXT query count in the hosts table counts all txt queries made by the host. It would be useful to have a count of total DNS queries made by said host instead. The txt query count field should be renamed to reflect the total dns query count and also remove the check for txt queries before adding the host to the host map & incrementing the count.

PR #610, opened 2021-02-12
Change TXT query count to max DNS query count

NahomieG:
- Changed the txt_query_count field to max_dns_query_count
  - max_dns_query_count represents the dns query count for the domain that had the most dns queries made to it by its respective host
- Added a check to only add the host to the host map & increment dns query count if the domain isn't an empty string and doesn't end in in-addr.arpa, since it can create an inaccurate dns query count
- Removed the check for txt queries, making the src ip get added to the host map & incrementing the query count regardless of query type.

For [this dataset](https://github.com/activecm/rita/files/6025257/dns-test.tar.gz), 172.31.26.157 made 
- 2074 queries to *.honestimnotevil.com
- 3 queries to *.aitai.ne.jp

Since *.honestimnotevil.com had the most amount of queries made to it, 2074 makes it into the max_dns_query_count field.

Closes: #609

### DIFF
--- a/parser/fsimporter.go
+++ b/parser/fsimporter.go
@@ -569,13 +569,15 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
 								}
 							}
 
-							// increment txt query count for host in uconn
-							if queryTypeName == "TXT" {
-								// We don't filter out the src ips like we do with the conn
-								// section since a c2 channel running over dns could have an
-								// internal ip to internal ip connection and not having that ip
-								// in the host table is limiting
+							// We don't filter out the src ips like we do with the conn
+							// section since a c2 channel running over dns could have an
+							// internal ip to internal ip connection and not having that ip
+							// in the host table is limiting
 
+							// in some of these strings, the empty space will get counted as a domain,
+							// don't add host or increment dns query count if queried domain
+							// is blank or ends in 'in-addr.arpa'
+							if (domain != "") && (!strings.HasSuffix(domain, "in-addr.arpa")) {
 								// Check if host map value is set, because this record could
 								// come before a relevant conns record
 								if _, ok := hostMap[srcKey]; !ok {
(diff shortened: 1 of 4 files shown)


parser/fsimporter.go
```diff
@@ -568,16 +569,17 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
...
-
-								// Run conn pair through filter to filter out certain connections
-								ignore := fs.filterConnPair(srcIP, dstIP)
-								if !ignore {
+							// get destination for dns record
+							dst := parseDNS.Destination
+							dstIP := net.ParseIP(dst)
 
+							// Run conn pair through filter to filter out certain connections
+							ignore := fs.filterConnPair(srcIP, dstIP)
```
Zalgo2462, line 577: Do we want to allow DNS for internal-> internal? It looks like we are keeping the filter for dns logs currently.

NahomieG: The change for removing the filter is in #608.

Zalgo2462: :+1: Pulled in #608.

Zalgo2462 (approved):
Looks good to me. In the future we may have to put a limit on the number of exploded dns entries we write into the database, throwing away the low frequency results. 

I did a little cleanup before pulling this in:
- created shouldInsertNewHostRecord function from code which sets newRecordFlag
- moved exploded dns logic into two functions: buildExplodedDNSArray and writeExplodedDNSEntries
- changed the push in the standardQuery to separate the dns document from the connection counts document
- renamed max_dns_query_count to max_dns so the field name max_dns_query_count.count doesn't repeat itself
Logan L
Move newRecordFlag code to its own function in host/analyzer.go
f9432e6
Logan L
simplifiy shouldInserNewHostRecord in host/analyzer.go
d569472
Logan L
Break host exploded dns analysis up into functions: buildExplodedDNSA…
6cd8497
Logan L
Separate max_dns_query_count document from main connection counter do…
49943d9
Logan L
Rename dat.max_dns_query_count to dat.max_dns
d8b5b46
Logan L
Remove resources reference in host analyzer in favor of logger reference

PR #613, opened 2021-03-03
Count the number of unique subdomains a host looked up for a given (super)domain rather than total number of queries

### LINKED ISSUE #612: Change Host (Exploded) DNS Query Counts to Subdomain Counts
Currently, the hosts analysis calculates a per-host exploded dns analysis in a similar way to the dataset-wide exploded dns analysis. Right now, the host based exploded dns analysis is calculating how many times a host queried for a subdomain of a particular domain ("Times Looked Up" in `show-exploded-dns`). However, the number of subdomains rather than the number of queries is usually a better indicator of domain generation algorithms.  We should change the host based exploded dns analysis to count the number of subdomains queried by each host ("Unique Subdomains" in `show-exploded-dns`).

Zalgo2462:
Fixes #612.

In the `show-exploded-dns` output, we display two values `Times Looked Up` and `Unique Subdomains`. 

Currently the host based exploded dns analysis i.e. `host.max_dns.count` is written to compute the number of times a host looked up the subdomains of a given (super)domain (`host.max_dns.query`). This is akin to the `Times Looked Up` field in the dataset wide exploded dns analysis.

This PR changes `host.max_dns.count` to compute the number of unique subdomains a host looked up for a given (super)domain. This makes the values match the `Unique Subdomains` values reported by `show-exploded-dns`.

### DIFF
--- a/pkg/host/analyzer.go
+++ b/pkg/host/analyzer.go
@@ -134,7 +134,7 @@ func buildExplodedDNSArray(dnsQueryCounts map[string]int64) []explodedDNS {
 	// make a new map to store the exploded dns query->count data
 	var explodedDNSMap map[string]int64
 	explodedDNSMap = make(map[string]int64)
-	for domain, count := range dnsQueryCounts {
+	for domain := range dnsQueryCounts {
 		// split name on periods
 		split := strings.Split(domain, ".")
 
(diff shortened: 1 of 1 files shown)


NahomieG (approved):
lgtm

PR #615, opened 2021-03-04
Fix human readable labels in show-beacons-fqdn

### LINKED ISSUE #614: Incorrect Headers in show-beacons-fqdn
Currently, there is an extra header field "Resolved IPs" in the human readable output for the fqdn beacons. It should be removed.

Zalgo2462:
Fixes #614


fullmetalcache (approved):
ltgm

Issue #577, opened 2020-08-04
Unexpected protocol: Extend trustedAppReferenceList to include more protocols


william-stearns:
Currently we list just 80/tcp/http and 443/tcp/ssl.  Consider adding other protocols that are likely to be allowed outbound.

william-stearns:
Current list is found in https://github.com/activecm/rita/blob/master/parser/fsimporter.go :

var trustedAppReferenceList = [...]trustedAppTiplet{
        {"tcp", 80, "http"},
        {"tcp", 443, "ssl"},
}

Initial list of potential additions:
22/tcp/ssh
25/tcp/smtp
53/udp/dns
53/tcp/dns
587/tcp/smtp,ssl
993/tcp/ssl

cbrenton-acm:
Let's make sure we test TCP/53. I think I've seen cases where legit DNS was
bing used but Zeek seemed unable to understand the app layer.

Thanks!
C

william-stearns:
I extracted the 53/tcp lines from ~9 months of bro logs and pulled out the port, protocol, service and state fields, sorting and counting them:
     35 53	tcp	-	OTH
  10917 53	tcp	-	REJ
   5546 53	tcp	-	RSTO
   1576 53	tcp	-	RSTOS0
     54 53	tcp	-	RSTR
    116 53	tcp	-	RSTRH
  12330 53	tcp	-	S0
     85 53	tcp	-	S1
      2 53	tcp	-	S2
      1 53	tcp	-	S3
    207 53	tcp	-	SF
   3309 53	tcp	-	SH
    145 53	tcp	-	SHR
      4 53	tcp	dce_rpc,dns	SF
     57 53	tcp	dns	OTH
    462 53	tcp	dns	RSTO
   4411 53	tcp	dns	RSTOS0
   6778 53	tcp	dns	RSTR
     20 53	tcp	dns	S1
   3091 53	tcp	dns	S2
      8 53	tcp	dns	S3
4430108 53	tcp	dns	SF
     33 53	tcp	dns	SH
      6 53	tcp	dns	SHR
      7 53	tcp	dns,dce_rpc	SF

william-stearns:
There's some additional discussion about this at https://docs.google.com/document/d/1lJApa-FwU1sZD88Y_Lhk4Kpq8iUQdeuBrhEZ_RYfrWU/edit# .

william-stearns:
Please add Hannah C to this.

PR #618, opened 2021-03-12
WIP Unexpected protocol: Extend trustedAppReferenceList to include more protocols #577

Hannnah1:
This expands the list of well known protocol/port/services. Given the potential impact on performance I only included a few that seem to appear most frequently, but it could easily be expanded further. All tests pass, however the intended result has not been verified as I do not know a way to view the data without the AC-Hunter interface.

### DIFF
--- a/parser/fsimporter.go
+++ b/parser/fsimporter.go
@@ -68,11 +68,18 @@ func NewFSImporter(res *resources.Resources,
 		neverIncluded:   util.ParseSubnets(res.Config.S.Filtering.NeverInclude),
 	}
 }
-
+// These protocols are innocent until proven guilty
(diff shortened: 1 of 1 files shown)


william-stearns (changes requested):
Well done!  This looks good.  A few minor changes:
{"tcp", 80, []string{"http"}},

{"tcp", 443, []string{"ssl", "http,ssl", "ssl,http"}},

{"tcp", 25, []string{"smtp", "ssl", "smtp,ssl", "ssl,smtp"}},

{"tcp", 143, []string{"imap", "ssl", "imap,ssl", "ssl,imap"}},

PR #622, opened 2021-03-16
Clean up TODO and NOTE markers. Remove old ip index in host collection.

Zalgo2462:
Over time, we've introduced TODO and NOTE comments to point out particular bits of code which we have concerns over. However, some of these have been addressed. This PR removes the markers which are no longer needed.

Additionally, one of these TODO notes questioned whether the `ip` index in the `host` collection was ever used. I ran the data through RITA's import and show-xxx commands as well as other visualization software to ensure the index is not used. 

#### Data visualization index use for host collection
index use headers
visualization index use

#### Import index use for host collection
import index use


database/meta.go
```diff
@@ -113,7 +113,7 @@ func (m *MetaDB) SetRollingSettings(db string, chunk int, numchunks int) error {
 			update["cid_list."+strconv.Itoa(i)+".set"] = false
 		}
 	} else if result.TotalChunks > numchunks {
-		// Note: this shouldn't ever get hit because the commands/import.go:setRolling function should fail on this check
+		// This shouldn't ever get hit because the commands/import.go:setRolling function should fail on this check
```
Zalgo2462, line 116: This note works as a comment and the TODO directly below it directs attention here.

parser/fileparser.go
```diff
@@ -182,7 +182,7 @@ func mapBroHeaderToParserType(header *fpt.BroHeader, broDataFactory func() pt.Br
 	for index, name := range header.Names {
 		lu, ok := fieldTypes[name]
 		if !ok {
-			//NOTE: an unmatched field which exists in the log but not the struct
+			//An unmatched field which exists in the log but not the struct
```
Zalgo2462, line 185: This is a well known behavior of RITA at this point. Perhaps it should be better documented elsewhere.

pkg/beaconfqdn/repository.go
```diff
@@ -61,7 +61,6 @@ type (
 		ResolvedIPs    []data.UniqueIP `bson:"resolved_ips"`
 	}
 
-	//TODO: Implement strobe for beacons FQDN
```
Zalgo2462, line 64: This has been implemented.

pkg/host/mongodb.go
```diff
@@ -31,7 +31,6 @@ func (r *repo) CreateIndexes() error {
 	// create hosts collection
 	// Desired indexes
 	indexes := []mgo.Index{
-		{Key: []string{"ip"}}, //TODO[AGENT]: Determine if this index is needed
```
Zalgo2462, line 34: Investigated and removed.

pkg/uconn/analyzer.go
```diff
@@ -145,8 +145,6 @@ func (a *analyzer) hostMaxDurQuery(maxDur float64, localIP data.UniqueIP, extern
 	// create query
 	query := bson.M{}
 
-	//TODO: Why do we update old chunks with new max durations??? (https://github.com/activecm/rita/pull/512/files) -LL
```
Zalgo2462, line 148: While this code could be refactored, generally, we create a new entry for each chunk and update all of the entries if an entry already exists for the current chunk.

ethack (approved):

PR #625, opened 2021-03-23
Added TotalBytes to show-beacons and html-report

Hannnah1:
I made it so the total bytes per connection shows up in the show-beacons command and html-report.  builds and passes tests.

### DIFF
--- a/pkg/beacon/repository.go
+++ b/pkg/beacon/repository.go
@@ -50,7 +50,7 @@ type Result struct {
 	data.UniqueIPPair `bson:",inline"`
 	Connections       int64   `bson:"connection_count"`
 	AvgBytes          float64 `bson:"avg_bytes"`
-	TB								int64    `bson:"total_bytes"`
+	TotBytes          int64    `bson:"total_bytes"`
 	Ts                TSData  `bson:"ts"`
 	Ds                DSData  `bson:"ds"`
 	Score             float64 `bson:"score"`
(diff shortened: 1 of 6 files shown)


pkg/beacon/repository.go
```diff
@@ -50,6 +50,7 @@ type Result struct {
 	data.UniqueIPPair `bson:",inline"`
 	Connections       int64   `bson:"connection_count"`
 	AvgBytes          float64 `bson:"avg_bytes"`
+	TB								int64    `bson:"total_bytes"`
```
fullmetalcache, line 53: Looks like indention is a bit off here. Guessing a tabs/spaces issue?

fullmetalcache: Would you be opposed to expanding TB out into TotalBytes? Just to cut down on the number of acronyms. Not a big deal, just a suggestion.

fullmetalcache (commented):
Just some very minor comments. Looks good!

pkg/beacon/repository.go
Hannnah1: TotalBytes is taken. I'm not too familiar with go, but I assume I cannot reuse the variable name even if it's within a struct. I can definitely expand it though.

fullmetalcache: Sorry, can you please clarify a bit? As long as there isn't another struct field with the same name in that struct then you should be able to use that name.

Hannnah1: No you're right I just got confused. Everything mentioned above has been changed. :)

fullmetalcache (approved):
Looks great!!

PR #631, opened 2021-03-29
make `--config` a global option on `rita` command

### LINKED ISSUE #595: Move --config to a global option
So instead of having to do `rita import --config etc/rita.yaml ...` we would do `rita --config etc/rita.yaml import ...`. This makes it easier to modify most commands since `--config` is less likely to change than the rita subcommand. It also makes it possible to create aliases like:
- `alias rita1=rita --config /etc/rita/config1.yaml`
- `alias rita2=rita --config /etc/rita/config2.yaml`

Where currently this would be impossible with an alias and would need a custom shell function instead.

vivek-26:
This PR does the following:
  * makes `--config` a global option on the `rita` command.
  * removes `--config` flag from subcommands (`import`, `html-report` etc.)
  * uses [`GlobalString`](https://github.com/urfave/cli/blob/v1.22.0/flag_generated.go#L709) to fetch config value from global state. For reference, see this [issue](https://github.com/urfave/cli/issues/716).

Closes #595

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -13,7 +13,7 @@ var (
 
 	// below are some prebuilt flags that get used often in various commands
 
-	// configFlag allows users to specify an alternate config file to use
+	// ConfigFlag allows users to specify an alternate config file to use
 	ConfigFlag = cli.StringFlag{
 		Name:  "config, c",
 		Usage: "Use a given `CONFIG_FILE` when running this command",
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -13,13 +13,6 @@ var (
 
 	// below are some prebuilt flags that get used often in various commands
 
-	// configFlag allows users to specify an alternate config file to use
-	configFlag = cli.StringFlag{
-		Name:  "config, c",
-		Usage: "Use a given `CONFIG_FILE` when running this command",
-		Value: "",
-	}
-
 	// forceFlag allows users to bypass prompts
 	forceFlag = cli.BoolFlag{
 		Name:  "force, f",
@@ -130,7 +123,7 @@ func bootstrapCommands(commands ...cli.Command) {
 	for _, command := range commands {
 		command.Before = func(c *cli.Context) error {
 			//Get access to the logger
-			configFile := c.String("config")
+			configFile := c.GlobalString("config")
 			res := resources.InitResources(configFile)
 			//Display args in logs
 			fields := log.Fields{
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -13,6 +13,13 @@ var (
 
 	// below are some prebuilt flags that get used often in various commands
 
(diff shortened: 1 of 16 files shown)


ethack:
This is great, thank you!

Do you think you could also migrate the `test-config` command to use the global option?
```
$ ./rita test-config --help                            
NAME:
   rita test-config - Check the configuration file for validity

USAGE:
   rita test-config [command options] [arguments...]

OPTIONS:
   --config value, -c value  specify a config file to be used
```

With these changes the config flag isn't recognized after the command. This invocation is what was needed prior to these changes and now produces an error. This could be a breaking change for some users. 
```
./rita show-beacons --config etc/rita.yaml test
Incorrect Usage: flag provided but not defined: -config
```

Do you know if there's a urlfave/cli option to have the `--config` flag recognized wherever it is placed? 

If there's not a simpler solution than putting the config flag back into all the subcommands and testing each case for both global and local config flags then I would rather we just announce the breaking change and possibly hold this until a major version bump. But if there's a clean fix that lets the flag be recognized anywhere then I'd be up for that to maintain backwards compatibility.

vivek-26:
> Do you think you could also migrate the `test-config` command to use the global option?

Sure, will do that.


> Do you know if there's a urlfave/cli option to have the `--config` flag recognized wherever it is placed?

As per this [issue](https://github.com/urfave/cli/issues/585), there are 2 ways to achieve this. We could do something like [this](https://github.com/urfave/cli/issues/585#issuecomment-272987605) or we can upgrade `urfave/cli` to `v2`. I wrote a very simple example using `urfave/cli`-`v2` and global flag is recognised by all subcommands. However, upgrading `urfave/cli` to `v2` might cause breaking changes. ([example](https://github.com/urfave/cli/blob/master/docs/migrate-v1-to-v2.md#flags-before-args))

vivek-26:
@ethack I have updated PR to make `--config` flag available to both root command and all subcommands.

Current CLI behaviour is as follows - 
<hr>

Command: `$:  ./rita --config /etc/rita.yaml test-config`
Behaviour: `test-config` subcommand is aware of the `--config` flag value passed to the root command.
<hr>

Command: `$:  ./rita test-config --config /etc/rita.yaml`
Behaviour: `test-config` subcommand uses `--config` flag passed to it.
<hr>

Commands: 
`alias rita1=rita --config /etc/rita/config.yaml`
`$:  rita1 test-config --config /etc/new-config.yaml`

Behaviour: `test-config` subcommand overwrites `--config` value set globally  - `/etc/rita/config.yaml` and uses `/etc/new-config.yaml`.
<hr>

Closes #595

vivek-26:
@ethack PTAL

ethack (approved):
LGTM. Thanks for the contribution!

Issue #635, opened 2021-04-05
Beacon FQDN Analysis Slow


fullmetalcache:
Beacon FQDN Analysis seems to run slowly on larger data sets. This issue is to investigate what, if any, improvements can be made.

PR #638, opened 2021-04-26
635 beaconfqdn slow

fullmetalcache:
Credits to @lisaSW as well

We modified how we are performing queries when performing beacon FQDN analysis. 

Previously, we would:

1) Get a list of resolved IPs for an FQDN
2) Get a list of Src IPs that connected to any of the resolved IPs
3) For each Src IP, we queried the UCONN table for entries where the Src IP was a Src and any of the resolved IPs were a Dst
4) We then performed the beacon FQDN analysis for the set of timestamps

Now, we do the following:

1) Get a list of resolved IPs for an FQDN
2) Query the UCONN table for all entries where any of the resolved IPs appear as a Dst and then group the results based up Src IP/UUID
3) We then iterate over the array of results such that each array entry contains, among other information, time stamps between a Src IP/UUID and any of the resolved IPs for an FQDN. Beacon FQDN analysis is performed on each of these sets of time stamps

This subtle difference resulted in over an 8x speed up from the initial implementation.


fullmetalcache:
Closes #635

william-stearns:
In pkg/beaconfqdn/dissector.go, there appears to be a hardcoded IP: 
"$or": [{"dst":"172.217.4.226"}]
Any chance this is leftover from testing?

fullmetalcache:
It's part of the example Mongo query for that piece of code. The IP is from the dnscat2-ja3-strobe-agent set. It's not actually used in the code; just there so that the query is valid without modification.

pkg/beaconfqdn/dissector.go
```diff
@@ -95,7 +191,7 @@ func (d *dissector) start() {
 					"icerts": bson.M{"$anyElementTrue": []interface{}{"$dat.icerts"}},
 				}},
 				{"$group": bson.M{
-					"_id":    "$src",
+					"_id":    bson.M{"src": "$src", "uuid": "$src_network_uuid", "network": "$src_network_name"},
```
Zalgo2462, line 194: We cannot group on network name as multiple network names may be associated with the same network uuid.
This is a downstream result of how winlogbeat identifies Sysmon agents.

Instead, please use `"network_name": bson.M{"$first": "$src_network_name"}`. Unfortunately, this will need to be added to most of the clauses below.

Zalgo2462 (changes requested):
Approach looks good. We need to tweak one thing regarding the network_names. 

Great work y'all!

Zalgo2462:
Merged master back into this branch. Solved merge conflicts by removing the icerts calculations in fqdn beacons since it was removed in master.

Also, the fqdn analysis is like wicked fast yo :raised_hands:

pkg/beaconfqdn/dissector.go
NahomieG: In the past, queries used `$last` to group on network uuids so that we'd be referencing the last known netbios name in case it ever changed. Is there a reason why we should use `$first` here instead?

Zalgo2462: Nope. Great catch. I'm just a big dummy. :clown_face: Please use `$last`.

fullmetalcache:
Believe I have the comments addressed and this is ready for the final approval.

Thanks!

pkg/beaconfqdn/dissector.go
```diff
@@ -46,14 +47,96 @@ func (d *dissector) close() {
...
+                "input":        "$dat.bytes",
+                "initialValue": [],
+                "in":           {"$concatArrays": ["$$value", "$$this"]},
+            },
+        },
+        "count":  {"$sum": "$dat.count"},
+        "tbytes": {"$sum": "$dat.tbytes"},
+    }},
+    {"$group": {
+        "_id":              {"src": "$src", "uuid": "$src_network_uuid", "network": "$src_network_name"},
```
Zalgo2462, line 77: Looks like this comment needs to be changed to match the recent changes.

Zalgo2462: I believe this PR is ready to pull in otherwise.

Zalgo2462 (changes requested):

pkg/beaconfqdn/dissector.go
fullmetalcache: Think i've got it fixed, thanks!

Zalgo2462 (approved):
LGTM. Tested and it seemed to work well.

bekirk:
I have a question on this.  I just tried this on a large dataset, and it is really slow, might finish in (2 hours of Bro data took over 30 minutes) 6-8 hours, i'm not sure yet.  

Are there any configuration changes I can make?  I tried upping the Threshold to 80 and even 200, and got a little improvement.  

Is it possible to run all the analysis beside FDQN Beacon, then come back and run just the FQDN Beacon so that I might be able to start looking at the other data while the FDQN Beacon is still processing?

Thank you,
Brian Kirk

Zalgo2462:
Hello, thank you for letting us know about the issue. The code patch here is set to be released as part of RITA v4.3.0 https://github.com/activecm/rita/releases/download/v4.3.0/install.sh. We are currently running quality control testing on the new release, but it is available as a pre-release. We expect that v4.3.0 will have a formal release sometime in the coming week. 

Would you be interested in giving version 4.3.0 a try and letting us know if the amount of time required to process the FQDN Beacons goes down?

Note: RITA v4.3.0 requires upgrading MongoDB to version 4.2. Unfortunately, this means that previous versions of RITA will no longer continue to work as the maximum version of MongoDB supported by previous versions is 3.6.

bekirk:
Sure I can test, I might need figure out a backout plan if I need to go back to mongodb 3.6

I have all my current Rita data indexed into splunk so I typically don't need to go back to previous data.

bekirk:
I installed 4.3.0 and tested.  I had some problems with the zeek and mongodb installs but since I already installed mongodb and I don't run bro nor zeek on this system, I was able to run the install with no zeek and no mongo and it worked.  

The beacons-FQDN is really fast now! I will see what the runtime is in the morning for a full day's worth of logs.  A difference between the beacons-fqdn and beacons, it doesn't have the total bytes like the beacons does now.  Also I notice that I have almost 10 times as many rows in the beacons-fqdn as I do in the beacons, is there some way to put all the rows for the same beacon activity into one row with a multivalue field like the "Port:Protocol:Service" in long-conns? Maybe you can even add the dest-ip, unless I am missing something with the what the beacons-fqdn is doing, I think it just help see what the possible dns request of the traffic was using a reverse lookup, like if it was login.microsoftonline.com you might be able to filter out the result. However for one connection to google I see all these FQDN's all from one source IP with the same values in every other field.:
clients1.google.com
sb-ssl.l.google.com
music.youtube.com
clients5.google.com
play.google.com
photos.google.com
sb-ssl.google.com
clients2.google.com
www.youtube.com
admin.google.com
maps.google.com
ytimg.l.google.com
clients6.google.com
clients.l.google.com
img.youtube.com
clients4.google.com
clients3.google.com
youtube-ui.l.google.com
calendar.google.com
maps-api-ssl.google.com
android.clients.google.com 

Maybe give an option to include the Dest-IP and another option if you want to condense the data with a multi-value field.

Thank you,
Brian

bekirk:
So the Import command ran in a reasonable time, however the rita-show-beacons-fqdn too almost 6 hours to run and is 5.1G in size:

```
/opt/rita/DataPreSplunk/2021-06-29$ ls -lh
total 5.2G
-rw-rw-r-- 1 rita rita 5.1G Jun 30 12:13 beacons-fqdn.log
-rw-rw-r-- 1 rita rita  19M Jun 30 12:13 beacons.log
```
Does this seam right?  

Thank you,
Brian

PR #644, opened 2021-05-27
Fixed missing </td> in report-beacons.go and report-beaconsfqdn.go

soveraev:
Hi 

I've fixed a minor error in the html-export command.
This made the data very hard to parse with other tools, for instance pandas.
I've fixed it by adding the required `</td>` in the code after `.Ds.ModeCount`


Zalgo2462 (approved):
Verified the fix by checking the resulting html source. Thank you for the fix!

PR #643, opened 2021-05-26
641 open conn handling

fullmetalcache:
Adds functionality for dealing with open-connection log entries that arise from this Zeek plugin: https://github.com/corelight/zeek-long-connections.

Closes #641

PR #647, opened 2021-06-12
 641 open conn handling alternative from @fullmetalcache

### LINKED ISSUE #641: Handle Processing Long Connections that Haven't Closed
Zeek does not write out connection duration until the connection has closed. This issue can be mitigated by using a custom Zeek script, such as the one from here: https://github.com/corelight/zeek-long-connections

This issue is to implement functionality in RITA to process these interim logs.

Care must be taken that data is not duplicated between the interim logs and when a connection closes. Particularly, TotalDuration and TotalBytes are two fields that need extra attention.

Additionally, a feature should be implemented to keep track of if a connection is open or closed.

Zalgo2462:
Opening draft to leave comments/ easily review changes. See #643. Closes #641.


commands/show-open-connections.go
```diff
@@ -0,0 +1,176 @@
...
+		var row []string
+
+		if showNetNames {
+			row = []string{
+				result.SrcNetworkName,
+				result.DstNetworkName,
+				result.SrcIP,
+				result.DstIP,
+				result.Tuple,
+				f(result.Duration),
```
Zalgo2462, line 157: Pretty sure this should be 
`openDuration(time.Duration(int(result.Duration * float64(time.Second))))`,

commands/show-open-connections.go
```diff
@@ -0,0 +1,176 @@
...
+				strconv.Itoa(result.Bytes),
+				result.UID,
```
Zalgo2462, line 159: We generally don't display the zeek uid's anywhere else in RITA. We might want to remove the column for consistency, but I'm okay with keeping it. Might want to title the column header Zeek UID.

pkg/uconn/repository.go
```diff
@@ -25,19 +25,25 @@ type update struct {
...
+	TotalDuration       float64
+	OpenTSList          []int64
+	TsList              []int64
+	OrigBytesList       []int64
+	OpenOrigBytes       int64
+	Tuples              []string
 	// InvalidCerts    []string
 	InvalidCertFlag bool
 	UPPSFlag        bool
+	ConnStateList   map[string]*ConnState
```
Zalgo2462, line 46: Could we rename this ConnStateMap?

parser/fsimporter.go
```diff
@@ -732,6 +756,188 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
 
 							mutex.Unlock()
 
+						/// *************************************************************///
+						///                           OPEN CONNS                         ///
+						/// *************************************************************///
+						case fs.res.Config.T.Structure.OpenConnTable:
+							fmt.Println("Here and stuff")
+							fmt.Println("Here and stuff")
```
Zalgo2462, line 764: Debug comments :)

parser/fsimporter.go
```diff
@@ -732,6 +756,188 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
...
+
+								// If an entry for this open connection is present, first check
+								// to make sure it hasn't been marked as closed. If it was marked
+								// as closed, that supersedes any open connections information.
+								// Otherwise, if it's open then check if this more up-to-date data
+								// (i.e., longer duration)
+								if _, ok := uconnMap[srcDstKey].ConnStateList[uid]; ok {
+									if uconnMap[srcDstKey].ConnStateList[uid].Open {
+
+										if duration > uconnMap[srcDstKey].ConnStateList[uid].Duration {
```
Zalgo2462, line 869: Mind collapsing these two conditions?

parser/fsimporter.go
```diff
@@ -732,6 +756,188 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
...
+								}
+
+								if duration > hostMap[srcKey].MaxDuration {
+									hostMap[srcKey].MaxDuration = duration
+								}
+								if duration > hostMap[dstKey].MaxDuration {
+									hostMap[dstKey].MaxDuration = duration
+								}
+
+								mutex.Unlock()
```
Zalgo2462, line 937: Can we add a comment about not incrementing the connection counters until a connection is closed here?

parser/parsetypes/open_conn.go
```diff
@@ -0,0 +1,70 @@
+package parsetypes
+
+import (
+	"github.com/activecm/rita/config"
+	"github.com/globalsign/mgo/bson"
+)
+
+// OpenConn provides a data structure for bro's open connection data
```
Zalgo2462, line 8: Mind changing this to zeek for new code?

parser/parsetypes/parsetypes.go
```diff
@@ -33,6 +33,10 @@ func NewBroDataFactory(fileType string) func() BroData {
 		return func() BroData {
 			return &HTTP{}
 		}
+	} else if strings.HasPrefix(fileType, "open") {
```
Zalgo2462, line 36: Might go with "open_conn" in case someone has a weird plugin that also uses "openXXXX" for a log name.
The other logs are default included in Zeek, so there is less opportunity for collisions.

pkg/blacklist/analyzer.go
```diff
@@ -203,6 +204,15 @@ func (a *analyzer) getUniqueConnsforBLDestination(blDestinationIP data.UniqueIP)
 			},
 			"bl_conn_count":  bson.M{"$sum": "$dat.count"},
 			"bl_total_bytes": bson.M{"$sum": "$dat.tbytes"},
+			// I don't think that either of these fields will ever be more than one value...
+			// ...but just in case
```
Zalgo2462, line 208: Could you expand on this a bit? I don't quite understand.

Zalgo2462 (commented):
I generally prefer the code in this PR to https://github.com/activecm/rita/pull/643/files. In particular, this code is easier to refactor if we split out the different parsing routines by entry type. 

I've commented up some minor changes, mostly naming and comments. I think there is one code change for formatting output. 

Still working testing the PRs at the moment.

commands/show-open-connections.go
Zalgo2462: image

Zalgo2462 (commented):
Tested with TCP long connections/ open connections and both PRs work well.

Testing ICMP C2 tonight.

commands/show-open-connections.go
fullmetalcache: Yep, good catch!

commands/show-open-connections.go
fullmetalcache: I thought it might be helpful in this case to allow for tracking the particular connection that is open. Not sure it will be useful in practice. Modifying the column header for now

parser/fsimporter.go
fullmetalcache: Good call. In the initial version there was more happening in there but with this alternative version it's a bit unnecessary to have them be separate statements.

parser/parsetypes/parsetypes.go
fullmetalcache: yep, sounds good!

fullmetalcache:
I believe that I have all of the comments above addressed. Thanks!

Zalgo2462:
#### Non-TCP Problems
 I tested this with icmpsh today and found that the `zeek-open-connections` script will not write an entry to `open_conn` if the transport is not TCP.
image
image

---

#### Alternatives
I suggest using the following events depending on the version of Zeek installed.
For versions >= 3.1.0:
image
Otherwise:
image


Testing with `new_connection` on 3.0.12, I was able to generate `open_conn` logs with icmp records while running icmpsh.

I'll put up a PR on the zeek-open-connections PR to get this fixed up.

Zalgo2462:
As soon as we fix tracking non TCP connections (https://github.com/activecm/zeek-open-connections/pull/4) and add a blurb to the README, I think this is good to go.

Zalgo2462:
Another reason to prefer this PR over the other one:

I was able to put a TCP session in the following state and keep it open.
> S3: Connection established and close attempt by responder seen (but no reply from originator).

Only this PR was able to track it as a still open connection.

Zalgo2462:
Good to merge, though I can't approve it since I technically opened it even though its @fullmetalcache's  work.

lisaSW (approved):
approved by @zalgo2462 and @fullmetalcache

PR #654, opened 2021-07-07
Speed up RITA parser

Zalgo2462:
#### Goal
This PR speeds up the RITA parser for both JSON and TSV.  In one benchmark parsing large amounts JSON files, this PR executes in a quarter of time needed by the master branch.

#### Note to the reviewer
I would recommend reviewing it by looking at the individual commits, as there have been a large number of changes. If this is overwhelming, we can split up the PR into sections by splitting off the individual commits into their own PRs.

#### Changes
- In https://github.com/activecm/rita/pull/654/commits/bfed5a98cc45868d62d88ddf8a28c505bde7e49f, I refactored the dependencies for the `FSImporter`. This was important because any profiling done on the `FSImporter` simply reported that `parseFiles` was taking a long time. `parseFiles` contained the parsing logic for each type of log file. In order to break up this method, I needed to simplify the method's dependencies. 

- In https://github.com/activecm/rita/pull/654/commits/05627e59d3c2f7e27fb48cffa74534c8eab839d3, the logic for parsing each type of file has been moved to its own file. Additionally, breaking up this logic allowed me to break up the mutex that was used to make writing out the parsing results threadsafe. In this revision, each map within the parse results has its own mutex, allowing for greater parallelism. 

- https://github.com/activecm/rita/pull/654/commits/741a9cd48da80d6e245fbfa1f4cf924494c17af4 adds in a printed message which tells the user how long it took to read in a batch of logs. Additionally, this commit embeds profiling logic as comments into the code to make it easier to enable cpu and ram profiling. We can revert this commit if need be.

- In https://github.com/activecm/rita/pull/654/commits/ae7cf37f66a09ad034e450a08dcc9b06b59829f1, I replaced calls to `strings.Split` with calls to `strings.Index` and slice operations in order to reduce string allocations in the TSV parsing code. Since this section of code was called for each line we parsed, the targeted allocations created a good deal of garbage that the GC had to constantly take care of. 

- In https://github.com/activecm/rita/pull/654/commits/ddef24bbbefcb2e9abc3973b25dffb12d0bd7328, `strconv.ParseInt` was replaced with `strconv.Atoi` because the Go library has added a performance hack for small numbers in one method but not the other. The `ParseInt` method was identified as a major time sink in the TSV parser via `go tool pprof`. 

- In https://github.com/activecm/rita/pull/654/commits/a0cfecda7cbba4763f67577ce675c32d949e66af, the map which tells the TSV parser which struct field each sequential token maps to was replaced with an array. The profiler indicated that the TSV parser spent the majority of its time hashing field names. Since the names of the fields were always accessed sequentially, it was possible to map from the sequential index directly to the target struct fields.

- In https://github.com/activecm/rita/pull/654/commits/0e911ee14f3ca5be85ad1ce10ab434f8f0aaf058 and https://github.com/activecm/rita/pull/654/commits/dbf8ce9e6e9d5e2cf921c819c04720c044666de5, the Golang `gzip` library was replaced with calls to the system `pigz` or `gzip` commands. The profiler indicated that much of the time in the parse loop was spent decompressing files. After a google search, I found that Docker has run into similar issues: https://github.com/moby/moby/issues/10181. I then implemented a fix similar to theirs. This fix saved a great deal of time for both the TSV and JSON parsers. 

- In https://github.com/activecm/rita/pull/654/commits/0616f3ed61ac81ae11de15d550321bc13e568d44, I replaced the standard  JSON parser with a faster library. The profiler indicated that much of the time spent by the JSON parser was deep in the Golang standard library. After a few searches, I found that [Prometheus](https://github.com/prometheus/prometheus/blob/f5655c47e8d88220fab4c0a034ab480d4bbd537b/go.mod#L34), [Docker](https://github.com/moby/moby/blob/b316cc059a81d9abe7d7edb4e3620b5aa01ef5e4/vendor.conf#L71), and [Grafana](https://github.com/grafana/grafana/blob/2e0dc835cfa0d651bf7fbc6448f1ab8640559be3/go.mod#L64) as well as many other well-known Go projects used the package `json-iter` to speed up their JSON marshalling. It was a drop in replacement and yielded a massive speedup. 

- In  https://github.com/activecm/rita/pull/654/commits/cd8dd38af38d30b06b0141d08b3370435ef2ba4e and https://github.com/activecm/rita/pull/654/commits/180de847eb6ddb15e008891d9d690427ef307b45, I converted the `UniqueIPSet` implementation over from using a slice to a hashmap. The profiler indicated that the parsing code spent a great deal of time running linear searches against the UniqueIPSets when processing hostnames. Switching the implementation over to a hashmap sped up the process considerably at the expense of around 4x the RAM. However, RAM usage has not grown considerably overall.

- In https://github.com/activecm/rita/pull/654/commits/1cf8e1445c9e820e2297b0aae975ddf9390ec371, I changed the batching limit to take the greater of the old default (4GB) or half of the system's total memory. This allows us to run the MongoDB analysis phase quicker when the system can support it. 

- In https://github.com/activecm/rita/pull/654/commits/c4b5b3fa83b503af6da624a9727cbf1bfbaaa45a and https://github.com/activecm/rita/pull/654/commits/12215e03146ce03da8982f00cab3f594d9a1bbc6,  I replaced array based string sets which used calls to  `StringInSlice` with map based string sets similar to the UniqueIP sets found in https://github.com/activecm/rita/pull/654/commits/cd8dd38af38d30b06b0141d08b3370435ef2ba4e

- In https://github.com/activecm/rita/pull/654/commits/43a7dc78cf4462672ef287b161e03887524d51cd, I added a bit of code in `util.ContainsIP` and `util.IPIsPubliclyRoutable` to cache the results of `net.IP.To4()` which is called by many of the net api methods. It turns out that when you parse an IPv4 address in Go, the IP is stored using 16 bytes, but then the net api slices out the 4 bytes it needs via `To4()` in all the api methods. 

- In https://github.com/activecm/rita/pull/654/commits/4f78c1ef45acfad67c67288f9a84c50bbef6aa9f, I reordered the checks in `data.UniqueIP` to prevent parsing empty UUIDs. In previous versions, we'd call `uuid.Parse("")` at least twice for every log entry. Each of these calls would result in allocating an error object on the heap that we'd immediately throw away, creating a good deal of work for the garbage collector. 

- In https://github.com/activecm/rita/pull/654/commits/3028e62f6e152f1d0ffe000d7a26f3d616f0cee4, I fixed up the SSL parser to increment the hostMap source/ destination counters to address issue #684. 

#### Performance Testing
I've tested the changes on a large JSON dataset  as well as several small TSV datasets. 
The JSON dataset consisted of 3414 gzipped Zeek logs of which 2063 were conn, http, ssl, and dns logs. The 2063 logs that needed to be processed by RITA totaled 70.385 GB. 

While testing for performance, I used Go's pprof tool to find the sections of code which took up the majority of the execution time. 
I commented out lines in fsimporter.go which builds the MongoDB collections from the parse results beginning with 
```go
// Set chunk before we continue so if process dies, we still verify with a delete if
// any data was written out.
fs.metaDB.SetChunk(fs.config.S.Rolling.CurrentChunk, fs.database.GetSelectedDB(), true)
```
down to 
```go
// record file+database name hash in metadabase to prevent duplicate content
fmt.Println("\t[-] Indexing log entries ... ")
err := fs.metaDB.AddNewFilesToIndex(indexedFileBatch)
if err != nil {
	fs.log.Error("Could not update the list of parsed files")
}
```

I added a line to print out how long it took to import each batch of files to the original master branch and ran it against the large JSON dataset on a DO droplet with (virtualized) SSD storage, 16GB of RAM and 8 CPU cores. The results were as follows:
```
Master branch
        [-] Finished parsing logs in 5m28.344s
        [-] Finished parsing logs in 8m5.327s
        [-] Finished parsing logs in 10m8.033s
	[-] Finished parsing logs in 15m42.579s
        [-] Finished parsing logs in 20m5.076s
        [-] Finished parsing logs in 19m41.801s
        [-] Finished parsing logs in 19m4.292s
        [-] Finished parsing logs in 15m17.52s
        [-] Finished parsing logs in 15m12.626s
        [-] Finished parsing logs in 15m22.514s
        [-] Finished parsing logs in 15m52.611s
        [-] Finished parsing logs in 12m2.84s
        [-] Finished parsing logs in 7m47.568s
        [-] Finished parsing logs in 9m22.946s
        [-] Finished parsing logs in 7m12.611s
        [-] Finished parsing logs in 6m58.84s
        [-] Finished parsing logs in 9m58.732s
        [-] Finished parsing logs in 7m56.549s
	TOTAL: 3h41m20.809s
```
customer-cpu-parse-master-pprof

Compare this to the results after the PR changes:
```
        [-] Finished parsing logs in 3m7.787s
        [-] Finished parsing logs in 6m45.158s
        [-] Finished parsing logs in 6m5.804s
        [-] Finished parsing logs in 6m16.351s
        [-] Finished parsing logs in 6m14.18s
        [-] Finished parsing logs in 5m57.668s
        [-] Finished parsing logs in 6m30.06s
        [-] Finished parsing logs in 5m37.612s
        [-] Finished parsing logs in 5m23.952s
	TOTAL 0h:51m58.572s
```
customer-cpu-parse-4f78c1e-pprof

I've attached high resolution SVGs of the images above as well as the raw pprof data which can be opened via `go tool pprof`.
[pprof_data.zip](https://github.com/activecm/rita/files/6834032/pprof_data.zip)


Zalgo2462:
The latest master merge commit https://github.com/activecm/rita/pull/654/commits/93d094aa79a5eaa91e62cc985ae0c7e227940915 implements #665

parser/repository.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/activecm/rita/pkg/host"
+	"github.com/activecm/rita/pkg/hostname"
+	"github.com/activecm/rita/pkg/uconn"
+	"github.com/activecm/rita/pkg/useragent"
+)
+
+// ParseResults contains the data which the analysis packages
+// expect from the parser as well as locks for safely
+// accessing the data from multipel goroutines.
+type ParseResults struct {
```
Zalgo2462, line 17: `ParseResults` is a struct which holds all of the result maps from the previous version of the parser. Additionally, each map is associated with an individual mutex in order to provide more granular thread safety.

https://github.com/activecm/rita/commit/05627e59d3c2f7e27fb48cffa74534c8eab839d3

pkg/beaconfqdn/analyzer.go
```diff
@@ -242,7 +242,7 @@ func (a *analyzer) start() {
 					"score":              score,
 					"cid":                a.chunk,
 					"src_network_name":   entry.Src.SrcNetworkName,
-					"resolved_ips":       entry.ResolvedIPs,
+					"resolved_ips":       entry.ResolvedIPs.Items(),
```
Zalgo2462, line 245: Note that .Items() must be called to convert the map into an array. This is likely to create a merge conflict with #676 as `ResolvedIPs` is migrated from `data.UniqueIPSet` to `[]data.UniqueIP` in that PR.

https://github.com/activecm/rita/commit/cd8dd38af38d30b06b0141d08b3370435ef2ba4e
https://github.com/activecm/rita/commit/180de847eb6ddb15e008891d9d690427ef307b45

pkg/certificate/analyzer.go
```diff
@@ -62,26 +62,29 @@ func (a *analyzer) start() {
 			// cap the list to an arbitrary amount (hopefully smaller than the 16 MB document size cap)
 			// anything approaching this limit will cause performance issues in software that depends on rita
 			// anything tuncated over this limit won't be visible as an IP connecting to an invalid cert
-			if len(datum.OrigIps) > 200003 {
-				datum.OrigIps = datum.OrigIps[:200003]
+			origIPs := datum.OrigIps.Items()
```
Zalgo2462, line 65: Note that .Items() must be called in order to convert the map into an array. 

https://github.com/activecm/rita/commit/cd8dd38af38d30b06b0141d08b3370435ef2ba4e
https://github.com/activecm/rita/commit/180de847eb6ddb15e008891d9d690427ef307b45

pkg/certificate/analyzer.go
```diff
@@ -62,26 +62,29 @@ func (a *analyzer) start() {
...
+			if len(origIPs) > 200003 {
+				origIPs = origIPs[:200003]
 			}
 
-			if len(datum.Tuples) > 20 {
-				datum.Tuples = datum.Tuples[:20]
+			tuples := datum.Tuples.Items()
```
Zalgo2462, line 70: Note that `.Items()` must be called in order to convert the map into an array. 
https://github.com/activecm/rita/commit/c4b5b3fa83b503af6da624a9727cbf1bfbaaa45a

pkg/data/sets.go
```diff
@@ -0,0 +1,29 @@
+package data
+
+type StringSet map[string]struct{}
```
Zalgo2462, line 3: Provide a string set data structure to avoid needing to linearly scan through arrays to build sets of strings. 

https://github.com/activecm/rita/commit/c4b5b3fa83b503af6da624a9727cbf1bfbaaa45a

pkg/beaconfqdn/analyzer.go
```diff
@@ -86,7 +86,7 @@ func (a *analyzer) start() {
 					"avg_bytes":        entry.TotalBytes / entry.ConnectionCount,
 					"connection_count": entry.ConnectionCount,
 					"src_network_name": entry.Src.SrcNetworkName,
-					"resolved_ips":     entry.ResolvedIPs,
+					"resolved_ips":     entry.ResolvedIPs.Items(),
```
Zalgo2462, line 89: Note that .Items() must be called to convert the map into an array. This is likely to create a merge conflict with #676 as `ResolvedIPs` is migrated from `data.UniqueIPSet` to `[]data.UniqueIP` in that PR.

https://github.com/activecm/rita/commit/cd8dd38af38d30b06b0141d08b3370435ef2ba4e
https://github.com/activecm/rita/commit/180de847eb6ddb15e008891d9d690427ef307b45

pkg/hostname/analyzer.go
```diff
@@ -83,8 +83,8 @@ func (a *analyzer) start() {
 				output.query = bson.M{
 					"$push": bson.M{
 						"dat": bson.M{
-							"ips":     data.ResolvedIPs,
-							"src_ips": data.ClientIPs,
+							"ips":     data.ResolvedIPs.Items(),
```
Zalgo2462, line 86: Note that .Items() must be called to convert the map into an array. 

https://github.com/activecm/rita/commit/cd8dd38af38d30b06b0141d08b3370435ef2ba4e
https://github.com/activecm/rita/commit/180de847eb6ddb15e008891d9d690427ef307b45

pkg/uconn/analyzer.go
```diff
@@ -64,8 +64,9 @@ func (a *analyzer) start() {
 			// create query
 			query := bson.M{}
 
-			if len(datum.Tuples) > 5 {
-				datum.Tuples = datum.Tuples[:5]
+			tuples := datum.Tuples.Items()
```
Zalgo2462, line 67: Note that `.Items()` must be called in order to convert the map into an array. 
https://github.com/activecm/rita/commit/c4b5b3fa83b503af6da624a9727cbf1bfbaaa45a

pkg/useragent/analyzer.go
```diff
@@ -60,21 +60,23 @@ func (a *analyzer) start() {
...
-			if len(datum.OrigIps) > 10 {
-				datum.OrigIps = datum.OrigIps[:10]
+			origIPs := datum.OrigIps.Items()
+			if len(origIPs) > 10 {
+				origIPs = origIPs[:10]
 			}
 
-			if len(datum.Requests) > 10 {
-				datum.Requests = datum.Requests[:10]
+			requests := datum.Requests.Items()
```
Zalgo2462, line 68: Note that `.Items()` must be called in order to convert the map into an array. 
https://github.com/activecm/rita/commit/c4b5b3fa83b503af6da624a9727cbf1bfbaaa45a

pkg/useragent/analyzer.go
```diff
@@ -60,21 +60,23 @@ func (a *analyzer) start() {
 			// set up writer output
 			var output update
 
-			if len(datum.OrigIps) > 10 {
-				datum.OrigIps = datum.OrigIps[:10]
+			origIPs := datum.OrigIps.Items()
```
Zalgo2462, line 63: Note that .Items() must be called to convert the map into an array. 

https://github.com/activecm/rita/commit/cd8dd38af38d30b06b0141d08b3370435ef2ba4e
https://github.com/activecm/rita/commit/180de847eb6ddb15e008891d9d690427ef307b45

parser/fsimporter.go
```diff
@@ -58,22 +52,16 @@ type (
 )
 
 //NewFSImporter creates a new file system importer
-func NewFSImporter(res *resources.Resources,
-	indexingThreads int, parseThreads int, importFiles []string) *FSImporter {
+func NewFSImporter(res *resources.Resources) *FSImporter {
+	// set batchSize to the max of 4GB or a half of system RAM to prevent running out of memory while importing
+	batchSize := int64(util.MaxUint64(4*(1<<30), (memory.TotalMemory() / 2)))
```
Zalgo2462, line 57: In 1cf8e14, I changed the batching limit to take the greater of the old default (4GB) or half of the system's total memory. This allows us to run the MongoDB analysis phase quicker when the system can support it.

util/ip.go
```diff
@@ -54,6 +54,11 @@ func ParseSubnets(subnets []string) (parsedSubnets []*net.IPNet) {
 
 //IPIsPubliclyRoutable checks if an IP address is publicly routable. See privateIPBlocks.
 func IPIsPubliclyRoutable(ip net.IP) bool {
+	// cache IPv4 conversion so it not performed every in every ip.IsXXX method
```
Zalgo2462, line 57: In 43a7dc7, I added a bit of code in util.ContainsIP and util.IPIsPubliclyRoutable to cache the results of net.IP.To4() which is called by many of the net api methods. It turns out that when you parse an IPv4 address in Go, the IP is stored using 16 bytes, but then the net api slices out the 4 bytes it needs via To4() in all the api methods.

parser/ssl.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	// ///// SET INVALID CERTIFICATE FLAG FOR UNIQUE CONNECTION /////
+	if certificateIsInvalid {
+		retVals.UniqueConnMap[srcDstKey].InvalidCertFlag = true
+	}
+	return
+}
+
+func updateHostsBySSL(srcIP, dstIP net.IP, srcUniqIP, dstUniqIP data.UniqueIP, srcKey, dstKey string,
```
Zalgo2462, line 109: Fixes issue #684.

util/util.go
```diff
@@ -88,3 +99,40 @@ func StringInSlice(value string, list []string) bool {
 	}
 	return false
 }
+
+//Int64InSlice returns true if the int64 is an element of the array
+func Int64InSlice(value int64, list []int64) bool {
```
Zalgo2462, line 104: It seemed the `int64` linear searches did not take up too much time (mainly used when unique'ing unique connection timestamps). My guess is that the go compiler is smart enough to optimize the linear scan using SIMD or that the CPU is smart enough to speculatively execute checks in parallel.  (Moved from fsimporter)

Zalgo2462 (commented):
Left some comments noting where the different commits pop up in the PR.

Zalgo2462:
Merged master back into this branch. 
Integrated #690 and double checked that the rolling proxy beacons were still being reported correctly.

pkg/beaconproxy/mongodb.go
```diff
@@ -3,39 +3,42 @@ package beaconproxy
...
 	"github.com/vbauerster/mpb/decor"
+
+	log "github.com/sirupsen/logrus"
 )
 
 type repo struct {
-	res *resources.Resources
-	min int64
-	max int64
+	database *database.DB
```
fullmetalcache, line 19: Nice, I like how this removes a level from a lot of the places where we use this struct. Makes things a bit cleaner

pkg/data/ip.go
```diff
@@ -202,23 +209,24 @@ func (p UniqueIPPair) BSONKey() bson.M {
 //UniqueIPSet is a set of UniqueIPs which contains at most one instance of each UniqueIP
 //this implementation is based on a slice of UniqueIPs rather than a map[string]UniqueIP
 //since it requires less RAM.
-type UniqueIPSet []UniqueIP
+type UniqueIPSet map[string]UniqueIP
```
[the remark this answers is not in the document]
fullmetalcache: Very nice on the type replacement

fullmetalcache (commented):
On a first pass, everything looks good to me. Looks like the majority of changes were splitting out the parsers into different files, changing variable types/functions called, other optimization tricks, and some renaming/reorganizing of other variable types to improve readability. Just one minor question on the renaming of bro to zeek. Not a big deal though.

I will test this out and give it a second pass tomorrow. If everything else checks out, I am good with approving this.

Nice work man, so many good changes and learned so much here! Incredible how some of those variable types and function choices can make such a huge difference. If we don't already have something, we should consider having a doc or something else with some of these efficiency findings for types and functions so that we can refer to them in the future.

parser/files/reading.go
```diff
@@ -0,0 +1,463 @@
...
+				"missing_field": name,
+			}).Info("the log contains a field with no candidate in the data structure")
+			continue
+		}
+
+		if header.Types[index] != fieldInfo.zeekType {
+			err := errors.New("type mismatch found in log")
+			logger.WithFields(log.Fields{
+				"error":         err,
+				"type in log":   header.Types[index],
```
fullmetalcache, line 279: Minor thing but can we please remove the spaces and replace with underscores for the field names?

fullmetalcache (changes requested):
Find just one very minor thing on a second pass. Tested on a semi-large set on a system with 16GB RAM and 8 Cores allocated that is running on a server with 8 10K disks in RAID 10 config. Got around 8 minutes of parse time on current master. Got around 4 minutes with these changes. Nice!

parser/files/reading.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	if len(toReturn.Names) != len(toReturn.Types) {
+		return toReturn, errors.New("name / type mismatch")
+	}
+	return toReturn, nil
+}
+
+func mapZeekHeaderToParseType(header *BroHeader, broDataFactory func() pt.BroData, logger *log.Logger) (ZeekHeaderIndexMap, error) {
+	broData := broDataFactory()
```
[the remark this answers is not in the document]
Zalgo2462: I kept this one as `broData` since the type was kept as `pt.BroData` from previous versions. I've added an issue to the tracker to generally rename the references from Bro to Zeek (#693).

parser/files/reading.go
Zalgo2462: Will do

fullmetalcache (approved):
LGTM!

PR #655, opened 2021-07-07
Add Config Option to Filter External to Internal Communications

### LINKED ISSUE #522: Add filter option for external to internal traffic
Besides incoming blacklisted connections, external to internal traffic isn't super useful in any of our analysis modules. And incoming blacklisted connections is of questionable usefulness as well since the things that normally scan everything on the internet will also normally end up on blacklists. We're not trying to detect someone attacking coming in. We're trying to detect already compromised systems calling out.

To that end, we should add a new filtering option to allow disabling of external to internal traffic. We'd leave it enabled by default to preserve current functionality for now. But it would allow us to tune results further and reduce false positives.

fullmetalcache:
Adds the configuration option and logic for filtering external to internal communications.

Closes #522


Zalgo2462:
We should hold off on merging this in until https://github.com/activecm/rita/pull/654 is merged since `filterConnPair` has been moved out of the `fsimporter` in that PR.

Zalgo2462:
@fullmetalcache I've merged master back in and fixed the merge conflicts. Please check this over to make sure it meets your expectations before I review it.

Zalgo2462 (approved):
Tested with the following query against master to make sure external to internal connections are filtered out:

```js
db.getCollection('uconn').aggregate([
    {"$lookup": {
        "from": "host",
        "localField": "src",
        "foreignField": "ip",
        "as": "src_details"
    }},
    {"$match": {
        "src_details.local": false,
    }}
])
```

PR #660, opened 2021-07-15
extend zeek inactivity timeout.

william-stearns:
This mirrors what we do in docker-zeek.


ethack (approved):
LGTM.

My only comment is that the `>>$local_path/local.zeek` is not quoted. I know we set `$_ZEEK_PATH` elsewhere in the file and it's not likely to ever have a space though.

william-stearns:
Quoting updated on $local_path references too, thanks for the catch.

ethack (approved):

Zalgo2462 (approved):
Performed new install and upgrade on Ubuntu 18. LGTM.

PR #662, opened 2021-07-17
Add timestamp to HTML report templates

### LINKED ISSUE #594: Add Date/Time Stamp
Would it be possible to add a date/time stamp of when the HTML report was ran to the navigation bar on the top of the index.html file? Maybe next to User Agents on the far right or something. There's been a few times where we review the data and didn't realize that a new report failed to generate.

edwmorgan:
This adds a date/time block to the top bar of the HTML report representing the `maxTime` of the generated logs. The timestamp is visible from all panels in the HTML report.

Closes #594

### DIFF
--- a/reporting/report.go
+++ b/reporting/report.go
@@ -145,11 +145,7 @@ func writeDB(db string, wd string, showNetNames bool, res *resources.Resources) 
 	}
 	res.DB.SelectDB(db)
 
-	dbMetaInfo, err := res.MetaDB.GetDBMetaInfo(db)
-	if err != nil {
-		return err
-	}
-	maxTime := time.Unix(dbMetaInfo.TsRange.Max, 0).String()
+	maxTime := time.Now().Format(time.RFC1123)
 
 
 	err = writeDBHomePage(db, maxTime)
--- a/reporting/report.go
+++ b/reporting/report.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"strconv"
+	"time"
 
 	htmlTempl "github.com/activecm/rita/reporting/templates"
 	"github.com/activecm/rita/resources"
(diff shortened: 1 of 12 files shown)


reporting/report.go
```diff
@@ -144,53 +145,60 @@ func writeDB(db string, wd string, showNetNames bool, res *resources.Resources)
 	}
 	res.DB.SelectDB(db)
 
-	err = writeDBHomePage(db)
+	dbMetaInfo, err := res.MetaDB.GetDBMetaInfo(db)
+	if err != nil {
+		return err
+	}
+	maxTime := time.Unix(dbMetaInfo.TsRange.Max, 0).String()
```
ethack, line 152: I messed up in my comment in issue https://github.com/activecm/rita/issues/594#issuecomment-744650240. @Nyhmesis was asking for a timestamp of when the report was generated. I suggested pulling from the database, which is the timestamp contained in the imported log entries. These are both valuable but very different.

For instance, I just generated a report with some sample data from a few years ago. Even though the report was generated today 2021-08-03, it shows a timestamp from 2018, which is when the sample data was captured.

image

To make this true to the original issue, could you change it to use `time.Now()` instead? I think this should work (but haven't tested it).

```go
maxTime := time.Now().String()
```

ethack (approved):
Looks good except for me sending you in the wrong direction for where to get the time from :innocent:

ethack (changes requested):
Looks good except for me sending you in the wrong direction for where to get the time from :innocent:

reporting/report.go
edwmorgan: Ah, that makes more sense. I'll make those changes.

edwmorgan:
Made the change. I used the RFC1123 formatter as it seemed (to me) the most human-readable option. Let me know if one of the other formatters would be better: https://pkg.go.dev/time#pkg-constants

ethack (approved):
Looks good to me. Thank you for the contribution!

PR #665, opened 2021-07-22
Remove Need for Users to Specify Proxy Servers, Fix Filter Bugs

### LINKED ISSUE #664: Process all CONNECT Entries for Beacons Proxy Module
Currently, users are required to specify the IP addresses of their proxy servers for the module to work correctly. Additionally, if any of the proxy addresses are internal, then the program will ignore the entries due to no exception in place for the internal->internal filter.

fullmetalcache:
Removes the need for a user to specify their proxy servers for the beacons proxy module. Now, all http.log entries with the CONNECT method will automatically be considered for processing for the beacons proxy module. Entries will also be ignored if the Src IP or FQDN have been specified for filtering.

This also ensure that entries are not filtered if they are an internal->internal connection

Closes #664


Zalgo2462 (approved):
Tested well. Code change looks good. 

These changes will need to be reflected in #654.

PR #668, opened 2021-07-29
Remove unnecessary deserialization logic

Zalgo2462:
This PR speeds up all three types of beaconing analysis by removing unnecessary deserialization logic. 

In several places, RITA checks whether documents exist in MongoDB by calling `myQuery.All(&myArr)` followed by `len(myArr) == 0`. Calling `.All` causes the MongoDB driver to load all of the resulting BSON data into RAM as Go types, wasting CPU cycles. 

Replacing these calls with `myQuery.Count()` avoids the deserialization overhead.

I profiled FQDN beaconing using pprof and trace and found that this deserialization made up 5/6ths of all the CPU time in the Go code (including the actual beaconing analysis!). 

#### Example performance test after applying changes to beacon, fqdn beacon, and proxy beacon analyses
I tested the changes on a dataset with a DNS C2 channel and found a significant speed up.

##### old: e6c740f (master at the time of writing)
```
time ./rita import dnscat2-ja3 dnscat2-ja3
real    5m24.634s
user    4m44.003s
sys     0m54.479s
```

##### new: dbd2d6
```
time ./rita import dnscat2-ja3 dnscat2-ja3
real    2m39.016s
user    2m53.425s
sys     0m35.254s
```


database/meta.go
```diff
@@ -355,8 +355,8 @@ func (m *MetaDB) MarkDBAnalyzed(name string, complete bool) error {
 	_, err = ssn.DB(m.config.S.MongoDB.MetaDB).C(m.config.T.Meta.DatabasesTable).
 		Upsert(bson.M{"_id": dbr.ID}, bson.M{
 			"$set": bson.D{
-				{"analyzed", complete},
-				{"analyze_version", versionTag},
+				{Name: "analyzed", Value: complete},
+				{Name: "analyze_version", Value: versionTag},
```
Zalgo2462, line 359: Linter fixes. The linter doesn't like it when you use define a struct without listing the key names. Fixed it up since the linter wouldn't quit yelling after I changed the lines below.

database/meta.go
```diff
@@ -433,14 +433,13 @@ func (m *MetaDB) IsChunkSet(cid int, db string) (bool, error) {
 		},
 	}
 
-	var results []interface{}
-	err := ssn.DB(m.config.S.MongoDB.MetaDB).C(m.config.T.Meta.DatabasesTable).Find(query).All(&results)
+	queryCount, err := ssn.DB(m.config.S.MongoDB.MetaDB).C(m.config.T.Meta.DatabasesTable).Find(query).Count()
```
Zalgo2462, line 436: Checks if a chunk record exists for the given db and chunk.

parser/fsimporter.go
```diff
@@ -961,7 +961,7 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
 							ja3Hash := parseSSL.JA3
 							src := parseSSL.Source
 							dst := parseSSL.Destination
-							host := parseSSL.ServerName
+							hostName := parseSSL.ServerName
```
Zalgo2462, line 964: The changes in this file relate to #684 and are in this PR to allow testing of the `host` collection connection count records. These changes are made irrelevant by #654 as an alternative fix is implemented there.

pkg/beacon/analyzer.go
```diff
@@ -295,46 +299,51 @@ func (a *analyzer) hostIcertQuery(icert bool, src data.UniqueIP, dst data.Unique
 	// create query
 	query := bson.M{}
 
-	// update host table if there is an invalid cert record between pair
-	if icert {
-
-		newFlag := false
+	// don't update host table if there isn't an invalid cert record between pair
+	if !icert {
```
Zalgo2462, line 303: Inverted the check and dedented, producing cleaner code but a gnarly diff :/

pkg/beacon/analyzer.go
```diff
@@ -295,46 +299,51 @@ func (a *analyzer) hostIcertQuery(icert bool, src data.UniqueIP, dst data.Unique
...
+		return updateInfo{}
+	}
 
-		var resList []interface{}
+	hostSelector := src.BSONKey()
+	hostSelector["dat.icdst"] = dst.BSONKey()
```
Zalgo2462, line 308: There is a bug here that is fixed in #683 and there will likely be merge conflicts.

pkg/host/analyzer.go
```diff
@@ -112,16 +112,14 @@ func (a *analyzer) start() {
 	}()
 }
 
-//shouldInsertNewRecord returns true if a host entry with the current CID does not exist in the database
-func (a *analyzer) shouldInsertNewHostRecord(ssn *mgo.Session, host data.UniqueIP) bool {
-	var hostCIDs []struct {
-		CID int `bson:"cid"`
-	}
+//shouldInsertNewHostSubdocument returns true if a host entry with the current CID does not exist in the database
+func (a *analyzer) shouldInsertNewHostSubdocument(ssn *mgo.Session, host data.UniqueIP) bool {
```
Zalgo2462, line 116: renamed because I was confused as I thought this function was checking whether a new top level entry should be inserted into the host collection

pkg/uconn/results.go
```diff
@@ -15,8 +15,8 @@ func LongConnResults(res *resources.Resources, thresh int, limit int, noLimit bo
 	var longConnResults []LongConnResult
 
 	longConnQuery := []bson.M{
-		bson.M{"$match": bson.M{"dat.maxdur": bson.M{"$gt": thresh}}},
```
Zalgo2462, line 18: linting fixes

pkg/useragent/analyzer.go
```diff
@@ -124,22 +124,16 @@ func (a *analyzer) start() {
...
-					//TODO: Consider adding the chunk to the query instead of checking it after the query
+					entryHostQuery["cid"] = a.chunk
 
-					_ = ssn.DB(a.db.GetSelectedDB()).C(a.conf.T.Structure.HostTable).Find(entryHostQuery).All(&hostEntries)
+					nExistingEntries, _ := ssn.DB(a.db.GetSelectedDB()).C(a.conf.T.Structure.HostTable).Find(entryHostQuery).Count()
 
 					//TODO: I think there might be a bug here. From what I can tell, if we set the new record flag
 					// when there is a rare signature record from a previous chunk, then we will end up pushing a duplicate
 					// record for each chunk. We should investigate removing this chunk check here. -LL
-					if len(hostEntries) <= 0 || hostEntries[0].CID != a.chunk {
```
Zalgo2462, line 142: Moved the chunk check into the mongoDB query ( the first TODO). However, this analyzer is pretty much completely rewritten in #683 and there will be merge conflicts.

pkg/useragent/results.go
```diff
@@ -16,18 +16,18 @@ func Results(res *resources.Resources, sortDirection, limit int, noLimit bool) (
 	var useragentResults []Result
 
 	useragentQuery := []bson.M{
-		bson.M{"$project": bson.M{"user_agent": 1, "seen": "$dat.seen"}},
```
Zalgo2462, line 19: linting fixes

fullmetalcache (approved):
Did some simple timing tests and things look good! I wrapped timing checks around the whole block of "fs.buildX" function calls. I know this isn't the most accurate for this but it gives a general idea of if this makes things faster. I tested on three data sets and found that this improves the speed by about 20%. Nice!

There is a bug related to adding rare signature scores to the host table. It appears that there might be a race condition. We are hoping that #683 will address this. This PR will be approved and we will deal with the bug if it pops up again while testing #683 

LGTM!

PR #676, opened 2021-08-02
Gather FQDN IPs using DNS lookups from previous chunks

### LINKED ISSUE #669: FQDN beaconing analysis only aggregates IPs which appear in the current chunked import's DNS logs
#### Current behavior:

RITA performs FQDN beaconing analysis by grouping the unique connections between a source IP and the destination IPs associated with an FQDN. Currently, RITA only associates those IPs which appeared in the current chunk's DNS logs with each hostname. 

This may lead to FQDN beacons not being properly updated. Consider importing a rolling dataset over 24 hours with 24 chunks. In one chunk RITA may see two IP addresses associated with a particular FQDN. As a result, RITA will group these two IPs together and insert a record into the `beaconFQDN` collection for each source which connected to at least one of these IPs.  However, if the IP addresses no longer appear in the DNS logs imported under subsequent chunks, the `beaconFQDN` records will not be updated with any new connections to the IP addresses. This is particularly important when considering caching DNS resolvers.

Additionally, when importing large datasets, RITA may import subsets of the log files into the same chunk. This means that RITA may consider different groupings of IP addresses for the same FQDN while updating the same chunk, leading to consistency issues.

#### Desired behavior:

RITA should consider all of the IP addresses associated with a given FQDN as recorded in the `hostnames` collection when grouping unique connections for FQDN beacon analysis.

In the case of performing a rollign import of 24 chunks over 24 hours, this would ensure that the FQDN beaconing analysis would group IPs for each hostname using the past 24 hours of DNS lookups rather than just the last hour's worth.

In the case of importing a large dataset, this would ensure that unique connections associated with an FQDN beacon would never go missing from the analysis.

Zalgo2462:
Fixes #669 by having the FQDN beaconing module extract all of the external hosts we saw in the parse results from the `hostMap` and query the `hostnames` collection to obtain the list of FQDNs we need to update along with all of the associated IPs.

Note: old FQDN -> IP mappings are removed from the `hostnames` collection by the remover package in the function `pkg/remover/mongodb.go:removeOutdatedCIDs(cid int) error`


.gitignore
```diff
@@ -4,6 +4,7 @@
 *.swp
 *.swo
 *.exe
+/.vscode/
```
Zalgo2462, line 7: I used `dlv` while developing this PR. VSCode stores `dlv` configurations in `.vscode`. VSCode is free to use for both open source and commercial development.

go.mod
```diff
@@ -12,14 +12,14 @@ require (
 	github.com/activecm/mgosec v0.1.1
 	github.com/activecm/rita-bl v0.0.0-20200806232046-0db4a39fcf49
 	github.com/blang/semver v3.5.1+incompatible
+	github.com/briandowns/spinner v1.16.0
```
Zalgo2462, line 15: While running the query which gathers the hostnames associated with the newly parsed in host IPs, we display a busy spinner.

parser/fsimporter.go
```diff
@@ -199,7 +199,7 @@ func (fs *FSImporter) Run(indexedFiles []*fpt.IndexedFile) {
 		fs.buildBeacons(uconnMap)
 
 		// build or update the FQDN Beacons Table
-		fs.buildFQDNBeacons(hostnameMap)
+		fs.buildFQDNBeacons(hostMap)
```
Zalgo2462, line 202: FQDNBeacons now takes in the list of hosts in the current parse set rather than the list of DNS queries.

pkg/beaconfqdn/mongodb.go
```diff
@@ -117,9 +122,31 @@ func (r *repo) Upsert(hostnameMap map[string]*hostname.Input, minTimestamp, maxT
 		writerWorker.start()
 	}
 
+	s := spinner.New(spinner.CharSets[36], 200*time.Millisecond)
+	s.Prefix = "\t[-] Gathering FQDNs for Beacon Analysis ...\t"
+	s.Start()
```
fullmetalcache, line 127: Something seems to be just a bit off with moving from this spinner to the next spinner. I tested twice and see something like this each time:

image

parser/fsimporter.go
```diff
@@ -575,7 +575,7 @@ func (fs *FSImporter) buildFQDNBeacons(hostnameMap map[string]*hostname.Input, m
 			}
 
 			// send uconns to beacon analysis
```
fullmetalcache, line 577: I know this comment is from before, but it needs updated to read "hostMap" instead of uconns

pkg/beaconfqdn/mongodb.go
```diff
@@ -151,8 +178,189 @@ func (r *repo) Upsert(hostnameMap map[string]*hostname.Input, minTimestamp, maxT
...
+	return affectedHostnamesBuffer, err
+}
+
+// affectedHostnameIPsChunked implements affectedHostnameIPs. It is slower than affectedHostnameIPs_simple
+// and uses more RAM. However, unlike affectedHostnameIPs_simple, affectedHostnameIPs_chunked handles
+// hostMaps of all sizes.
+func (r *repo) affectedHostnameIPsChunked(hostMap map[string]*host.Input) ([]hostnameIPs, error) {
+	// preallocate externalHosts slice assuming at least half of the observed hosts are external
+	// util.Min ensures that we don't preallocate more than the maximum allowed number of hosts in a 16MB BSON document
+	var externalHosts []data.UniqueIP = make([]data.UniqueIP, 0, util.Min(200000, len(hostMap)/2))
```
fullmetalcache, line 233: Same comment here as above

pkg/beaconfqdn/mongodb.go
```diff
@@ -151,8 +178,189 @@ func (r *repo) Upsert(hostnameMap map[string]*hostname.Input, minTimestamp, maxT
...
+			"network_uuid": "$_id.network_uuid",
+		}},
+	}}
+])
+
+reverseDNSQueryWithIPs returns a MongoDB aggregation which returns the hostnames associated with the given
+UniqueIPs. Additionally, all of the IPs associated with each hostname are returned.
+*/
+func reverseDNSQueryWithIPs(uniqueIPs []data.UniqueIP) []bson.M {
+	var uniqueIPBsonSelectors []bson.M = make([]bson.M, 0, len(uniqueIPs))
```
fullmetalcache, line 335: Not a big deal and will just comment in this one place, but I don't think the "var" and data type aren't needed on the left with the "make" on the right.

pkg/beaconfqdn/mongodb.go
```diff
@@ -151,8 +178,189 @@ func (r *repo) Upsert(hostnameMap map[string]*hostname.Input, minTimestamp, maxT
...
+	var affectedHostnamesBuffer []hostnameIPs
+
+	ssn := r.database.Session.Copy()
+	defer ssn.Close()
+
+	// we will need to remove duplicate results from each query of 200,000 hosts, slowing down the process
+	// and consuming more RAM
+
+	// affectedHostnameIPMap maps hostnames to their respective ResolvedIPs
+	var affectedHostnameIPMap map[string][]data.UniqueIP = make(map[string][]data.UniqueIP, len(hostMap)/10)
```
fullmetalcache, line 243: Maybe quick comment here explaining the /10 factor.

fullmetalcache (changes requested):
Just a few minor comments/discussion ideas and a small bug on the first pass. Tested on gcat set and it seems to produce the same FQDN results as before, which is good!

To further test, I broke the gcat set out into two folders. In the first folder, I removed all of the conn logs. This meant that the first import would get all of the DNS entries but wouldn't get any of the conn info. We will call the imported dataset "gcat-test". The second folder had just the conn logs. I imported the conn logs into the  gcat-test set that was created with the first import. This meant that the second import would have to rely solely on the DNS entries from the first import for beacon FQDN analysis. After the first import, there were no beaconsFQDN results. After the second import, the number of beaconsFQDN results matched the number we had previously for the gcat sample-data set. I think this is good evidence that it's working.

Might want to do some quick tests on the amount of time that this adds on to the beacons FQDN processing. I know we've talked about it and have each looked at it, but at some point we will probably want to give everything in beacons FQDN yet another look over for any more tricks we can do to speed everything up.

pkg/beaconfqdn/mongodb.go
```diff
@@ -151,8 +178,189 @@ func (r *repo) Upsert(hostnameMap map[string]*hostname.Input, minTimestamp, maxT
...
+		return r.affectedHostnameIPsSimple(hostMap)
+	}
+	return r.affectedHostnameIPsChunked(hostMap)
+}
+
+// affectedHostnameIPsSimple implements affectedHostnameIPs but should only be called with hostMaps with
+// roughly less than 200,000 external hosts.
+func (r *repo) affectedHostnameIPsSimple(hostMap map[string]*host.Input) ([]hostnameIPs, error) {
+	// preallocate externalHosts slice assuming at least half of the observed hosts are external
+	var externalHosts []data.UniqueIP = make([]data.UniqueIP, 0, len(hostMap)/2)
```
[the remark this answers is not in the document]
Zalgo2462: The only disadvantage is RAM usage as you said. 

I went with `/2` because I took a guess at how slice allocation works and assumed that it doubled the size of the backing array when reallocation happens. It looks like that Go does actually double the size of the underlying slice when you `append` more items than the slice has capacity fore: https://github.com/golang/go/blob/0a7f00ae239570d664488085f9c395919bc69066/src/runtime/slice.go#L164. 

So, with `/2`, we guarantee that we will only need to re-allocate once in the worst case. I could add that in as a comment if you'd like.

I'm happy to change it to `len(hostMap)` however if you feel that best suits our needs.

pkg/beaconfqdn/mongodb.go
Zalgo2462: I cannot reproduce this issue in my terminal. In my terminal, the second spinner overwrites the first one.
I'll try adding a `println` call after `s.Stop()`, but I'll need you to let me know whether it helped or not.

pkg/beaconfqdn/mongodb.go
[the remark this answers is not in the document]
fullmetalcache: Oh nope, if there is just one re-allocation then I think it's worth it to go minimal. I was just concerned if it was going to be re-allocating with every new entry. This way is great though if it just does the one.

pkg/beaconfqdn/mongodb.go
fullmetalcache: It could be an ESXi thing. I've had other weird terminal things happen too that nobody else experienced.

pkg/beaconfqdn/mongodb.go
Zalgo2462: This assumes that there would be roughly 10 IPs per observed hostname in the dataset. I'm not sure that's a good assumption to be honest. I didn't want to pre-allocate with `len(hostMap)` since that would obviously be too big. But, I also didn't want to start from 0. Thoughts?

pkg/beaconfqdn/mongodb.go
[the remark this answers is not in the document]
Zalgo2462: Chill, I'll add the explanation above as a comment.

pkg/beaconfqdn/mongodb.go
fullmetalcache: I think 10 is reasonable. Maybe just add a comment mentioning that?

fullmetalcache: Oh hang on though, this wouldn't be 10 IPs per host. This would be len(hostMap)/10 IPs per host.

Zalgo2462:
I've addressed the changes by adding comments, using short variable assignments, and adding a call to `fmt.Println` after stopping the spinner. Please let me know if you run into any further issues.

> Might want to do some quick tests on the amount of time that this adds on to the beacons FQDN processing. I know we've talked about it and have each looked at it, but at some point we will probably want to give everything in beacons FQDN yet 
another look over for any more tricks we can do to speed everything up.

In my testing, on small datasets, the additional query performed here cancelled out the performance gains in the FQDN Beaconing analysis made by PR #668. On large datasets, the time savings made in #668 outweighed the additional processing time required by this PR. However, FQDN beaconing still seems rather slow on large datasets. I believe that addressing issue #686 may help speed up the process as well. In general, I agree that in the future, time will need to be devoted towards speeding up the FQDN analysis specifically.

pkg/beaconfqdn/mongodb.go
Zalgo2462: len(hostMap)/10 per hostname* or rather #IPs/10 : #FQDNs. 

So, this assumes that for every 10 IPs in the parsed in set of logs, there would be 1 FQDN.

fullmetalcache: Got it, was reading the make statement wrong!

fullmetalcache:
image
 The println seems to fix help. That chart didn't seem to finish for me in either case. I am not sure that's it's worth holding things up for though.

fullmetalcache (approved):

fullmetalcache:
Approving. The progress bar could very well be an issue just on my system. Up to you if you want to fix it. Otherwise, merge when you are ready!

Zalgo2462:
> image
> The println seems to help. That chart didn't seem to finish for me in either case. I am not sure that's it's worth holding things up for though.

In general, that chart won't finish out. Since we don't know know how long it will take, the animation just repeats itself until its done and when you call Stop() the animation stops where it was at.

PR #687, opened 2021-08-10
Add Additional Indexes to Hosts Collection for Faster Lookups

fullmetalcache:
Adding in some more indexes to the hosts collection to allow for faster lookup of some items.


Zalgo2462 (approved):
LGTM

PR #689, opened 2021-08-11
Add unique indexes to `beaconFQDN` and `beaconProxy` collections

### LINKED ISSUE #688: FQDN and Proxy Beacon updates are not properly indexed
When the FQDN analysis performs an upsert against the `beaconFQDN` collection, it uses the following selector:

https://github.com/activecm/rita/blob/e6c740f6ef1e2b88950eb63b1f3c6538720d0de3/pkg/beaconfqdn/repository.go#L82-L86

However, this selector is not fully covered by a MongoDB index:
https://github.com/activecm/rita/blob/e6c740f6ef1e2b88950eb63b1f3c6538720d0de3/pkg/beaconfqdn/mongodb.go#L50-L55

This means that MongoDB must select one of the indexes to perform a partial search against, and then perform a linear search over the partial search results to complete the query. 

image

With some datasets, I have seen MongoDB perform a linear search over ~300 documents for each `beaconFQDN` upsert.

A similar problem exists in the `beaconProxy` code.

Selector:
https://github.com/activecm/rita/blob/e6c740f6ef1e2b88950eb63b1f3c6538720d0de3/pkg/beaconproxy/repository.go#L123-L129

Indexes:
https://github.com/activecm/rita/blob/e6c740f6ef1e2b88950eb63b1f3c6538720d0de3/pkg/beaconproxy/mongodb.go#L48-L54

Unique indexes should be added to cover these upsert selectors such that no linear searches should be performed.

Zalgo2462:
Fixes #688

Adds the necessary unique indexes to cover the upserts performed in the fqdn and proxy beacon analyses. 

Running a `.explain` for the FQDN beacon selector (see #688) now returns the following plan without a filter stage:
image

### DIFF
--- a/pkg/beaconfqdn/mongodb.go
+++ b/pkg/beaconfqdn/mongodb.go
@@ -49,6 +49,7 @@ func (r *repo) CreateIndexes() error {
 	// set desired indexes
 	indexes := []mgo.Index{
 		{Key: []string{"-score"}},
+		{Key: []string{"src", "fqdn", "src_network_uuid"}, Unique: true},
 		{Key: []string{"src", "src_network_uuid"}},
 		{Key: []string{"fqdn"}},
 		{Key: []string{"-connection_count"}},


pkg/beaconproxy/mongodb.go
```diff
@@ -47,6 +47,7 @@ func (r *repo) CreateIndexes() error {
 	// set desired indexes
 	indexes := []mgo.Index{
 		{Key: []string{"-score"}},
+		{Key: []string{"src", "dst", "fqdn", "src_network_uuid", "dst_network_uuid"}, Unique: true},
```
fullmetalcache, line 50: In #690 , we are changing the keying to be src-fqdn and cutting out the dst (i.e., proxy) as part of the keying since it isn't really relevant for analysis. Can we please cut dst and dst_network_uuid from this key and just have it be "src" and "fqdn"?

fullmetalcache (changes requested):
Just the one change please, the other key looks good!

Thanks!

pkg/beaconproxy/mongodb.go
Zalgo2462: Let's revert the `beaconProxy` index commit from this PR and apply the change only in #690.

Zalgo2462:
Removed the change to `beaconProxy` and left [a comment](https://github.com/activecm/rita/pull/690#discussion_r690803155) on #690 with the fix to be applied there.

fullmetalcache (approved):
lgtm!

PR #690, opened 2021-08-17
Analyze Past 24 Hours with Beacons Proxy

### LINKED ISSUE #670: Proxy beaconing analysis only aggregates connections which appeared in the current chunked import's HTTP logs
#### Current Behavior:
Currently, RITA only considers the proxy connections found in the current chunk's HTTP logs. 

This may lead to proxy beacons not being properly updated. Consider importing a rolling dataset over 24 hours with 24 chunks. Each proxy beacon will only consider data from the past hour, not the past 24 hours. 

Additionally, when importing large datasets, RITA may import subsets of the log files into the same chunk. This means that RITA may consider different connections for the same proxy beacon while updating the same chunk, leading to consistency issues.

#### Desired Behavior:
Proxy beaconing should consider proxy connections from past chunks. 

We will likely need to store the timestamps for each `(source IP, proxied FQDN)` unique connection alongside the summary results in a `dat` array in the `beaconProxy` collection, or create a new `proxyUconns` collection. 

When importing 24 hours of data across 24 chunks, this will enable RITA to aggregate proxied connections from the past 24 hours rather than just the past hour.

In the case of importing a large dataset, this would ensure that proxied connections would never be dropped from the proxy beacon analysis.

### LINKED ISSUE #691: http.log Can Contain Blank "Host" Field
When testing beacon proxy analysis, I found that the "host" field can sometimes be blank in http.log. This can result in strange behavior for some modules, particularly beacons proxy.

fullmetalcache:
Shoutout to @lisaSW for closely working together with me on this. Was a team effort!

Previously, beacons proxy was only analyzing the current hour's worth of data. This PR is to address this issue so that beacons proxy will analyze the rolling 24 hour period, as is done with other modules. A new table, uconnproxy, is created to facilitate this behavior.

Beacons proxy no longer uses the proxy IP as part of the key for uconnproxy entries (previously beaconproxymap entries). This is because the proxy IP is not relevant in terms of determining beaconing behavior from a host to an fqdn; the proxy IP purely serves as an intermediary. The uconnproxy entries are now keyed on just src-fqdn pairs.

This PR also includes a bug fix for when the "host" field in the http.log file is blank, which was found to occur in some cases during testing. The backup measure that is used is to parse the fqdn from the URI field. This backup measure is the secondary choice as it takes additional processing time over just pulling the fqdn from the host field.

Minor renaming of mongo variables with respect to the proxy IP was also done to help avoid some confusion with the module (i.e., dst_x info is now "proxy", which contains the relevant host info of ip, network name, network uuid).

The module does not store a list of proxy IPs if the server happens to change during the analysis period; only the most recent proxy server will be stored. 

Closes #670
Closes #691

### DIFF
--- a/config/tables.go
+++ b/config/tables.go
@@ -21,13 +21,14 @@ type (
 
 	//StructureTableCfg contains the names of the base level collections
 	StructureTableCfg struct {
-		ConnTable       string `default:"conn"`
-		DNSTable        string `default:"dns"`
-		HostTable       string `default:"host"`
-		HTTPTable       string `default:"http"`
-		OpenConnTable   string `default:"openconn"`
-		SSLTable        string `default:"ssl"`
-		UniqueConnTable string `default:"uconn"`
+		ConnTable            string `default:"conn"`
+		DNSTable             string `default:"dns"`
+		HostTable            string `default:"host"`
+		HTTPTable            string `default:"http"`
+		OpenConnTable        string `default:"openconn"`
+		SSLTable             string `default:"ssl"`
+		UniqueConnTable      string `default:"uconn"`
+		UniqueConnProxyTable string `default:"uconnproxy"`
 	}
 
 	//DNSTableCfg is used to control the dns analysis module
--- a/config/tables.go
+++ b/config/tables.go
@@ -28,7 +28,7 @@ type (
 		OpenConnTable        string `default:"openconn"`
 		SSLTable             string `default:"ssl"`
 		UniqueConnTable      string `default:"uconn"`
-		UniqueConnProxyTable string `default:"uconnproxy"`
+		UniqueConnProxyTable string `default:"uconnProxy"`
 	}
 
 	//DNSTableCfg is used to control the dns analysis module
--- a/parser/fsimporter.go
+++ b/parser/fsimporter.go
@@ -44,7 +44,6 @@ type (
 		parseThreads         int
 		batchSizeBytes       int64
 		internal             []*net.IPNet
-		httpProxyServers     []*net.IPNet
 		alwaysIncluded       []*net.IPNet
 		neverIncluded        []*net.IPNet
 		alwaysIncludedDomain []string
@@ -71,7 +70,6 @@ func NewFSImporter(res *resources.Resources,
 		parseThreads:         parseThreads,
 		batchSizeBytes:       2 * (2 << 30), // 2 gigabytes (used to not run out of memory while importing)
 		internal:             util.ParseSubnets(res.Config.S.Filtering.InternalSubnets),
-		httpProxyServers:     util.ParseSubnets(res.Config.S.Filtering.HTTPProxyServers),
 		alwaysIncluded:       util.ParseSubnets(res.Config.S.Filtering.AlwaysInclude),
 		neverIncluded:        util.ParseSubnets(res.Config.S.Filtering.NeverInclude),
 		alwaysIncludedDomain: res.Config.S.Filtering.AlwaysIncludeDomain,
@@ -674,7 +672,30 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
 							// parse host
 							fqdn := parseHTTP.Host
 
-							if fs.filterDomain(fqdn) || fs.filterConnPair(srcIP, dstIP) {
+							// parse method type
+							method := parseHTTP.Method
+
+							// check if destination is a proxy server based on HTTP method
+							dstIsProxy := (method == "CONNECT")
+
+							// if the HTTP method is CONNECT, then the srcIP is communicating
+							// to an FQDN through the dstIP proxy. We need to handle that
+							// as a special case here so that we don't filter internal->internal
+							// connections if the dstIP is an internal IP because the dstIP
+							// is an intermediary and not the final destination.
+							//
+							// The dstIP filter check is not included for proxy connections either
+							// because it isn't really the destination and I don't think that it makes
+							// sense in this context to check for it. If the proxy IP is external,
+							// this will also allow a user to filter results from other modules
+							// (e.g., beacons), where false positives might arise due to the proxy IP
+							// appearing as a destination, while still allowing for processing that
+							// data for the proxy modules
+							if dstIsProxy {
+								if fs.filterDomain(fqdn) || fs.filterSingleIP(srcIP) {
+									continue
+								}
+							} else if fs.filterDomain(fqdn) || fs.filterConnPair(srcIP, dstIP) {
 								continue
 							}
 
@@ -686,15 +707,9 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
 							// get aggregation keys for ip addresses and connection pair
 							srcProxyFQDNKey := srcProxyFQDNTrio.MapKey()
 
-							// check if destination is a proxy server
-							dstIsProxy := fs.checkIfProxyServer(dstIP)
-
-							// parse method type
-							method := parseHTTP.Method
-
 							// check if internal IP is requesting a connection
 							// through a proxy
-							if method == "CONNECT" && dstIsProxy {
+							if dstIsProxy {
 
 								// Safely store the number of conns for this uconn
 								mutex.Lock()
--- a/parser/fsimporter.go
+++ b/parser/fsimporter.go
@@ -20,6 +20,7 @@ import (
 	"github.com/activecm/rita/pkg/certificate"
 	"github.com/activecm/rita/pkg/data"
 	"github.com/activecm/rita/pkg/explodeddns"
+	"github.com/activecm/rita/pkg/uconnproxy"
 
 	"github.com/activecm/rita/pkg/host"
 	"github.com/activecm/rita/pkg/hostname"
@@ -174,7 +175,7 @@ func (fs *FSImporter) Run(indexedFiles []*fpt.IndexedFile) {
 		fmt.Printf("\t[-] Processing batch %d of %d\n", i+1, len(batchedIndexedFiles))
 
 		// parse in those files!
-		uconnMap, hostMap, explodeddnsMap, hostnameMap, proxyHostnameMap, useragentMap, certMap := fs.parseFiles(indexedFileBatch, fs.parseThreads, fs.res.Log)
+		uconnMap, uconnProxyMap, hostMap, explodeddnsMap, hostnameMap, useragentMap, certMap := fs.parseFiles(indexedFileBatch, fs.parseThreads, fs.res.Log)
 
 		// Set chunk before we continue so if process dies, we still verify with a delete if
 		// any data was written out.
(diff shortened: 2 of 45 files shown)


pkg/beaconproxy/mongodb.go
```diff
@@ -63,7 +63,7 @@ func (r *repo) CreateIndexes() error {
 }
```
Zalgo2462, line 64: Please change the indexes line above on lines 48-54 to 
```go
indexes := []mgo.Index{
	{Key: []string{"-score"}},
	{Key: []string{"src", "fqdn", "src_network_uuid"}},
	{Key: []string{"src", "src_network_uuid"}},
	{Key: []string{"fqdn"}},
//	{Key: []string{"proxy.ip", "proxy.network_uuid"}}, this one's up to you if you think we might need it in the future
	{Key: []string{"-connection_count"}},
}
```

This incorporates the original index fix that was included in the first draft of #689.

Zalgo2462 (changes requested):

config/tables.go
```diff
@@ -21,13 +21,14 @@ type (
...
-		SSLTable        string `default:"ssl"`
-		UniqueConnTable string `default:"uconn"`
+		ConnTable            string `default:"conn"`
+		DNSTable             string `default:"dns"`
+		HostTable            string `default:"host"`
+		HTTPTable            string `default:"http"`
+		OpenConnTable        string `default:"openconn"`
+		SSLTable             string `default:"ssl"`
+		UniqueConnTable      string `default:"uconn"`
+		UniqueConnProxyTable string `default:"uconnproxy"`
```
Zalgo2462, line 31: Could we use `uconnProxy` so that way we don't have to remember which of `beaconProxy` and `unconnProxy` is uppercase? It looks like most of the multiword collections added so far are camelCase. I don't care much if we go all lowercase or camelCase, but I would like to keep it consistent.

pkg/beaconproxy/analyzer.go
```diff
@@ -65,42 +66,32 @@ func (a *analyzer) start() {
...
-
-			// create query
-			query := bson.M{}
-
-			// if beacon has turned into a strobe, we will not have any timestamps here,
-			// and need to update beaconProxy table with the strobeFQDN flag.
+			// if uconnproxy has turned into a strobe, we will not have any timestamps here,
+			// and we need to update uconnproxy table with the strobe flag. This is being done
+			// here and not in uconnproxy because uconnproxy doesn't do reads, and doesn't know
+			// the updated conn count
```
Zalgo2462, line 72: Are strobes ever turned back into regular unique connections if their connection counts decrement below the threshold? I think there might be a bug in the original beaconing logic regarding this as well.

Would a fix be to simply add "$set": bson.M{"strobe": false} to query when `TsList != nil` here and in the beacons package?

pkg/uconnproxy/mongodb.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// if collection exists, we don't need to do anything else
+	for _, name := range names {
+		if name == collectionName {
+			return nil
+		}
+	}
+
+	indexes := []mgo.Index{
+		{Key: []string{"src", "src_network_uuid", "fqdn"}, Unique: true},
```
Zalgo2462, line 42: Please  swap the ordering of this key to be 
`{Key: []string{"src", "fqdn", "src_network_uuid"}, Unique: true},`

MongoDB can use the prefix of a compound key to service a query. By reordering this, we gain the ability to search for (src, fqdn) pairs without the network uuid which can be helpful for quick debugging. In the current ordering, we have the ability to search for (src, src_network_uuid), but that search is already covered by the key below.

Zalgo2462 (changes requested):
- Found a few small things to change, but generally they relate to simplifying the code or naming rather than correctness
- We may have a bug where rolling strobes are never turned back into normal connections when their connection counters go back down
- Found some performance items that would be easier to fix up now before merging with #654.

pkg/beaconproxy/mongodb.go
fullmetalcache: Changed the indexes to the ones you mentioned

parser/fsimporter.go
```diff
@@ -672,6 +677,31 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
...
+
+								// handle if URI is suffixed with :443, :80, etc.
+								if strings.Contains(uri, ":") {
+									uri = strings.Split(uri, ":")[0]
+								}
+
+								// handle if URI is suffixed with a path like
+								// /index.html, /home, etc.
+								if strings.Contains(uri, "/") {
+									uri = strings.Split(uri, "/")[0]
```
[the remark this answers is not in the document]
fullmetalcache: Updated, tested, and added comments. Thanks for providing the logic! I do remember now you mentioning the speed difference between Split and Slice/Index. Will keep that in mind for the future.

parser/fsimporter.go
```diff
@@ -714,25 +744,24 @@ func (fs *FSImporter) parseFiles(indexedFiles []*fpt.IndexedFile, parsingThreads
...
 								// Check if the map value is set
-								if _, ok := proxyHostnameMap[srcProxyFQDNKey]; !ok {
-									// create new host record with src and dst
-									proxyHostnameMap[srcProxyFQDNKey] = &beaconproxy.Input{
-										Hosts: srcProxyFQDNTrio,
+								if _, ok := uconnProxyMap[srcFQDNKey]; !ok {
+									// create new uconnProxyMap record with src and fqdn
+									uconnProxyMap[srcFQDNKey] = &uconnproxy.Input{
+										Hosts: srcFQDNPair,
+										Proxy: dstUniqIP,
```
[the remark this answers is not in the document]
fullmetalcache: Sounds good!

pkg/beaconfqdn/repository.go
```diff
@@ -48,36 +48,37 @@ type (
...
+		FQDN           string           `bson:"fqdn"`
+		SrcIP          string           `bson:"src"`
+		SrcNetworkName string           `bson:"src_network_name"`
+		SrcNetworkUUID bson.Binary      `bson:"src_network_uuid"`
+		Connections    int64            `bson:"connection_count"`
+		AvgBytes       float64          `bson:"avg_bytes"`
+		Ts             TSData           `bson:"ts"`
+		Ds             DSData           `bson:"ds"`
+		Score          float64          `bson:"score"`
+		ResolvedIPs    data.UniqueIPSet `bson:"resolved_ips"`
```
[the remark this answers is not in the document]
fullmetalcache: Changed it back

pkg/uconnproxy/repository.go
```diff
@@ -0,0 +1,87 @@
...
+// proxy server and a count of the connections.
+type Input struct {
+	Hosts           UniqueSrcHostname
+	TsList          []int64
+	Proxy           data.UniqueIP
+	ConnectionCount int64
+}
+
+//NewUniqueSrcHostname binds a pair of UniqueIPs and an FQDN
+func NewUniqueSrcHostname(source data.UniqueIP, fqdn string) UniqueSrcHostname {
```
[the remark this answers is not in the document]
fullmetalcache: Moved the duplicated methods and functions into data/fqdn.go. I renamed uniqueSrcHostnamePair to be uniqueSrcFQDNPair as I think hostname could be a bit ambiguous as to whether we are talking about a system with a hostname or an fqdn that points to a system.

pkg/beaconproxy/mongodb.go
```diff
@@ -125,15 +125,12 @@ func (r *repo) Upsert(proxyHostnameMap map[string]*Input) {
...
-
-		start := time.Now()
-
+	for _, entry := range uconnProxyMap {
 		// pass entry to dissector
 		dissectorWorker.collect(entry)
 
 		// progress bar increment
-		bar.IncrBy(1, time.Since(start))
+		bar.IncrBy(1)
```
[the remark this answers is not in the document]
fullmetalcache: Yes, I went ahead and updated it in all of the packages. Not sure what sorcery was allowing it to continue working with stuffing in the extra argument. My linter was complaining about it and is what led me to it. This is the function definition that I see in the package:

image

Could be something weird with my system too?

config/tables.go
fullmetalcache: Updated to camelCase for the collection name

pkg/beaconproxy/analyzer.go
fullmetalcache: Yep, that should do it! I added in the extra set field for the not-strobe case. Good catch! I updated in uconn and uconnproxy as well, since we set strobe fields there too.

Also, realized the strobe field for beaconproxy is stored as "strobe" from the RITA side and is attempted to be retrieved with strobeFQDN in the middleware. This should probably be strobeProxy in both places. I can't remember if there was a reason for not just naming it strobe. Thoughts?

In the uconnproxy collection, I am just going to keep it as "strobe" unless there are objections.

pkg/uconnproxy/mongodb.go
fullmetalcache: I was not aware of that, good to know! Re-ordered

pkg/beaconproxy/analyzer.go
Zalgo2462: I'm good with whatever makes the least amount of work. I don't think there was a specific reason for calling it one thing over another. I'm also okay with using `strobeFQDN` in the beaconProxy collection in addition to the beaconFQDN collection. At the end of the day, both collections represent strobes going out to FQDNs. 

However, if you want to call it strobeProxy here and issue a fix-up to the middleware that works for me too.

fullmetalcache: Cool, we can just leave it as strobeFQDN to reduce the workload

pkg/beaconproxy/mongodb.go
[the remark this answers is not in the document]
Zalgo2462: image
This is the definition for the function in the mpb package on my machine. It says its held at version 3.3.4. I think you can set up the progress bars to give an estimated finish time, for which it needs the delta times. However, since we don't display anything like that, I think we're good to remove the delta times anyways.

Zalgo2462 (changes requested):
I was able to split up a dataset we had used for testing the proxy functionality into chunks using a Python script I whipped up. After doing so, the new PR seemed to work well. The results matched the results from importing all of the files at once, so the main issue has definitely been solved :tada:. 

However, old data is not being removed from the `beaconProxy` or `uconnProxy` collection when the chunk ID wraps back around. 

Please add any logic necessary for removing outdated entries to the `pkg/remover` package.
It may be as simple as adding the two collections to the `removeOutdatedCIDs()` method in `pkg/remover/mongodb.go`.

fullmetalcache:
Added the lines for removing old chunks and confirmed working. Split http.log in half, manually imported each file and specified the same CID for both import, confirmed that the chunks are no longer duplicated.

Zalgo2462 (approved):
LGTM

PR #695, opened 2021-08-23
Fix skipping final field in log when the number of log fields does not match the number of fields in the header

### LINKED ISSUE #694: RITA parser crashes when there are more TSV fields in a log entry than there are header fields
If a Zeek log is malformed such that it contains extra fields that were not defined in the log's header, RITA crashes.

For example, consider a DNS log with the following header:
```
#fields\tts\tuid\tid.orig_h\tid.orig_p\tid.resp_h\tid.resp_p\tproto\ttrans_id\trtt\tquery\tqclass\tqclass_name\tqtype\tqtype_name\trcode\trcode_name\tAA\tTC\tRD\tRA\tZ\tanswers\tTTLs\trejected\tagent_hostname\tagent_uuid
```

and a log entry as follows:
```
1517336042.279652\tCVvreZ2AGtqjnP3nnl\t192.168.88.2\t55638\t165.227.88.15\t53\tudp\t42937\t0.069982\t6dde0175375169c68f.dnsc.r-1x.com\t1\tC_INTERNET\t16\tTXT\t0\tNOERROR\tF
\tF\tT\tT\t0\tTXT 18 302f017537c68f5169\t60.000000\tF\tCarol\t5934e4c5-9acb-498f-a706-b4b7200a47aa\t
```

The log entry above has a trailing tab character, so this log entry contains an extra field set to the empty string at the end of the line. Since RITA only expects each line to contain the same number of fields as items in the header, RITA panics and crashes.

Zalgo2462:
Fixes #694. 

In this PR, we ensure that RITA does not attempt to read the end of a log entry if we've already parsed the needed number of fields.

### DIFF
--- a/parser/files/reading.go
+++ b/parser/files/reading.go
@@ -448,15 +448,15 @@ func ParseTSVLine(lineString string, header *BroHeader,
 	}
 
 	//handle last field
-	if lineString != header.Empty && lineString != header.Unset {
-		if fieldMap.NthLogFieldExistsInParseType[tokenCounter] {
-			parseTSVField(
-				lineString,
-				header.Types[tokenCounter],
-				data.Field(fieldMap.NthLogFieldParseTypeOffset[tokenCounter]),
-				logger,
-			)
-		}
+	if tokenCounter < len(header.Names) && /* skip field if there is no matching entry in the names header*/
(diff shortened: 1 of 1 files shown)


fullmetalcache (approved):
Looks good and confirmed that this fixes the crash the for dnscat2-ja3-strobe-agent set. LGTM!

PR #697, opened 2021-08-23
Drop stobe limit down to 86400 

### LINKED ISSUE #696: Reduce strobe limit to 1 connection per second
The current strobe limit is set to 250,000 connections. https://github.com/activecm/rita/blob/afd9a3ee5c3ad1a464f297d38f7871a72a0735ed/etc/rita.yaml#L185

Given that RITA is most often ran against daily datasets, we should drop this limit down to 86,400. In other words, any pair of hosts that connects once per second should be seen as having a persistent connection and the pair should be brought to the user's attention even if the pair has a low beacon score.

Additionally, lowering this limit helps RITA scale to larger datasets.

Zalgo2462:
Fixes #696

Assuming the parsed in dataset represents a day, this drops the strobe limit to 1 connection per second. 

If any pair of hosts is connecting one per second, we want to bring this to the user's attention regardless of the beacon score. Additionally, this prevents RITA from running the beacon analysis modules on very large lists of connections.

### DIFF
--- a/config/static.go
+++ b/config/static.go
@@ -121,7 +121,7 @@ type (
 
 	//StrobeStaticCfg controls the maximum number of connections between any two given hosts
 	StrobeStaticCfg struct {
-		ConnectionLimit int `yaml:"ConnectionLimit" default:"250000"`
+		ConnectionLimit int `yaml:"ConnectionLimit" default:"86400"`
 	}
 )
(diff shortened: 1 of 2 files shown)


caffeinatedpixel (approved):
LGTM

PR #710, opened 2021-10-04
709 ip6 arpa

### LINKED ISSUE #709: IP v6 arpa request are showing up in the "exploded dns" report
While testing RITA, I started noticing that ip6.arpa requests were showing up in the "exploded dns" report, where 'in-addr.arpa' requests were not. This appears to just be an oversight while coding the filters. 

How to reproduce: Run RITA against Zeek DNS logs containing IP v6 arpa lookups.

RITA version: 4.4.0

nighttardis:
Fixes #709 and excludes ip6.arpa requests similar to how in-addr.arpa requests are excluded.


Zalgo2462 (commented):
The code LGTM. Thank you for submitting this patch!

The ACM team will need to come back and test this before pulling it in, but the code itself has my seal of approval.

PR #717, opened 2022-01-16
Add Debian Install into install.sh

but-i-am-dominator:
Install script modified to install Rita on Debian 10 and 11. 
The Zeek Binary comes from The Opensuse Project. This is a trusted source and the install has been tested.
Mongodb does not yet have a release for Debian 11 yet. This install has been tested on Debian 11 with no issues. 
Tested on clean installs of updated Debian 10/11 virtual machines with no issues to note.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -245,6 +245,21 @@ __install_zeek() {
         __install_packages zeek-lts
     fi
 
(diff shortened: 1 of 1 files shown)


but-i-am-dominator:
Not all Commits signed. Correcting.

PR #718, opened 2022-01-16
Add Debian Install to install.sh 

but-i-am-dominator:
Mongodb does not have a release file for Debian 11 (bullseye). This has been tested with no issues.
The Zeek installation binary comes from The OpenSUSE Project. This is an established and trusted project.

This has been tested on VMs with fresh installs of Debian 10/11 with no issues noted.

David Quartarolo is an ACM contractor and can be contacted via Slack if there are any questions.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -124,6 +124,7 @@ __install() {
 
     # Get system information
     __gather_OS
+    __gather_debian_num
     __bro_installed
     __gather_zeek
     __gather_mongo
@@ -155,7 +156,7 @@ __install() {
 
     if [ "$_INSTALL_MONGO" = "true" ]; then
         if [ "$_MONGO_INSTALLED" = "false" ]; then
-            __load "$_ITEM Installing MongoDB" __install_mongodb "$_MONGO_VERSION" 
+            __load "$_ITEM Installing MongoDB" __install_mongodb "$_MONGO_VERSION"
         elif ! __satisfies_version "$_MONGO_INSTALLED_VERSION" "$_MONGO_VERSION" ; then
 
             # Check that the user wants to upgrade
@@ -186,7 +187,7 @@ __install() {
             # Wait for service to come to life
             printf "$_ITEM Sleeping to give the Mongo service some time to fully start..."
             sleep 10
-             
+
             # Set compatibility version in case we updated Mongo. It's fine to do this even if we didn't
             # update Mongo...it's just a bit cleaner to do it here to cut down on code redundancy and logic checks
             __load "$_ITEM Setting Mongo feature compatibility to $_MONGO_VERSION" __update_feature_compatibility "$_MONGO_VERSION"
@@ -237,6 +238,13 @@ __install_zeek() {
                 __add_deb_repo "deb [ arch=$(dpkg --print-architecture) ] http://download.opensuse.org/repositories/security:/zeek/xUbuntu_$(lsb_release -rs)/ /" \
                 "security:zeek" \
                 "https://download.opensuse.org/repositories/security:/zeek/xUbuntu_$(lsb_release -rs)/Release.key"
+                ;;
+            Debian)
+                __install_packages cmake make gcc g++ flex bison libpcap-dev libssl-dev python3 python3-dev python3-git python3-semantic-version swig zlib1g-dev gpg
+                __add_deb_repo  "deb http://download.opensuse.org/repositories/security:/zeek/$_Debian_Release/ /" \
+                "security:zeek" \
+                "https://download.opensuse.org/repositories/security:zeek/$_Debian_Release/Release.key"
+                __freshen_packages
                 ;;
             CentOS|RedHatEnterprise|RedHatEnterpriseServer)
                 __add_rpm_repo "https://download.opensuse.org/repositories/security:/zeek/CentOS_7/security:zeek.repo"
@@ -415,6 +423,13 @@ __install_mongodb() {
                 "mongodb-org-$1" \
                 "https://www.mongodb.org/static/pgp/server-$1.asc"
             ;;
+        Debian)
+        # Mongodb does not have a release file for Debian 11 "Bullseye" yet.
+            __add_deb_repo "deb http://repo.mongodb.org/apt/debian buster/mongodb-org/$1 main" \
+                "mongodb-org-$1" \
+                "https://www.mongodb.org/static/pgp/server-$1.asc"
+            ;;
+
         CentOS|RedHatEnterprise|RedHatEnterpriseServer)
             if [ ! -s /etc/yum.repos.d/mongodb-org-$1.repo ]; then
                 cat << EOF > /etc/yum.repos.d/mongodb-org-$1.repo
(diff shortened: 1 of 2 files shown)


but-i-am-dominator:
@william-stearns  When you have time for some feedback, I would appreciate it.

install.sh
```diff
@@ -238,6 +239,13 @@ __install_zeek() {
                 "security:zeek" \
                 "https://download.opensuse.org/repositories/security:/zeek/xUbuntu_$(lsb_release -rs)/Release.key"
                 ;;
+            Debian)
+                __install_packages cmake make gcc g++ flex bison libpcap-dev libssl-dev python3 python3-dev python3-git python3-semantic-version swig zlib1g-dev gpg
```
Zalgo2462, line 243: I know these dependencies are required to build Zeek. However, I don't believe these are required to install Zeek from a deb package.

install.sh
```diff
@@ -238,6 +239,13 @@ __install_zeek() {
...
+                __add_deb_repo  "deb http://download.opensuse.org/repositories/security:/zeek/$_Debian_Release/ /" \
```
Zalgo2462, line 244: Since we probably won't need to access the variable elsewhere in the code, I would prefer if we made the Debian_Release variable a local variable in this section of the switch statement. Most repositories will use the OS codename rather than the tags the opensuse build service is using.

Zalgo2462:
Can confirm that RITA successfully installs along with Zeek on Debian 10 and Debian 11 using this patch.
I've attached the results of running the installs with `set -x` which echos each line of the script as it executes.
[deb10-install.txt](https://github.com/activecm/rita/files/7885754/deb10-install.txt)
[deb11-install.txt](https://github.com/activecm/rita/files/7885755/deb11-install.txt)

install.sh
but-i-am-dominator: I do not mind at all! After testing with the line removed, I needed to add gpg back as a required dependency.

install.sh
but-i-am-dominator: Understood. I moved this to a local scope and retested it.

install.sh
```diff
@@ -238,6 +238,20 @@ __install_zeek() {
                 "security:zeek" \
                 "https://download.opensuse.org/repositories/security:/zeek/xUbuntu_$(lsb_release -rs)/Release.key"
                 ;;
+            Debian)
+                __install_packages gpg
```
Zalgo2462, line 242: Please test running the installation script on Debian without Zeek by passing the flag `--disable-zeek`. 

Since setting up the repository for both MongoDB and Zeek requires trusting a GPG key, I'm thinking that the MongoDB installation might not run unless gpg is installed. 

If this is the case, please remove this line and add `gpg` to the line `    for pkg in curl coreutils lsb-release yum-utils;` in `__install_installer_deps`

but-i-am-dominator: @Zalgo2462  Sorry for the delay. As you expected, running the install script with the "--disable-zeek" flag caused a failure. After making the suggested change, everything works as expected.

Zalgo2462 (approved):
LGTM. This should be included in the next RITA release.

greengeek:
I just downloaded Debian 11 server (debian-11.2.0-amd64-netinst.iso) and tried the 4.5.1 release and it failed.   It got to this part of the script:

`Would you like to continue running the zeek configuration script and generate a new node.cfg file? (y/n) ? y`

It then returns this:
...
```
[!] Automatic Zeek configuration failed.
[!] Please edit /opt/zeek/etc/node.cfg and run
[!] 'sudo zeekctl deploy' to start Zeek.
[!] Pausing for 20 seconds before continuing.
        [!] Adding Zeek IDS to the path in /etc/profile.d/zeek-path.sh
[-] Installing MongoDB... SUCCESS
```
...

It skipped over the autodetect and said it couldn't find any sniffable interfaces and moved on to the MongoDB part of the install install.  I am running this in VM Workstation and the interface is ens33.  

To verify it wasn't just user error, I downloaded Ubuntu Server 18.04 LTS and ran the 4.5.0 install.sh script and it was able to find the interfaces and gave a bunch more output finding the interface, creating the config file, and enabling and starting the service. 

I tried two runs of it on Debian and got both the same output.  I didn't try to manually install it to see if I could get past the error.

greengeek:
Here is the full screen output:

`

 _ \ _ _| __ __|  \
   /   |     |   _ \
_|_\ ___|   _| _/  _\  v4.5.1

Brought to you by Active CounterMeasures

[-] In order to run the installer, several basic packages must be installed.
        [-] Updating packages... SUCCESS
        [-] Ensuring curl is installed... SUCCESS
        [-] Ensuring coreutils is installed... SUCCESS
        [-] Ensuring lsb-release is installed... SUCCESS
        [-] Ensuring yum-utils is installed... SUCCESS
        [-] Ensuring gpg is installed... SUCCESS
[-] This installer will:
        [-] Install Zeek IDS to /opt/zeek
        [-] Install MongoDB
        [-] Install RITA to /usr/local/bin/rita
        [-] Create a runtime directory for RITA in /var/lib/rita
        [-] Create a configuration directory for RITA in /etc/rita
[-] Installing Zeek IDS... SUCCESS
Note: It is now time to select capture interface(s). Keep the following in mind when making selections:
      1. The interfaces you most likely want to use for capturing start with "eth" or "en" (e.g. eth0, eno1, enp1s0, enx78e7d1ea46da).
         You will generally NOT want to use loopback, bridged, or virtual interfaces (e.g. lo, br-c446eb08dde, veth582437d).
         If you choose to select interfaces belonging to the latter category, proceed at your own risk.

      2. Ensure that your capture interfaces are up before continuing.

Would you like to continue running the zeek configuration script and generate a new node.cfg file? (y/n) ? y
Continuing, all requirements met
This system has 2 cores.
There are no potentially sniffable interfaces.  This script will not be able to generate a node.cfg file as at least one interface is required, exiting.
[!] Automatic Zeek configuration failed.
[!] Please edit /opt/zeek/etc/node.cfg and run
[!] 'sudo zeekctl deploy' to start Zeek.
[!] Pausing for 20 seconds before continuing.
        [!] Adding Zeek IDS to the path in /etc/profile.d/zeek-path.sh
[-] Installing MongoDB... SUCCESS
[!] Starting MongoDB and enabling on startup.
Created symlink /etc/systemd/system/multi-user.target.wants/mongod.service → /lib/systemd/system/mongod.service.
[!] Starting MongoDB process completed.
[!] You can access the MongoDB shell with 'mongo'.
[!] If you need to stop MongoDB,
[!] run 'sudo systemctl stop mongod'.
[-] Setting Mongo feature compatibility to 4.2... SUCCESS
[-] Installing RITA... SUCCESS
[!] To finish the installation, reload the system profile with
[!] 'source /etc/profile'.

 _ \ _ _| __ __|  \
   /   |     |   _ \
_|_\ ___|   _| _/  _\  v4.5.1

Brought to you by Active CounterMeasures

Thank you for installing RITA! Happy hunting!
`

but-i-am-dominator:
@greengeek  The install script grabs an external script to configure the interfaces which Zeek will use to capture packets. 
[Script here:](https://raw.githubusercontent.com/activecm/bro-install/master/gen-node-cfg.sh)

The script grabs the interfaces capable of sniffing based on if they are up and if they are not the loopback interface.
I think your interfaces were not detected for one reason or another. 

The easiest way of testing this is with the command:
`ip -o link | egrep '(state UP|state UNKNOWN|state DORMANT)' | awk '{print $2}' | sed -e 's/:$//' | egrep -v '(^lo$)'`

None of this is unique to the Debian install but, this process is part of the base install script. I retested the install script on a clean build of Debian using the ISO file that you referenced (debian-11.2.0-amd64-netinst.iso) and did not encounter any issues.

I hope this helps but, if you need more assistance or if you find a bug that we can fix: please open up a new issue so that it gets more visibility. (This pull request has already been merged and I don't want any additional comments to be missed.)

PR #726, opened 2022-03-21
Use non-header-including byte size field for tcp connections

### LINKED ISSUE #725: Use non-header including byte size field for tcp connections
Currently, the usage of orig_bytes/resp_bytes and orig_ip_bytes/resp_ip_bytes throughout the code is inconsistent. There is also a problem with beacon scores being skewed by variation in the header size of its connections

- [ ] make usage of fields consistent 
- [ ] use orig_bytes/resp_bytes (header-excluding) for tcp connections, and orig_ip_bytes/resp_ip_bytes for all others

lisaSW:
Closes #725 
- updated incoming/outgoing bytes for tcp connections to use the orig_bytes/resp_bytes fields, which do not include header size. Non-tcp connections will use orig_ip_bytes/resp_ip_bytes, which include the header size.
- field usage is now consistent

### DIFF
--- a/parser/conn.go
+++ b/parser/conn.go
@@ -40,7 +40,19 @@ func parseConnEntry(parseConn *parsetypes.Conn, filter filter, retVals ParseResu
 	srcDstKey := srcDstPair.MapKey()
 
 	roundedDuration := math.Ceil(parseConn.Duration*10000) / 10000
-	twoWayIPBytes := int64(parseConn.OrigIPBytes + parseConn.RespIPBytes)
+
+	// get bytes passed in both directions
+	// note: use orig_bytes field for tcp only, as it does not include the header
+	// use orig_ip_bytes field for all other protocols as the header+data is the only option
+	var bytesSent, bytesReceived int64
+	if parseConn.Proto == "tcp" {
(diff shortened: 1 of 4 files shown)


ethack (approved):
It makes sense to me.

There is one edge case I've seen that may be worth addressing: when the `OrigBytes`/`RespBytes` are incorrect. In this case they will be greater than the `OrigIPBytes`/`RespIPBytes` values. But I think that should never actually be true, so another check could be to fall back to using `OrigIPBytes`/`RespIPBytes` if `OrigBytes > OrigIPBytes || RespBytes > RespIPBytes`.

Zalgo2462 (changes requested):
This change breaks compatibility with ACM's Active Flow product. Active Flow only reports the IP bytes fields since this is what is available in netflow v9 data. Annoyingly, it falsely reports the OrigBytes and RespBytes fields as 0's instead of unset. 

We either need to add a special case in here for RITA, patch Active Flow, or both. 

Conflicting Active Flow code: https://github.com/activecm/AC-Hunter/blob/8c7b481e2e757c8dc2e6c59de42c753f458a8897/active-flow/writer/bro/util.go#L42

Issue #587, opened 2020-09-11
Ubuntu 20.04 support


0x6d6f7468:
It's finally time, folks!

The installer does not currently play nicely with Ubuntu 20.04. I've been given the green light to take a look.

I believe issue #562 contains relevant info. I thought I saw a few more, and will link to them if I find them.

Currently when I run the RITA installer on my Ubuntu 20.04 VM, the installer dies without error. Likely related to #543. Thankfully, I'm able to briefly see that the overwritten installer message is `[-] Installing MongoDB`, which indicates that the installer is silently failing to install Mongo. Good times.

Thanks!

0x6d6f7468:
Removed a /dev/null output redirect on line 622, and that finally yielded a proper error message. Something with the usual kludge isn't working just yet, but at least we know it's a Mongo issue. Just like old times...

0x6d6f7468:
Looks like the Ubuntu 18.04 kludge might not work on Ubuntu 20.04. Recreating the mongo repo file, I got the following error:

![image](https://user-images.githubusercontent.com/25512187/92867809-44184b00-f3be-11ea-8172-6684e4787e83.png)

Chasing that down to just installing `mongodb-org` manually, I get a more detailed error:

![image](https://user-images.githubusercontent.com/25512187/92867863-53979400-f3be-11ea-9db8-d5367ff76951.png)

Installing `mongodb-org*` gives the following error, and I've got a bad feeling about this one:

![image](https://user-images.githubusercontent.com/25512187/92867931-65793700-f3be-11ea-88cf-6532b42aecec.png)

Yikes... I'm gonna get some actual sleep and look into this more tomorrow or this weekend. I hope there's still a reasonable way to get Mongo 3.6 installed on Ubuntu 20.04.

obaidrahman2020:
It is almost impossible to use Rita at the moment. The issues I have faced are as follows:
I installed Ubuntu 20.04 and it will fail at mongo db install. I manually installed mongodb 4.4 as a version between 3.4-3.7 wont install on this version of Ubuntu. however rita eventually fails with the mongodb 4.4 during log import.

I then tried to install the 18.04 version, but that is out of support and a bigger pain since there are lots of hurdle to simply make it work, and not worth the pain.
I will highly appreciate if you can suggest a workable solution on Ubuntu.

0x6d6f7468:
@obaidrahman2020 Thanks for the comment.

[...]

PR #732, opened 2022-04-19
Support Ubuntu 20.04 LTS

Zalgo2462:
Following the discussion in #587, this PR instructs the installer to use the Ubuntu 18 MongoDB 4.2 package when installing on Ubuntu 20. I have tested this out on Ubuntu 20 and was able to successfully install RITA.


fullmetalcache (approved):
lgtm

PR #738, opened 2022-06-08
Don't enable malware domains threat intel list

Zalgo2462:
The malware domains threat intel list is no longer available. This PR remove the threat intel source and alleviates a stall that occurred the first time a user went to import logs with RITA each day. 

The actual implementation for the malware domains list is defined in the rita-bl repository. This PR removes the references to it from RITA, but the implementation should be removed there as well.


lisaSW (approved):
lgtm

PR #741, opened 2022-07-06
Change host summarizer to record max total duration instead of max individual duration found in the uconn collection

Zalgo2462:
This PR changes the `host` package summarizer to record the unique connection with the highest total duration in the `dat.max_duration` and `dat.mdip` fields. Previously, RITA recorded the longest individual connection in this field.


caffeinatedpixel (approved):
lgtm

PR #743, opened 2022-08-17
Patch to limit dataset timestamp range to 24 hours

lisaSW:
- since zeek records connections when they close, some connections that started before the ingested observation period can skew the ts range. We need to cap observation period to the last 24 hours for accurate beaconing analysis.


Zalgo2462 (approved):
Tested well. LGTM.

PR #744, opened 2022-08-17
Patch to fix uconn timestamp storage 

lisaSW:
This is a patch to store every timestamp in the uconn ts list rather than a unique set. This will not impact any of the scoring algorithms, which already gather the timestamps with `$addToSet`

### DIFF
--- a/parser/conn.go
+++ b/parser/conn.go
@@ -124,7 +124,7 @@ func updateUniqueConnectionsByConn(srcIP, dstIP net.IP, srcDstPair data.UniqueIP
 	// ///// INCREMENT THE CONNECTION COUNT FOR THE UNIQUE CONNECTION /////
 	retVals.UniqueConnMap[srcDstKey].ConnectionCount++
 
-	// ///// UNION TIMESTAMP WITH UNIQUE CONNECTION TIMESTAMP SET /////
+	// ///// APPEND TIMESTAMP TO UNIQUE CONNECTION TIMESTAMP LIST /////
 	retVals.UniqueConnMap[srcDstKey].TsList = append(
 		retVals.UniqueConnMap[srcDstKey].TsList, parseConn.TimeStamp,
 	)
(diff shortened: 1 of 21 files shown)


Zalgo2462 (commented):
The code looks good. I will give it a run here in a bit.

We might want to consider using `TsListFull` instead of `TsList` when calling `getTsHistogramScore` on line 218. This would keep the timestamp list in line with the timeline/ histogram scores.

It would be convenient if we could also use `TsListFull` for the rest of the scores. Below, I've written a bit about using `TsListFull` for the ts skew and dispersion scores. If you see any reasons which would prevent us from doing so that I have not addressed below, please let me know.

From a first glance, it looks like the Bowley skew would actually work with `tsListFull`.  If we use `TsListFull`, then `tsLow`, `tsMid`, and `tsHigh` might all be 0. If just `tsLow` is 0, then it doesn't look like we will run into any issues. In the case that both `tsLow` and `tsMid` are 0, we have a special case that makes sure that `tsSkew` gets set to 0 on line 137.  The skew score is defined as `1 - tsSkew`, so a beacon with all of its delta times within half a second would score high on skew. I think that is a reasonable expectation.

It looks like the issue with the dispersion score hits when we go to create the `tsMadmScore` on line 183. We divide by `tsMid` here which could be 0 if we use `tsListFull`. I believe the only distributions where we'd see `tsMid` get set to 0 would be distributions with a huge spike at 0 and a tail of smaller bins out to the right. In these cases, the median deviation from the median should also be near 0. So, I think there is good reason to add a special case here where we just set  the `tsMadmScore` to 1 if `tsMid` is 0.

Another motivation for setting the `tsMadmScore` to 1 when `tsMid` is 0 comes from looking at what happens to the similar coefficient of variation in this case. I believe we can roughly model the problem case by an exponential distribution with a large rate parameter. The CoV is `std. dev / mean`. If we plug in the definitions for these terms for the exponential distribution, we get `sqrt(1/r^2) / (1/ r)`. This simplifies as `(1/r) / (1/r)` which simplifies down to 1, matching our hard coded case.

Zalgo2462 (approved):
We discussed using TsListFull for IP beacon scoring, but that can be implemented in a separate PR if the testing goes well. I tested this PR as it stands and it works well.

PR #745, opened 2022-08-18
Migrate beacons analysis to using full timestamp list

lisaSW:
Beacons analysis was using a unique set of timestamps for its analysis. This PR migrates it to using the full timestamp list for each section of analysis and scoring.


Zalgo2462 (changes requested):
We need to cap an edge case where the dispersion score could go negative. Looks good otherwise.

Zalgo2462 (approved):
LGTM

PR #747, opened 2022-08-23
Bump to 4.6.0

caffeinatedpixel:
Update RITA version to v4.6.0


Zalgo2462 (approved):
I believe we have merged everything in that was slated for the 4.6.0 release.

PR #748, opened 2022-08-24
Fixed some beacon scores from being NaN

caffeinatedpixel:
Added a check when calculating `dsMadmScore` to set it to `0` if it is less than zero or `NaN`. 
During my troubleshooting, I noted that only `dsScore` had instances of `NaN`.  It looks like the other score parameters have checks for dividing by zero.
Tested and confirmed that previously `NaN` scored beacons now have real scores.

### DIFF
--- a/pkg/beacon/analyzer.go
+++ b/pkg/beacon/analyzer.go
@@ -178,7 +178,7 @@ func (a *analyzer) start() {
 
 				//lower dispersion is better
 				dsMadmScore := 1.0 - float64(dsMadm)/float64(dsMid)
-				if dsMadmScore < 0 {
+				if dsMadmScore < 0 || math.IsNaN(dsMadmScore) {
 					dsMadmScore = 0
 				}
 
--- a/pkg/beacon/analyzer.go
+++ b/pkg/beacon/analyzer.go
@@ -177,8 +177,11 @@ func (a *analyzer) start() {
 				}
 
 				//lower dispersion is better
-				dsMadmScore := 1.0 - float64(dsMadm)/float64(dsMid)
-				if dsMadmScore < 0 || math.IsNaN(dsMadmScore) {
+				dsMadmScore := 0.0
+				if dsMid >= 1 {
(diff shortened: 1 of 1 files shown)


pkg/beacon/analyzer.go
```diff
@@ -178,7 +178,7 @@ func (a *analyzer) start() {
 
 				//lower dispersion is better
 				dsMadmScore := 1.0 - float64(dsMadm)/float64(dsMid)
```
Zalgo2462, line 180: For tsMadmScore, we set the score to 1 if the median delta time between connections is zero (avoiding a NaN).
This makes sense because if a connection is rapidly connecting multiple times per second and clustered around 0 seconds, we want to make sure the user sees the beacon.

This fix sets dsMadmScore to 0 if the median data size is zero. I think this also makes sense. If the beacon never sent anything out, then the connections likely failed in the first place.

Its unfortunate we need to handle these edge cases in different ways despite calculating the same metric over different data sets. Nonetheless, I think it is the right call.

We might want to adjust this code to better match the pattern above for clarity.

I believe the following might match a bit better:
```go
dsMadmScore := 0
if dsMid >= 1 {
    dsMadmScore := 1.0 - float64(dsMadm)/float64(dsMid)
}
if dsMadmScore < 0 {
    dsMadmScore = 0
}
```

I'm sorry I didn't catch this during the initial review. The change that introduced this bug is at https://github.com/activecm/rita/pull/742/files#diff-33dddc462a7f31c78c6c17d7eff29a1931a354a8fcdaf9b257204945c2d9b869R177

Zalgo2462 (approved):
I'd like to change the patch to match the pattern for the similar fix to the tsMadmScore, but the code is correct, and I don't necessarily want to hold up the release.

pkg/beacon/analyzer.go
caffeinatedpixel: Changed to match the same behavior as tsMadmScore.

PR #749, opened 2022-08-26
Breaking proxy beacon aggregation

lisaSW:
- fixed breaking proxy beacons aggregtion step typo


caffeinatedpixel (approved):
Ensured that beacons proxy doesn't fail, the dataset gets marked as analyzed, and that the proxy beacons look good.

PR #761, opened 2022-10-27
Stop aggregation phase if there aren't local hosts

### LINKED ISSUE #758: RITA Hangs When No Local Hosts Are Found During Host Aggregation
Host aggregation may fail and hang when importing a chunk of logs that does not contain any local hosts.
The resulting dataset will need to be deleted if this occurs. 

image

Zalgo2462:
Fixes #758

If the InternalSubnets configuration key is incorrectly set, RITA may not find any internal hosts in the Zeek logs being parsed in. In versions 4.6.0, a regression was introduced which causes RITA to hang when it cannot find any internal hosts. 

To reproduce this issue, take the current build of RITA, set your InternalSubnets to "0.0.0.0/32", and import a set of Zeek logs. RITA will fail during the Host Aggregation step.

This PR fixes the source of the hang by checking if there are any internal hosts to analyze before each aggregation phase and exiting early if there aren't any. 

Similarly, to test this PR, set your InternalSubnets to a value that doesn't match any of the hosts in your Zeek logs, and try to run them through RITA. Here is the output from a run from my testing:
```
[+] Importing [/home/logan/Development/ACM/AC-Hunter/sample-data/bro/dnscat2-ja3]:
        [-] Processing batch 1 of 1
        [-] Parsing logs to: dnscat-no-internal-nets ... 
        [-] Parsing /fake/path/conn.log.gz -> dnscat-no-internal-nets
        [-] Parsing /fake/path/dns.log.gz -> dnscat-no-internal-nets
        [-] Parsing /fake/pathhttp.log.gz -> dnscat-no-internal-nets
        [-] Parsing /fake/path/ssl.log.gz -> dnscat-no-internal-nets
        [-] Finished parsing logs in 1.926s
        [-] Host Analysis:            4 / 4  [==================] 100 %
        [!] Skipping Host Aggregation: No Internal Hosts
        [!] No Uconn data to analyze
                [!!] No local network traffic found, please check InternalSubnets in your RITA config (/etc/rita/config.yaml)
        [!] No Proxy Uconn data to analyze
        [!] No TLS or HTTP connections to analyze
        [-] Exploded DNS Analysis:    65530 / 65530  [==================] 100 %
        [-] Hostname Analysis:        65530 / 65530  [==================] 100 %
        [!] No Beacon data to analyze
        [-] Gathering FQDNs for Beacon Analysis ...     [                    ] 
        [!] No FQDN Beacon data to analyze
        [!] No Proxy Beacon data to analyze
        [!] No TLS or HTTP Beacon data to analyze
        [-] UserAgent Analysis:       18 / 18  [==================] 100 %
        [-] UserAgent Aggregation:    18 / 18  [==================] 100 %
        [!] No invalid certificate data to analyze
        [-] Indexing log entries ... 
        [-] Updating metadatabase ... 
        [-] Done!

### CODE REVIEW

File: pkg/beacon/mongodb.go [hunk 1]
```diff
@@ -134,6 +135,21 @@ func (r *repo) Upsert(uconnMap map[string]*uconn.Input, hostMap map[string]*host
 
 	// Phase 2: Summary
 
+	// grab the local hosts we have seen during the current analysis period
```
Zalgo2462, line 138: In each analysis package, I moved the loop which gathers the local hosts above the initialization code for the summary phase workers. Then I added a check to see if the list of local hosts is empty. If so, then the code prints a message and exits the analysis package early.

pkg/beacon/mongodb.go
```diff
@@ -151,15 +167,6 @@ func (r *repo) Upsert(uconnMap map[string]*uconn.Input, hostMap map[string]*host
...
-	var localHosts []data.UniqueIP
-	for _, entry := range hostMap {
-		if entry.IsLocal {
-			localHosts = append(localHosts, entry.Host)
-		}
-	}
-
 	// add a progress bar for troubleshooting
 	p = mpb.New(mpb.WithWidth(20))
 	bar = p.AddBar(int64(len(localHosts)),
```
Zalgo2462, line 172: This is the line causing the error. If we create a progress bar with length 0, then the library assumes that the progress bar should default to the time of creation (why? who knows...). Since the bar is never filled, the code then locks on the line `p.Wait()` below. 

Please help me double check there aren't any other places where we might initialize a progress bar with length 0. I believe all of the bars which depend on the length of the input maps should be safe since the `fsimporter` ensures that they are filled before calling the analysis package.

caffeinatedpixel (approved):
Tested, LGTM. Looked through the progress bar creators and did not find any that would be passed an empty map.

PR #762, opened 2022-11-14
Calculations for max DNS query count overflowing 16MB limit in rolling datasets

lisaSW:
Removed analysis and associated storage for max dns query count score


Zalgo2462 (approved):
Code changes look good. I built the PR and compared it to the master build. The only difference I could find in the output was the removal of the `dat` entries in the `host` collection.

PR #763, opened 2022-11-30
Add rita clean command to remove RITA datasets without MetaDB entries

Zalgo2462:
Users cannot remove broken RITA datasets if the MetaDatabase goes out of sync with the list of databases in MongoDB. This should not happen unless there is a serious bug or crash. Nonetheless, it has happened in the past. 

This PR adds the clean command which searches for RITA datasets without MetaDB records and offers to remove them from the database, freeing up disk space without resorting to MongoDB administration tools.


caffeinatedpixel (approved):
Tested, LGTM

PR #771, opened 2022-12-16
Remove beaconfqdn package

Zalgo2462:
This PR removes the FQDN beacon analysis. The SNI beacon analysis provides better, cleaner results with less resources.


caffeinatedpixel (changes requested):
Lines 28, 122-148 need to be removed in `reporting/templates/templates.go`
Otherwise looks good

caffeinatedpixel (approved):

PR #773, opened 2022-12-19
Beacon scoring patch

lisaSW:
- Part 1: removes 0 second intervals from timestamp scoring calculations. This will address the problem of a high TS score given to src-dst pairs that have a large number of connections that happened at the same time.
image

- Part 2: Part of the histgram score was designed to give a high score to connections that have flat histogram, even if there are more than one flat section (like a beacon that alternates between 1 and 5 connections per hour). Since the cv-based scoring will score that case low, this was meant to allow those beacons to still score high. This PR adds a needed exception to make sure that a minimum of half the day is filled in the histogram before allowing this scoring to override the cv version.
image

- Testing was done across multiple datasets for both expected beacons and false positives. Results can be found in orange sections of internal team spreadsheet.


pkg/beacon/analyzer.go
```diff
@@ -65,28 +65,41 @@ func (a *analyzer) start() {
...
-				diff[i] = res.TsList[i+1] - res.TsList[i]
+				interval := res.TsList[i+1] - res.TsList[i]
+				diffFull[i] = interval
+				if interval > 0 {
+					diff = append(diff, interval)
+				}
+
 			}
 			sort.Sort(util.SortableInt64(diff))
+			sort.Sort(util.SortableInt64(diffFull))
```
Zalgo2462, line 89: Performance wise, it would be faster to generate `diffFull`, sort it, and then loop through `diffFull` appending the non-zero entries to `diff`. This would remove the need to sort `diff` on its own.

Zalgo2462 (changes requested):
I see one small change that can be made to improve performance. 

Otherwise the changes, LGTM.

pkg/beacon/analyzer.go
Zalgo2462: I've pushed up a fix for this. I am currently testing to ensure that no changes were made to the resulting scores.

Zalgo2462: I have verified that none of the scores changed with this fix.

Zalgo2462 (approved):
LGTM

Zalgo2462:
We should update the doc in this package at some point to reflect these changes.

PR #778, opened 2023-01-09
Update Beacons Proxy Scoring

lisaSW:
- Updated Beacon Proxy to use new scoring metrics
- Added yaml weights for scoring
- updated command line and html reporting


Zalgo2462 (approved):
Tested well. Matches other beacon packages.

PR #790, opened 2023-03-21
Change show-long-connections to sort by Total Duration

Zalgo2462:
This PR adds a section to the unique connection analyzer which sums up the connection count, total connection duration, and total (two-way IP) bytes for each unique connection across all of the available chunks and open connection data. The new fields are included in each unique connection document under the following fields:
- `count`
- `tbytes`
- `tdur`

Indexes have been added for each of these fields. 

Building on these schema changes, this PR also updates the `show-long-connections` command to display the unique connections which have been active for the longest **total** duration. Previously, this module only printed out the longest individual connection for each unique connection connection. 

The following fields have been added to the long connections display:
- total duration
- connection count
- total (two-way IP) bytes

The open/ close connection state has also been added to the HTML report. Each of the connection times listed in the HTML report have been properly formatted as seen in the human-readable cli output as well. 

I have tested this PR by running one-off imports as well as by using the following bash script:
```bash
rita_import_chunked() {
    local logdir="$1"
    local dataset="$2"
    for i in {00..23}; do 
        local files=`echo "${logdir}"*.${i}\:*-*`
        ./rita import -R -NC 24 $files "${dataset}"
    done
}
```

I ensured that the summary fields for the rolling imports matched the entries in the `dat` subdocuments in the one-off imports. 

This change requires running a MongoDB aggregation for each unique connection. However, we cannot avoid doing so if we want to roll-up all of the individual `dat` documents during each import. Still, we should pay extra attention to performance during the PR review. On small datasets, I have not seen an adverse change in performance.


Zalgo2462:
I've also added the missing maxdur and strobe indices to bring them in line with the other fields.

pkg/uconn/results.go
```diff
@@ -50,10 +56,13 @@ func LongConnResults(res *resources.Resources, thresh int, limit int, noLimit bo
...
 			"dst_network_uuid": 1,
 			"dst_network_name": 1,
+			"count":            1,
+			"tbytes":           1,
+			"tdur":             1,
 			"tuples":           bson.M{"$slice": []interface{}{"$tuples", 5}},
 			"open":             1,
 		}},
-		{"$sort": bson.M{"maxdur": -1}},
+		{"$sort": bson.M{"tdur": -1}},
```
lisaSW, line 65: Would you mind adding maxdur as a secondary sort here? That way if someone is looking at an older dataset it obviously won't have anything in the tdur column but at least the results will still be sorted by maxdur.

pkg/uconn/analyzer.go
```diff
@@ -51,13 +56,24 @@ func (a *analyzer) close() {
 func (a *analyzer) start() {
 	a.analysisWg.Add(1)
 	go func() {
+		ssn := a.db.Session.Copy()
+		defer ssn.Close()
+
+		uconnColl := ssn.DB(a.db.GetSelectedDB()).C(a.conf.T.Structure.UniqueConnTable)
 
 		for datum := range a.analysisChannel {
```
lisaSW, line 65: Could you add a short comment above each of the next three lines/function calls just giving a really short explanation of each?

lisaSW (changes requested):
Everything looks great. I tested this on both rolling and regular datasets and compared the results to verify that everything is being calculated correctly. I timed the import a few times on two large datasets and there is an increase in import time but not a very significant one.

pkg/uconn/results.go
Zalgo2462: Added

pkg/uconn/analyzer.go
Zalgo2462: Added

lisaSW (approved):

PR #791, opened 2023-03-29
Parsed log field validation

lisaSW:
This PR adds some checks in the log ingestion process. The checks validate that the bare minimum fields required for processing into a collection are present in a log entry before attempting to further parse that long entry for that collection. This was prompted by some malformed/incomplete ip addresses causing RITA to crash.

### DIFF
--- a/parser/dns.go
+++ b/parser/dns.go
@@ -25,7 +25,7 @@ func parseDNSEntry(parseDNS *parsetypes.DNS, filter filter, retVals ParseResults
 			"uid": parseDNS.UID,
 			"src": parseDNS.Source,
 			"dst": parseDNS.Destination,
-		}).Error("Unable to parse valid ip address pair from http log entry, skipping entry.")
+		}).Error("Unable to parse valid ip address pair from dns log entry, skipping entry.")
 		return
 	}
 
--- a/parser/dns.go
+++ b/parser/dns.go
@@ -6,20 +6,43 @@ import (
 	"github.com/activecm/rita/parser/parsetypes"
 	"github.com/activecm/rita/pkg/data"
 	"github.com/activecm/rita/pkg/hostname"
+
+	log "github.com/sirupsen/logrus"
 )
 
-func parseDNSEntry(parseDNS *parsetypes.DNS, filter filter, retVals ParseResults) {
+func parseDNSEntry(parseDNS *parsetypes.DNS, filter filter, retVals ParseResults, logger *log.Logger) {
 
 	// extract and store the dns client ip address
 	src := parseDNS.Source
 	srcIP := net.ParseIP(src)
+
+	// verify that ip address was parsed successfully
+	if srcIP == nil {
+		logger.WithFields(log.Fields{
+			"uid": parseDNS.UID,
+			"src": parseDNS.Source,
+		}).Error("Unable to get valid client ip address from dns log entry, skipping entry.")
+		return
+	}
+
+	// get domain
+	domain := parseDNS.Query
+
+	// verify that domain was parsed successfully
+	if domain == "" {
+		logger.WithFields(log.Fields{
+			"uid":   parseDNS.UID,
+			"query": parseDNS.Query,
+		}).Error("Unable to parse valid domain from dns log entry, skipping entry.")
+		return
+	}
 
 	// Run domain through filter to filter out certain domains
 	// We don't filter out the src ips like we do with the conn
 	// section since a c2 channel running over dns could have an
 	// internal ip to internal ip connection and not having that ip
 	// in the host table is limiting
-	ignore := (filter.filterDomain(parseDNS.Query) || filter.filterSingleIP(srcIP))
+	ignore := (filter.filterDomain(domain) || filter.filterSingleIP(srcIP))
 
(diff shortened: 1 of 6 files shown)


parser/http.go
```diff
@@ -52,7 +64,18 @@ func parseHTTPEntry(parseHTTP *parsetypes.HTTP, filter filter, retVals ParseResu
 		uri = uri[:maxIndex]
 
 		// at this point, the URI should be parsed down to just an FQDN
+		// set fqdn or skip log entry if still unable to parse value
+		if uri == "" {
```
Zalgo2462, line 68: Please remove this check in order to make sure we don't drop anything we would have previously accepted as valid. 

It looks like both the URI and Host fields may be unset in valid Zeek HTTP records. 
I believe this happens when we capture an HTTP request without a host header and with the URI just set to `/`. 

Here is an example Zeek log where something like this occurred:
```
#fields ts      uid     id.orig_h       id.orig_p       id.resp_h       id.resp_p       trans_depth     method  host    uri     referrer        version user_agent      request_body_len response_body_len       status_code     status_msg      info_code       info_msg        tags    username        password        proxied orig_fuids      orig_filenames  orig_mime_types  resp_fuids      resp_filenames  resp_mime_types
1518768200.830914      Cfqzwb2r0r1VWZhsOe      10.55.100.106   60960   23.63.152.72    80       1       -       -       -       -       1.0     -       0       218     408     Request Time-out        -       -       (empty) -       -       -       -       -       -FC0oK8vYEC1rxh8th       -       text/html
```
---

Perhaps we should set the fqdn to be the destination IP address when this occurs so we don't index it as a connection to `""`. This would likely be better addressed in a separate ticket.

parser/dns.go
```diff
@@ -6,20 +6,43 @@ import (
...
 )
 
-func parseDNSEntry(parseDNS *parsetypes.DNS, filter filter, retVals ParseResults) {
+func parseDNSEntry(parseDNS *parsetypes.DNS, filter filter, retVals ParseResults, logger *log.Logger) {
 
 	// extract and store the dns client ip address
 	src := parseDNS.Source
 	srcIP := net.ParseIP(src)
 
+	// verify that ip address was parsed successfully
```
Zalgo2462, line 19: PR #788 introduces a dependency on the `dstIP` as well in the DNS parser. 
https://github.com/activecm/rita/pull/788/files#diff-819801100fce036a714f6949e25d050e069ebe2a7448cbaed5618e091de6bc31R17

Please add a check here similar to the one in the conn log parser to check the `dstIP` as well.

parser/dns.go
```diff
@@ -6,20 +6,43 @@ import (
...
+			"src": parseDNS.Source,
+		}).Error("Unable to get valid client ip address from dns log entry, skipping entry.")
+		return
+	}
+
+	// get domain
+	domain := parseDNS.Query
+
+	// verify that domain was parsed successfully
+	if domain == "" {
```
Zalgo2462, line 32: The domain may be empty in valid DNS queries. 
Here is an example entry in a Zeek log which has an `(empty)` query field:
```
1518765993.164098       CQrwEE4MAKPCM5ATGd      192.168.88.2    12058   199.7.83.42     53      udp     2409    0.073556        (empty) 1       C_INTERNET      48      DNSKEY  0       NOERROR T       F       F       F       1       <unknown type=48>,<unknown type=48>,<unknown type=48>,<unknown type=46> 172800.000000,172800.000000,172800.000000,172800.000000 F
```

Technically this change will throw out some data RITA would previously accept. If a client makes a dummy DNS A type query for "", currently, we would add that client's IP address to the hostname collection record for "". This check would prevent a hostname record being created for "". 

I lean towards removing this change to keep the current behavior. We may want to reconsider how we handle DNS records with an empty query, but I think it would be better addressed as a separate issue.

Zalgo2462 (changes requested):
I tested this PR by taking a set of zeek logs and corrupting the source and destination IP addresses in various ways in the conn, http, ssl, and dns logs. The IP checks all worked well. There is one spot in the dns parser where we will need to check the `dstIP` as well due to change introduced in #788.

I believe we should remove the empty domain checks. In the dns, http, and ssl logs, the fqdn is allowed to be empty in the Zeek logs, and RITA will currently parse these entries without crashing. 
Perhaps, we shouldn't index these connections as heading to the empty string, but I don't believe we should throw out this data entirely.

parser/dns.go
```diff
@@ -6,21 +6,37 @@ import (
...
+	srcIP := net.ParseIP(src)
 	dstIP := net.ParseIP(dst)
 
+	// verify that both addresses were able to be parsed successfully
+	if (srcIP == nil) || (dstIP == nil) {
+		logger.WithFields(log.Fields{
+			"uid": parseDNS.UID,
+			"src": parseDNS.Source,
+			"dst": parseDNS.Destination,
+		}).Error("Unable to parse valid ip address pair from http log entry, skipping entry.")
```
Zalgo2462, line 28: This log message says http where it needs to say dns. I'll push up a quick fix and pull the branch in. Everything else looks good to me.

Zalgo2462 (approved):

PR #792, opened 2023-04-10
Removal of connection count portion of beacon scoring and adjustment of skew

lisaSW:
## Problem 1
The connection count portion of TS Score portion of beacon scoring is artificially inflating the TS score. Originally, the connection count score was meant to reward beacons going off very frequently, with a perfect score being given to beacons that went off every 10 min or less. Since that led to both malware detection failure and false positives, the formula was adjusted to reward beacons that went off at least once per hour. However, the default thresholds in RITA for the beacon modules are already attempting to weed out analyzing host pairs that don't have at least 20 connections.  Since we want the lat least once per hour in a dataset and gave perfect connection count scores to hosts that had at least 24 connections, this score ended up being redundant and inflated the TS Score unnecessarily. This PR removes the connection count portion of TS Scoring and sets a minimum value for the DefaultConnectionThresh  config values of Beacon, Beacon SNI and Beacon Proxy

## Problem 2

<p align="center">
image
</p> 

Currently we set the Skew portion of the TS Score to 100% if Q2 = Q1 or Q2 = Q3 since the Bowley measure of skewness is unreliable when the left or right quartile are equal to the median. This PR builds on that functionality by adding a check to verify that the interquartile range (Q3– Q1) is at least 10. Since the IQR is the denominator of the skew formula, a very small range can only result in a low score, making it unreliable for this purpose.

<p align="center">
image
</p>

##### Connection Count removal: 

- C2 implant benchmarks: No change for most, 2 had a score decrease of 3% or less. 
- False positive benchmarks: No change for most, 3 had slight improvement (score decrease) of 2-3% 
 
##### Skew Score Update: 

- C2 implant benchmarks: No change for most, one had a score decrease of 2% and one had an increase of 7%. 
- False positive benchmarks: No change for most, two had slight improvement (score decrease) of 1-2% 
- Below is an example of a C2 benchmark with a small IQR that benefitted from the change:
<p align="center">
image
</p>

### DIFF
--- a/config/static.go
+++ b/config/static.go
@@ -22,6 +22,12 @@ import (
 // within ~16777216 bytes
 const maxStrobeConnectionLimit int = 86400
 
+// Define the mimimum connection limit for beacon analysis. Currently we require a 24
+// hour block of data for a dataset. Analyzing hosts that have fewer than at least one
+// connection per hour could significantly increase both the analysis time and the number
+// of false positives, so the threshold must be 23 or above.
+const minBeaconConnectionThreshLimit int = 23
+
 type (
 	//StaticCfg is the container for other static config sections
 	StaticCfg struct {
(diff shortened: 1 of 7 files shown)


Zalgo2462 (approved):
Code looks good. I tested the changes on a handful of datasets looking at IP beacons, SNI beacons, and proxy beacons. I saw a couple true positives drop by 1-2%. I also saw many false positives drop by 5-9%. Awesome. :+1:

Zalgo2462:
We will want to update the docs in each of these packages in another PR.

PR #793, opened 2023-04-11
Duration Scoring Update

lisaSW:
## Problem
Some C2 use office hours and only run for a single block during the day, and the duration score is reducing their scores unnecessarily since they do not span the duration of the dataset. This PR contains the following updates to address this issue:

 * Duration will only be calculated if more than 6 hours are represented in the connection frequency histogram

* The final score will be upgraded to take the maximum value of two subscores:

    * Duration Score = max( Dataset Timespan Coverage, Consistency )

    * Dataset Timespan Coverage (original duration score method) = $ts _{max} - ts _{min} \over TS _{max} - TS _{min}$
where $ts$ = a timestamp of the connection pair and $TS$ = a timestamp of the overall dataset

    * Consistency = ${longest\ run\ of\ consecutive\ hours\ seen} \over 12\ hours$
(consecutive includes wrap around from start to end of dataset)

#### Case with consistency score benefit:
<p align="center">
image
</p>

#### Case with coverage score  benefit:
<p align="center">
image
</p>

#### Case with six hour minimum representation benefit:
<p align="center">
image
</p>

## Testing
- C2 implant benchmarks: 
   - No change for most, 5 had a score improvement (increase) of 1-2% and one had an increase of 10% (this one is represented in the first example chart above). 
- False positive benchmarks: 
   - All except 2 saw a score improvement of (decrease) of 1-5% 
   - 2 scores saw an increase in score by 9-10%. However, given that the graphs of the two contained 8 consecutive hours of activity, the duration score increase seems appropriate and the final score was still in the 60% range.

### DIFF
--- a/config/static.go
+++ b/config/static.go
@@ -22,6 +22,12 @@ import (
 // within ~16777216 bytes
 const maxStrobeConnectionLimit int = 86400
 
+// Define the mimimum connection limit for beacon analysis. Currently we require a 24
+// hour block of data for a dataset. Analyzing hosts that have fewer than at least one
+// connection per hour could significantly increase both the analysis time and the number
+// of false positives, so the threshold must be 23 or above.
+const minBeaconConnectionThreshLimit int = 23
+
 type (
 	//StaticCfg is the container for other static config sections
 	StaticCfg struct {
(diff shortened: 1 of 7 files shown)


Zalgo2462 (approved):
For beacons with less than 6 timeline bars populated, this PR significantly cut their scores which is good. However, it does look like on average, scores increased or stayed the same. In particular, the top end of the results seems to have a page or two more of false positives to sift through in some datasets. 

Nonetheless, I think this is worth the extra detections for when beacons repeatedly cross the dataset boundary. 

Everything tested well. LGTM.

PR #814, opened 2023-07-21
Adjust show-dns-fqdn-ips to prompt user to supply database and FQDN i…

kirkhauck:
…f supplied with fewer than two arguments

### Summary
This adjusts the logic to `show-dns-fqdn-ips` argument error handling. Now it will prompt the user to enter both a database and an IP address whenever the user enters fewer than two arguments.


Zalgo2462 (approved):
LGTM

PR #815, opened 2023-07-31
730 update docker resources

### LINKED ISSUE #730: Fix/ Update Resources for Running RITA with Docker
As seen in #729, Go will no longer prioritize the DNS link provided by Docker over the `/hosts` file generated inside the container by Docker. This means that we can no longer link the database container to the RITA container using the name `localhost`. As a result, we can no longer use the stock RITA config file when running RITA with Docker. The `ConnectionString` needs to be set to `db_1` or `rita_db_1` rather than `localhost`. 

We should set a name for the db container in the docker-compose file. Additionally, we should upload a ready-to-go configuration file for running RITA with docker. Finally, we should update the documentation to make it easier to run RITA under Docker.

kirkhauck:
## Summary
Resolves bug preventing running RITA with Docker Compose. Created a ready-to-go configuration file for Docker named `rita_docker.yaml`, which is just a copy of `rita.yaml` but with "localhost" replaced with "db" in `ConnectionString`. Removed "links" key from rita service in `docker-compose.yml` since it is no longer needed. Pointed the Dockerfile to `rita_docker.yaml` instead of `rita.yaml`.

The usage guide was updated to new `docker compose` syntax and to assume user is not in the docker group and needs to use `sudo`.

## Related Issue
closes #730 

## Testing
Pull down RITA image
```
sudo docker pull quay.io/activecm/rita
```

Setup your variables (default config path is `./etc/rita_docker.yaml`)
```
export CONFIG=/path/to/your/rita/config.yaml
export LOGS=/path/to/your/zeek/logs
```

Test the configuration
```
sudo -E docker compose run --rm rita test-config
```

Import a dataset
```
sudo -E docker compose run --rm rita import /logs your-dataset
```

Test a command (e.g. `show-beacons`)
```
sudo -E docker compose run --rm rita show-beacons your-dataset
```

There should be no errors when executing these commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN make CGO_ENABLED=$CGO_ENABLED GOARCH=$GOARCH GOOS=$GOOS
 FROM scratch
 
 WORKDIR /
-COPY --from=rita-builder /go/src/github.com/activecm/rita/etc/rita.yaml /etc/rita/config.yaml
+COPY --from=rita-builder /go/src/github.com/activecm/rita/etc/rita_docker.yaml /etc/rita/config.yaml
 COPY --from=rita-builder /go/src/github.com/activecm/rita/rita /rita
 
 ENTRYPOINT ["/rita"]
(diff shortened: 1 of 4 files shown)


Zalgo2462 (approved):
Tested well :+1:

PR #816, opened 2023-09-06
335 add ipv6 support for single ips

### LINKED ISSUE #335: Add IPv6 support for single IPs
The following line will attempt to append a `/32` on any string configured in the Filtering section if it fails to parse. The rationale is that the user may simply have meant a single IP and forgot to append `/32` to make it CIDR format. However, this also assumes IPv4 as IPv6 would have a suffix of `/128`.
https://github.com/activecm/rita/blob/305-add-neverinclude-to-filtering/parser/filter.go#L66

Suggestions to fix this would be to see if the Golang `net` library has any functions that would help tell these apart. Or check for `:` characters to identify an IPv6 address. Side note: I don't think checking for `.` will work to detect IPv4 as I just saw an RFC where an IPv6 address could be written with a combination of `:` and `.`.

kirkhauck:
### Summary
This resolves the issue of `ParseSubnets` not differentiating between IPv4 and IPv6 addresses before appending a subnet mask. It also adds `TestFilterSingleIPv6` to test the new functionality.

### Related Issue

closes #335

### Testing

`cd` into `parser` and run:
```
go test
```

### DIFF
--- a/util/ip.go
+++ b/util/ip.go
@@ -26,30 +26,43 @@ func init() {
 		})
 }
 
-//ParseSubnets parses the provided subnets into net.ipnet format
-func ParseSubnets(subnets []string) (parsedSubnets []*net.IPNet) {
+// ParseSubnets parses the provided subnets into net.IPNet format
+func ParseSubnets(subnets []string) []*net.IPNet {
+	var parsedSubnets []*net.IPNet
 
 	for _, entry := range subnets {
-		//try to parse out cidr range
+		// Try to parse out CIDR range
 		_, block, err := net.ParseCIDR(entry)
 
-		//if there was an error, check if entry was an IP not a range
+		// If there was an error, check if entry was an IP, not a range
 		if err != nil {
-			// try to parse out IP as range of single host
-			_, block, err = net.ParseCIDR(entry + "/32")
+			ipAddr := net.ParseIP(entry)
+			if ipAddr == nil {
+				fmt.Fprintf(os.Stdout, "Error parsing entry: %s\n", err.Error())
+				continue
+			}
 
-			// if error, report and return
+			// Check if it's an IPv4 or IPv6 address and append the appropriate subnet mask
+			var subnetMask string
+			if ipAddr.To4() != nil {
+				subnetMask = "/32" // IPv4
+			} else {
+				subnetMask = "/128" // IPv6
+			}
+
+			// Append the subnet mask and parse as a CIDR range
+			_, block, err = net.ParseCIDR(entry + subnetMask)
+
 			if err != nil {
 				fmt.Fprintf(os.Stdout, "Error parsing CIDR entry: %s\n", err.Error())
-				os.Exit(-1)
-				return
+				continue
 			}
 		}
 
-		// add cidr range to list
+		// Add CIDR range to the list
 		parsedSubnets = append(parsedSubnets, block)
 	}
-	return
+	return parsedSubnets
 }
 
 //IPIsPubliclyRoutable checks if an IP address is publicly routable. See privateIPBlocks.
--- a/util/ip.go
+++ b/util/ip.go
@@ -34,7 +34,7 @@ func ParseSubnets(subnets []string) []*net.IPNet {
 		// Try to parse out CIDR range
 		_, block, err := net.ParseCIDR(entry)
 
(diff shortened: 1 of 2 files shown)


util/ip.go
```diff
@@ -26,30 +26,43 @@ func init() {
...
+
+			// Append the subnet mask and parse as a CIDR range
+			_, block, err = net.ParseCIDR(entry + subnetMask)
 
-			// if error, report and return
 			if err != nil {
 				fmt.Fprintf(os.Stdout, "Error parsing CIDR entry: %s\n", err.Error())
-				os.Exit(-1)
-				return
+				continue
```
Zalgo2462, line 58: Good call on changing `os.Exit(-1)`. Unfortunately, this stumbles upon a piece of tech debt in RITA that needs fixed up. We should not be crashing out the program here or skipping over the error. Instead we should be returning an error up the call chain all the way to `import.go` so it can properly exit out of the program. Fixing this will require edits in `util/ip.go`, `parser/filter.go`, `parser/fsimporter.go`, and `commands/import.go`.

By `continue`'ing here, we hide the error from the calling function. Instead, we should stop running this function early and return the error to the calling function. In general, each calling function needs to know if an error occurred so it can either recover from the error, return it up the call chain, or (at the top of the chain) crash the program. 

In order to accomplish this, we will need to make the following changes to this function:
- change the return type from `[]*net.IPNet` to `([]*net.IPNet, error)`
- change the return on line 65 to `return parsedSubnets, nil`
- change lines 42 and 58 to `return parsedSubnets, err`

After making these changes, we will need to ensure all the call sites check the error and return it up the chain if needed.  The next functions up the chain are `filter.go:newFilter()` and, in turn, `fsimporter.go:newFSImporter()`. In both of these, we will need to:
- move any in-line calls to the last function we changed to their own line
  - In `newFilter()`, the calls to `ParseSubnets()` are in-line, so they can't extract the error value. Move these calls to their own lines by assigning them to new variables at the beginning of the function. The in-line calls can be replaced with references to these new variables.
- grab the error at the call site 
  - For example, in `newFilter()`, we'd unpack the error in the new lines at the beginning:
   `internalNets, err := util.ParseSubnets(conf.S.Filtering.InternalSubnets)`
- check if `err` isn't nil and return it up the call chain if it isn't
  - `newFilter()` example change: `if err != nil { return filter{}, err }`
- change the return type to add in the error
  - `newFilter()` example change: `filter` becomes `(filter, error)`

After making these type of changes in both  `filter.go:newFilter()` and `fsimporter.go:newFSImporter()`, line 235 in `import.go:run()` will need to be changed to catch and check the error value. In this case, if the err is not nil, the function should return a `cli.NewExitError()` as seen elsewhere in the function. These error values will then cause the main function for RITA to exit with the error given.

The tests will also need to be updated. All of the in-line calls to `ParseSubnets()` will need to be moved to their own lines with their own variable declarations. In these cases, we might use the underscore to throw away the error value since we know it will be `nil` (Ex: `internalNets, _ := util.ParseSubnets(...)`). 

If you want to go further with it, add a `ip_test.go` file and write up a `func TestParseSubnets(t *testing.T)` test that ensures that `ParseSubnets` returns the expected `net.IPNet`s given a set of input strings and only returns an error when an invalid IP address/ CIDR network is given.

Zalgo2462 (changes requested):
The changes look good for addressing the issue. Unfortunately, this PR bumps up against a piece of tech debt in RITA that also needs to be resolved before this can be pulled in.

util/ip.go
kirkhauck: Done and done. I updated the error handling and added `TestParseSubnets` to `ip_test.go`. It should account for valid and invalid IPs.

Zalgo2462:
Code looks good. Tests pass :heavy_check_mark: But, there appears to be an issue with the formatting. I thought vscode would format when you saved down your code if you have the Go extension (https://code.visualstudio.com/docs/languages/go). Let me know if you can get that working. If not, let's meet up tomorrow to look for a fix.

kirkhauck:
Checked on it. I had autosave setup and was relying on that. Turns out that the Go extension doesn't do the auto formatting unless you manually save. Going into each file and pressing `Ctrl + S` saved the day. Lesson learned.

Zalgo2462 (approved):
LGTM

PR #820, opened 2023-12-06
Disable extract-certs-pem.zeek

william-stearns:
It does not appear we use the saved certificates created by this (no longer available) script.


Zalgo2462 (approved):
I was able to successfully install RITA, Zeek, and MongoDB on Ubuntu 20 using this script.

Issue #807, opened 2023-06-21
Bug using docker-compose service linking


Korving-F:
In the repository's docker-compose.yml file an aliasing trick is used so that the default rita.yaml config works out-of-the-box. This linking seems to generally work as intended with other aliases, except for the use of `localhost` and therefore the backend connection still breaks with an `Failed to connect to database: no reachable servers`-error. Not sure how to properly fix it, or what the underlying root cause exactly is though.

I tested this with Docker Compose version v2.17.3, docker version 24.0.2. Rita version v4.8.0. 

```yaml
# docker-compose.yaml
links:
 # give db an alias of "localhost" so that RITA's default config works unchanged
 - db:localhost
```

```yaml
# rita.yaml
MongoDB:
  # See https://docs.mongodb.com/manual/reference/connection-string/
  ConnectionString: mongodb://localhost:27017
```

sko9370:
I came across the same issue as you and it seems like the best practice would be to have rita reference the db container directly by its name "db," leaving the docker-compose.yaml the same but changing the ConnectionString to `mongodb://db:27017`. This is reflected in a [separate config specifically for docker compose use](https://github.com/activecm/rita/blob/master/etc/rita_docker.yaml) that was created it seems after your issue. I created a pull request to specify this explicitly in the documentation.

Issue #796, opened 2023-04-21
Failed to connect to database: no reachable servers


HeroS3c-new:
I get this error: `Failed to connect to database: no reachable servers`
When I use the official docker hub image.
command: `docker run quay.io/activecm/rita import pcap_to_log`

pcap_to_log is a folder that contains all log files generated from a pcap using zeek.

output of `ls -l pcap_to_log`:

>-rw-r--r-- 1 loris loris   280 Apr 15 15:57 capture_loss.log
-rw-r--r-- 1 loris loris  2044 Apr 15 15:57 conn.2023-04-13-14-27-52.log
-rw-r--r-- 1 loris loris   733 Apr 15 15:57 conn.log
-rw-r--r-- 1 loris loris  1118 Apr 15 15:57 dns.2023-04-13-14-27-47.log
-rw-r--r-- 1 loris loris  4063 Apr 15 15:57 files.2023-04-13-14-27-47.log
-rw-r--r-- 1 loris loris  2843 Apr 15 15:57 http.2023-04-13-14-27-47.log
-rw-r--r-- 1 loris loris 28021 Apr 15 15:57 loaded_scripts.2023-04-13-14-27-47.log
-rw-r--r-- 1 loris loris   254 Apr 15 15:57 packet_filter.2023-04-13-14-27-47.log
-rw-r--r-- 1 loris loris   707 Apr 15 15:57 stats.2023-04-13-14-27-47.log
-rw-r--r-- 1 loris loris   719 Apr 15 15:57 stats.log
-rw-r--r-- 1 loris loris  1157 Apr 15 15:57 telemetry.2023-04-13-14-27-47.log
-rw-r--r-- 1 loris loris  1157 Apr 15 15:57 telemetry.log

Zalgo2462:
Hello, please check out https://github.com/activecm/rita/blob/master/docs/Docker%20Usage.md

We recommend using docker compose in order to easily link together RITA and MongoDB. 

You will first need a copy of the RITA configuration file. A copy can be found at https://raw.githubusercontent.com/activecm/rita/master/etc/rita.yaml

Next, edit that configuration file to ensure the internal subnets match your own. (See [these lines](https://github.com/activecm/rita/blob/master/etc/rita.yaml#L81)) If you are using standard RFC1918 internal subnets, you may skip this step. 

Then, grab a copy of the docker-compose file at https://raw.githubusercontent.com/activecm/rita/master/docker-compose.yml

[...]

PR #823, opened 2024-03-21
Updated docker documentation and db container name to avoid confusion with db volume

sko9370:
Related to issue #807, updated docker compose usage documentation to explicitly mention that users should use "rita_docker.yaml" and not "rita.yaml" as the template. Additionally, renamed the mongodb container name from "db" to mongo-db to avoid confusion and overlap with the db volume also used in the docker-compose.yml.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 
 services:
-  db:
+  mongo-db:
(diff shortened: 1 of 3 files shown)


dspruell:
+1, helpful improvements. May also relate to #796.

PR #832, opened 2026-01-08
bumping patch version for DB hotfix

SamuelCarroll:
Want to add a new binary in addition to this, would allow AC-Hunter change to just pull latest release/binary


Makefile
```diff
@@ -1,4 +1,5 @@
-VERSION := $(shell git describe --abbrev=0 --tags)
+#VERSION := $(shell git describe --abbrev=0 --tags)
+VERSION := v4.9.1
```
0x6d6f7468: We should probably leave this as it was. This info will be auto-updated when we create the new tag in rita-legacy for release.

install.sh
```diff
@@ -4,7 +4,7 @@
 # activecountermeasures.com
 
 # CONSTANTS
-_RITA_VERSION="v4.8.0"
+_RITA_VERSION="v4.9.1"
```
0x6d6f7468: Please change this to `v4.8.1`, as I don't think we're planning to make this a new minor release (over just a patch update)

install.sh
```diff
@@ -4,7 +4,7 @@
...
 _MONGO_VERSION="4.2"
```
0x6d6f7468: This probably also needs to be updated, just to ensure that correct versions of Mongo are installed.

0x6d6f7468 (changes requested):

install.sh
SamuelCarroll: We have both a v4.8.1 and a v4.9.0 on the releases page
https://github.com/activecm/rita-legacy/releases

EDIT: double checking it looks like the v4.9.0 was just a draft release

0x6d6f7468: Apologies, perhaps 4.8.2 then as 4.8.1 is the latest released version.

install.sh
```diff
@@ -216,7 +216,7 @@ __install_installer_deps() {
 }
 
 __install_zeek_ubuntu() {
-    if [[ "$_OS_CODENAME" == "focal" || "$_OS_CODENAME" == "jammy" ]]; then 
+    if [[ "$_OS_CODENAME" == "focal" || "$_OS_CODENAME" == "jammy" || "$_OS_CODENAME" == "noble" ]]; then
```
0x6d6f7468: The condition for `noble` can probably be dropped, to my knowledge rita-legacy does not support Ubuntu 24.04.

install.sh
```diff
@@ -426,7 +426,7 @@ __update_feature_compatibility() {
 }
 
 __install_mongodb_ubuntu() {
-    if [[ "$_OS_CODENAME" == "bionic" || "$_OS_CODENAME" == "focal" ]]; then 
+    if [[ "$_OS_CODENAME" == "bionic" || "$_OS_CODENAME" == "focal" || "$_OS_CODENAME" == "noble "]]; then
```
0x6d6f7468: Same thing here.

SamuelCarroll:
Tried merging with develop to solve issues, lots of things pop up as changes as a result...

PR #842, opened 2026-01-12
Update Mongo to Address CVE-2025-14847

0x6d6f7468:
Bump Mongo version to 4.4.30 to address CVE-2025-14847.

Tested on an Ubuntu 18.04 DO droplet with the install script and a manually compiled RITA binary to make sure that the correct Mongo version is installed and the binary works properly (note that quay.io is only included below due to how we are tagging the image at build time; this was tested locally and not pulling from the Internet):

image

Also made sure that Docker images build properly and docker compose runs properly:

image

image

Also updated Dockerfile container versions to use go version 1.22, to reflect an upstream change to the `go.mod` file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.17-alpine AS rita-builder
+FROM golang:1.22-alpine AS rita-builder
 
 RUN apk add --no-cache git make ca-certificates wget build-base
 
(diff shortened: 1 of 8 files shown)


0x6d6f7468:
Performing additional testing at Liza's request; All of the following were performed from the docker compose context:

importing data:

image

Listing data:

image

HTML report generation (haven't figured out a good way to retrieve the report):

image

SNI beacon retrieval (to compensate for not getting the HTML report):

image

0x6d6f7468:
Figured out copying the report from a container.

image

image

All looks good.

